PR #807, opened 2024-02-29
RETH Yield Source

cashd:
This PR adds Rocket Pool Ether (RETH) as a yield source. RETH is a yield bearing token that updates in price everyday by oracles that aggregate and report the total rewards value accrued by Rocket Pool's validator set. This integration was heavily influenced by the stETH integration #718.


contracts/src/instances/reth/RETHBase.sol
```diff
@@ -0,0 +1,153 @@
...
+        view
+        override
+        returns (uint256 shareAmount)
+    {
+        return _rocketTokenReth.balanceOf(address(this));
+    }
+
+    /// @dev We override the message value check since this integration is
+    ///      payable.
+    function _checkMessageValue() internal pure override {}
```
jalextowle, line 143: We should update this so that it the contract isn't payable since ETH isn't a deposit asset.

contracts/src/deployers/reth/RETHHyperdriveDeployerCoordinator.sol
```diff
@@ -0,0 +1,126 @@
...
+        // Otherwise, transfer vault shares from the LP and approve the
+        // Hyperdrive pool.
+        rocketTokenReth.transferFrom(_lp, address(this), _contribution);
+        rocketTokenReth.approve(address(_hyperdrive), _contribution);
+
+        return value;
+    }
+
+    /// @dev Allows the contract to receive ether.
+    function _checkMessageValue() internal view override {}
```
jalextowle, line 92: We should make this prevent payments since we don't have ETH as a deposit asset.

contracts/src/deployers/reth/RETHHyperdriveDeployerCoordinator.sol
```diff
@@ -0,0 +1,126 @@
...
+        address _lp,
+        uint256 _contribution,
+        IHyperdrive.Options memory _options
+    ) internal override returns (uint256 value) {
+        // If base is the deposit asset, revert because depositing as base
+        // is not supported for the rETH integration.
+        if (_options.asBase) {
+            revert IHyperdrive.UnsupportedToken();
+        }
+        // Otherwise, transfer vault shares from the LP and approve the
```
jalextowle, line 83: Nit: Let's add a newline

```suggestion

        // Otherwise, transfer vault shares from the LP and approve the
```

test/instances/reth/RETHHyperdrive.t.sol
```diff
@@ -0,0 +1,1052 @@
...
+
+        // Start recording event logs.
+        vm.recordLogs();
+    }
+
+    /// Deploy and Initialize ///
+
+    function test__reth__deployAndInitialize() external {
+        // Deploy and Initialize the rETH hyperdrive instance. Excess ether is
+        // sent, and should be returned to the sender.
```
jalextowle, line 254: This comment should be updated. We don't send ether.

test/instances/reth/RETHHyperdrive.t.sol
```diff
@@ -0,0 +1,1052 @@
...
+        verifyRethWithdrawal(
+            bob,
+            baseProceeds,
+            true,
+            totalRethSupplyBefore,
+            bobBalancesBefore,
+            hyperdriveBalancesBefore
+        );
+
+        vm.stopPrank();
```
jalextowle, line 567: Nit: We don't need this since it's occurring at the end of the test.

test/instances/reth/RETHHyperdrive.t.sol
```diff
@@ -0,0 +1,1052 @@
...
+        verifyRethWithdrawal(
+            bob,
+            shareProceeds,
+            false,
+            totalRethSupplyBefore,
+            bobBalancesBefore,
+            hyperdriveBalancesBefore
+        );
+
+        vm.stopPrank();
```
jalextowle, line 630: Nit: We don't need this since it's occurring at the end of the test.

test/instances/reth/RETHHyperdrive.t.sol
```diff
@@ -0,0 +1,1052 @@
...
+                traderBalancesBefore.rethBalance
+            );
+
+            // Ensure the total supply was updated correctly.
+            assertApproxEqAbs(
+                rocketTokenRETH.totalSupply(),
+                totalSharesBefore + sharesMinted,
+                1
+            );
+        } else {
```
jalextowle, line 912: We can remove this right?

test/instances/reth/RETHHyperdrive.t.sol
```diff
@@ -0,0 +1,1052 @@
...
+            bob,
+            shareProceeds,
+            false,
+            totalRethSupplyBefore,
+            bobBalancesBefore,
+            hyperdriveBalancesBefore
+        );
+    }
+
+    function basePaidAfterFee(
```
jalextowle, line 852: We can remove this right?

jalextowle (commented):
This is looking great! There are a couple of nits.

jalextowle (approved):

contracts/src/instances/reth/RETHBase.sol
```diff
@@ -0,0 +1,153 @@
...
+
+    /// @dev Allows ether to be received only from the Rocket Pool rETH
+    ///      token contract. Supports withdrawing as ethers from this
+    ///      yield source.
+    receive() external payable {
+        if (msg.sender != address(_rocketTokenReth)) {
+            revert IHyperdrive.TransferFailed();
+        }
+    }
```
jalextowle, line 152: Nice!

contracts/src/instances/reth/RETHBase.sol
cashd: This one needs to stay because the contract receives ether when we are withdrawing as base, aka burning reth -> ether

jalextowle: We only use `_checkMessageValue` in `openLong`, `openShort`, `addLiquidity`, and `initialize`. This just prevents ETH from being sent by the user into Hyperdrive via one of these functions. It will not prevent ether from being received by the `receive` function. Let me know if I'm missing something, but if you revert to make this non-payable, all of the existing tests should still pass.

cashd: Great explanation thank you! Yeah you are right I just needed to change what error I am expecting to revert.

contracts/src/instances/reth/RETHBase.sol
```diff
@@ -138,9 +138,12 @@ abstract contract RETHBase is HyperdriveBase {
         return _rocketTokenReth.balanceOf(address(this));
     }
 
-    /// @dev We override the message value check since this integration is
-    ///      payable.
-    function _checkMessageValue() internal pure override {}
+    /// @dev Disallows the contract to receive ether, when opening positions.
+    function _checkMessageValue() internal view override {
+        if (msg.value > 0) {
+            revert IHyperdrive.TransferFailed();
```
jalextowle, line 144: The error is message we typically use here is “NotPayable.” Once we update this, it’s g2g in my book!

PR #833, opened 2024-03-08
Updated `get_spot_price_after_long`

jalextowle:
Previously, `get_spot_price_after` long would simulate the spot price after opening a long without taking fees into account. This PR updates the function to take fees into account. As part of this, `calculate_open_long` no longer relies on this function to catch negative interest cases.


crates/hyperdrive-math/src/long/open.rs
```diff
@@ -58,7 +61,83 @@ impl State {
...
+    };
+
+    use super::*;
+
+    #[tokio::test]
+    async fn fuzz_get_spot_price_after_long() -> Result<()> {
+        // Spawn a test chain and create two agents -- Alice and Bob. Alice
+        // is funded with a large amount of capital so that she can initialize
+        // the pool. Bob is funded with a small amount of capital so that we
+        // can test `get_max_short` when budget is the primary constraint.
```
dpaiton, line 89: last sentence here looks like copy pasta from somewhere else -- no max short in this test right

dpaiton (approved):
Thanks for fixing this!

dpaiton (approved):

jrhea (approved):

PR #840, opened 2024-03-13
Support more options in the factory

jalextowle:
This PR makes it possible for the HyperdriveFactory to initialize with either `options.asBase = true` or `options.asBase = false`. Since it is cumbersome for the factory to handle vault shares (think about the case of Lido shares for an example), this PR moves the initialization logic into the deployer coordinators since deployer coordinators are tightly coupled to yield sources and can include bespoke logic.

### DIFF
--- a/contracts/src/deployers/HyperdriveDeployerCoordinator.sol
+++ b/contracts/src/deployers/HyperdriveDeployerCoordinator.sol
@@ -312,6 +312,63 @@ abstract contract HyperdriveDeployerCoordinator is
         return target;
     }
 
+    /// @notice Initializes a pool that was deployed by this coordinator.
+    /// @dev This function utilizes several helper functions that provide
+    ///      flexibility to implementations.
+    /// @param _deploymentId The ID of the deployment.
+    /// @param _lp The LP that is initializing the pool.
+    /// @param _contribution The amount of capital to supply. The units of this
+    ///        quantity are either base or vault shares, depending on the value
+    ///        of `_options.asBase`.
+    /// @param _apr The target APR.
+    /// @param _options The options that configure how the initialization is
+    ///        settled.
+    /// @return lpShares The initial number of LP shares created.
+    function initialize(
+        bytes32 _deploymentId,
+        address _lp,
+        uint256 _contribution,
+        uint256 _apr,
+        IHyperdrive.Options memory _options
+    ) external payable returns (uint256 lpShares) {
+        // Check that the message value is valid.
+        _checkMessageValue();
+
+        // Ensure that the instance has been fully deployed.
+        IHyperdrive hyperdrive = IHyperdrive(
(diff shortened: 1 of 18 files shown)


contracts/src/deployers/HyperdriveDeployerCoordinator.sol
```diff
@@ -323,6 +380,28 @@ abstract contract HyperdriveDeployerCoordinator is
...
+    /// @param _options The options that configure how the initialization is
+    ///        settled.
+    /// @return value The value that should be sent in the initialize
+    ///         transaction.
+    function _prepareInitialize(
+        IHyperdrive _hyperdrive,
+        address _lp,
+        uint256 _contribution,
+        IHyperdrive.Options memory _options
+    ) internal virtual returns (uint256 value);
```
jrhea, line 399: is there a reason the return value is named here and unnamed in the implementations?

contracts/src/deployers/erc4626/ERC4626HyperdriveDeployerCoordinator.sol
```diff
@@ -39,6 +44,53 @@ contract ERC4626HyperdriveDeployerCoordinator is HyperdriveDeployerCoordinator {
...
+    /// @return The value that should be sent in the initialize transaction.
+    function _prepareInitialize(
+        IHyperdrive _hyperdrive,
+        address _lp,
+        uint256 _contribution,
+        IHyperdrive.Options memory _options
+    ) internal override returns (uint256) {
+        // If base is the deposit asset, transfer base from the LP and approve
+        // the Hyperdrive pool.
+        if (_options.asBase) {
```
jrhea, line 65: all this if statement really has to do is set the "token".  the rest of the code could be outside the conditional

jrhea (approved):
lgtm. optional suggestion

also, why was code coverage skipped?

contracts/src/deployers/erc4626/ERC4626HyperdriveDeployerCoordinator.sol
```diff
@@ -39,6 +44,53 @@ contract ERC4626HyperdriveDeployerCoordinator is HyperdriveDeployerCoordinator {
...
+        // Hyperdrive pool.
+        else {
+            address vault = address(
+                IERC4626Hyperdrive(address(_hyperdrive)).vault()
+            );
+            ERC20(vault).safeTransferFrom(_lp, address(this), _contribution);
+            ERC20(vault).forceApprove(address(_hyperdrive), _contribution);
+        }
+
+        return 0;
```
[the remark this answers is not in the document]
jalextowle: I added a comment to this effect.

jalextowle:
> lgtm. optional suggestion
> 
> also, why was code coverage skipped?

I'm not sure why it was skipped, but it doesn't looked like it's being skipped in the latest pipeline.

PR #841, opened 2024-03-13
Renzo Integration

sentilesdal:
Adds Renzo's ezETH as an integration.  Because their pool was launched Dec 15, 2023, there wasn't a large window for testing on a mainnet fork.  The have a series of operator delegators that use a series of oracles for the token TVLs which limits testing capability without heavy mocking of the whole protocol.  Instead, I used a 15 position duration to test with the actual deployed renzo code.  

Notes:
- Error bars had to be increased because of compounding rounding errors with Renzo's calculateTVLs().  I noted where this happened with `TODO`s
- We can pass a `uin256 refererId` to `restakingManager.depositETH()`.  I left it blank for now but happy to add.
- commented out tests that don't apply,  I'll remove them after review makes sure they're not needed.


contracts/src/instances/ezeth/ezETHBase.sol
```diff
@@ -0,0 +1,158 @@
...
+    /// @notice Instantiates the ezETH Hyperdrive base contract.
+    /// @param __restakeManager The Renzo Restakemanager contract.
+    constructor(IRestakeManager __restakeManager, IERC20 __ezETH) {
+        _restakeManager = __restakeManager;
+        _ezETH = __ezETH;
+        // TODO: see if we can get this to work so we don't need to pass in __ezETH
+        // (, bytes memory data) = address(__restakeManager).call(
+        //     abi.encodeWithSignature("ezETH()")
+        // );
+        // _ezETH = abi.decode(data, (IERC20));
```
cashd, line 35: this should work if you decode as an address type

contracts/src/instances/ezeth/ezETHBase.sol
```diff
@@ -0,0 +1,158 @@
...
+    /// Yield Source ///
+
+    /// @dev Accepts a transfer from the user in base token.
+    /// @param _baseAmount The base amount to deposit.
+    /// @return sharesMinted The shares that were minted in the deposit.
+    /// @return refund The amount of ETH to refund. This should be zero for
+    ///         yield sources that don't accept ETH.
+    function _depositWithBase(
+        uint256 _baseAmount,
+        bytes calldata
```
cashd, line 47: nit add "unused"

contracts/src/instances/ezeth/ezETHBase.sol
```diff
@@ -0,0 +1,158 @@
...
+    /// @dev Process a withdrawal in vault shares and send the proceeds to the
+    ///      destination.
+    /// @param _shareAmount The amount of vault shares to withdraw.
+    /// @param _destination The destination of the withdrawal.
+    function _withdrawWithShares(
+        uint256 _shareAmount,
+        address _destination,
+        bytes calldata // unused
+    ) internal override {
+        // Transfer the stETH shares to the destination.
```
cashd, line 101: nit *ezETH

contracts/src/interfaces/IRestakeManager.sol
```diff
@@ -0,0 +1,57 @@
+// SPDX-License-Identifier: Apache-2.0
+pragma solidity ^0.8.0;
```
cashd, line 2: can we pin as `8.2.0`

test/instances/ezETH/EzETHHyperdrive.t.sol
```diff
@@ -0,0 +1,1159 @@
...
+        // Verify that the correct events were emitted.
+        verifyFactoryEvents(
+            deployerCoordinator,
+            hyperdrive,
+            bob,
+            contribution,
+            FIXED_RATE,
+            config.minimumShareReserves,
+            new bytes(0),
+            // NOTE: Tolerance since stETH uses mulDivDown for share calculations.
```
cashd, line 354: *ezETH

test/instances/ezETH/EzETHHyperdrive.t.sol
```diff
@@ -0,0 +1,1159 @@
...
+    function test_open_long_with_ezeth(uint256 basePaid) external {
+        // Get some balance information before the deposit.
+        (, , uint256 totalPooledEtherBefore) = RESTAKE_MANAGER.calculateTVLs();
+        uint256 totalSharesBefore = EZETH.totalSupply();
+        AccountBalances memory bobBalancesBefore = getAccountBalances(bob);
+        AccountBalances memory hyperdriveBalancesBefore = getAccountBalances(
+            address(hyperdrive)
+        );
+
+        // Bob opens a long by depositing stETH.
```
cashd, line 465: *ezETH

test/instances/ezETH/EzETHHyperdrive.t.sol
```diff
@@ -0,0 +1,1159 @@
...
+        basePaid = basePaid.normalizeToRange(
+            2 * hyperdrive.getPoolConfig().minimumTransactionAmount,
+            HyperdriveUtils.calculateMaxLong(hyperdrive)
+        );
+
+        (, , uint256 totalPooledEther) = RESTAKE_MANAGER.calculateTVLs();
+        uint256 totalShares = EZETH.totalSupply();
```
cashd, line 472: I would create a util function to go from base -> shares in the test file

test/instances/ezETH/EzETHHyperdrive.t.sol
```diff
@@ -0,0 +1,1159 @@
...
+        uint256 sharesPaid = basePaid.mulDivDown(totalShares, totalPooledEther);
+        openLong(bob, sharesPaid, false);
+
+        // Ensure that Lido's aggregates and the token balances were updated
```
cashd, line 476: *Renzo

contracts/src/interfaces/IRestakeManager.sol
jalextowle: 0.8.20*

contracts/src/instances/ezeth/ezETHBase.sol
sentilesdal: yeah that worked, thanks!

contracts/src/deployers/ezeth/EzETHHyperdriveCoreDeployer.sol
```diff
@@ -0,0 +1,69 @@
...
+    IERC20 public immutable ezETH;
+
+    /// @notice Instantiates the core deployer.
+    /// @param _restakeManager The Renzo contract.
+    constructor(IRestakeManager _restakeManager) {
+        restakeManager = _restakeManager;
+        (, bytes memory data) = address(_restakeManager).call(
+            abi.encodeWithSignature("ezETH()")
+        );
+        ezETH = IERC20(abi.decode(data, (address)));
```
jalextowle, line 29: Why don't we just put this function in the RestakeManager interface?

contracts/src/deployers/ezeth/EzETHHyperdriveCoreDeployer.sol
```diff
@@ -0,0 +1,69 @@
...
+/// @author DELV
+/// @title EzETHHyperdriveCoreDeployer
+/// @notice The core deployer for the EzETHHyperdrive implementation.
+/// @custom:disclaimer The language used in this code is for coding convenience
+///                    only, and is not intended to, and does not, have any
+///                    particular legal or regulatory significance.
+contract EzETHHyperdriveCoreDeployer is IHyperdriveCoreDeployer {
+    /// @notice The Renzo contract.
+    IRestakeManager public immutable restakeManager;
+    /// @notice the ezETH token contract.
```
jalextowle, line 19: Nit: Newline

```suggestion
    
    /// @notice the ezETH token contract.
```

contracts/src/deployers/ezeth/EzETHHyperdriveDeployerCoordinator.sol
```diff
@@ -0,0 +1,133 @@
...
+/// @custom:disclaimer The language used in this code is for coding convenience
+/// @notice The deployer coordinator for the EzETHHyperdrive implementation.
+///                    only, and is not intended to, and does not, have any
+///                    particular legal or regulatory significance.
+contract EzETHHyperdriveDeployerCoordinator is HyperdriveDeployerCoordinator {
+    using FixedPointMath for uint256;
+
+    /// @notice The Renzo contract.
+    IRestakeManager public immutable restakeManager;
+    /// @notice The ezETH token contract.
```
jalextowle, line 22: Nit: This repo's style is to always put a newline between separate blocks of code. Looks like this change will need to be made in a bunch of places.

```suggestion

    /// @notice The ezETH token contract.
```

contracts/src/deployers/ezeth/EzETHHyperdriveDeployerCoordinator.sol
```diff
@@ -0,0 +1,133 @@
...
+            _target2Deployer,
+            _target3Deployer,
+            _target4Deployer
+        )
+    {
+        restakeManager = _restakeManager;
+        (, bytes memory data) = address(restakeManager).call(
+            abi.encodeWithSignature("ezETH()")
+        );
+        ezETH = IERC20(abi.decode(data, (address)));
```
jalextowle, line 55: Why isn't this just in the interface? It seems like a view function too, so if we were going to do a low-level call, a staticcall would be better.

contracts/src/interfaces/IezETHHyperdriveRead.sol
```diff
@@ -0,0 +1,11 @@
+// SPDX-License-Identifier: Apache-2.0
+pragma solidity 0.8.20;
+
+import { IHyperdriveRead } from "./IHyperdriveRead.sol";
+import { ILido } from "./ILido.sol";
+
+interface IezETHHyperdriveRead is IHyperdriveRead {
```
jalextowle, line 7: Nit: We'll want to change the file name too

```suggestion
interface IEzETHHyperdriveRead is IHyperdriveRead {
```

contracts/src/interfaces/IezETHHyperdriveRead.sol
```diff
@@ -0,0 +1,11 @@
...
+    /// @notice Gets the Lido contract used as this pool's yield source.
+    /// @return The Lido contract.
+    function lido() external view returns (ILido);
```
jalextowle, line 10: This should probably be something else, right?

contracts/src/interfaces/IezETHHyperdriveCore.sol
```diff
@@ -0,0 +1,15 @@
...
+
+interface IezETHHyperdriveCore is IHyperdriveCore {
+    /// @notice Transfers the contract's balance of a target token to the fee
+    ///         collector address.
+    /// @dev WARN: It is unlikely but possible that there is a selector overlap
+    ///      with 'transferFrom'. Any integrating contracts should be checked
+    ///      for that, as it may result in an unexpected call from this address.
+    /// @param _target The target token to sweep.
+    function sweep(IERC20 _target) external;
+}
```
jalextowle, line 15: This is part of `IHyperdriveCore` and can be removed entirely. You'll only need the updated read interface for this integration.

contracts/src/interfaces/IezETHHyperdrive.sol
```diff
@@ -0,0 +1,13 @@
+// SPDX-License-Identifier: Apache-2.0
+pragma solidity 0.8.20;
+
+import { IHyperdrive } from "./IHyperdrive.sol";
+import { IezETHHyperdriveCore } from "./IezETHHyperdriveCore.sol";
+import { IezETHHyperdriveRead } from "./IezETHHyperdriveRead.sol";
+
+// prettier-ignore
+interface IezETHHyperdrive is
```
jalextowle, line 9: Casing on the interface and file name. `IEzETHHyperdrive`.

contracts/src/instances/ezeth/ezETHBase.sol
```diff
@@ -0,0 +1,156 @@
...
+import { IERC20 } from "../../interfaces/IERC20.sol";
+import { IHyperdrive } from "../../interfaces/IHyperdrive.sol";
+import { IRestakeManager } from "../../interfaces/IRestakeManager.sol";
+import { HyperdriveBase } from "../../internal/HyperdriveBase.sol";
+import { FixedPointMath, ONE } from "../../libraries/FixedPointMath.sol";
+
+/// @author DELV
+/// @title ezETH Base Contract
+/// @notice The base contract for the ezETH Hyperdrive implementation.
+/// @dev
```
jalextowle, line 13: We're missing the dev comment.

contracts/src/instances/ezeth/ezETHBase.sol
```diff
@@ -0,0 +1,156 @@
...
+
+        // If the user sent more ether than the amount specified, refund the
+        // excess ether.
+        unchecked {
+            refund = msg.value - _baseAmount;
+        }
+
+        uint256 balanceBefore = _ezETH.balanceOf(address(this));
+        // Submit the provided ether to Renzo to be deposited.
+        // a referrer id can be put in here
```
jalextowle, line 60: Move the comment to the top of the code block (right before `uint256 balanceBefore). We should remove the sentence about the referral ID since it isn't being included. We should also explain how we compute shares minted.

contracts/src/instances/ezeth/ezETHBase.sol
```diff
@@ -0,0 +1,156 @@
...
+    /// @param _shareAmount The vault shares amount.
+    /// @return baseAmount The base amount.
+    function _convertToBase(
+        uint256 _shareAmount
+    ) internal view override returns (uint256) {
+        (, , uint256 totalTVL) = _restakeManager.calculateTVLs();
+        uint256 ezETHSupply = _ezETH.totalSupply();
+
+        // Price in ETH / ezETH, does not include eigenlayer points.
+        uint256 sharePrice = totalTVL.divDown(ezETHSupply);
```
jalextowle, line 113: This is one of the culprits for the loose error bounds. We should be using their function: https://github.com/Renzo-Protocol/contracts-public/blob/0530cbd20590677da5fa6fd32a0a8b250a9eb028/contracts/Oracle/RenzoOracle.sol#L134.

contracts/src/instances/ezeth/ezETHBase.sol
```diff
@@ -0,0 +1,156 @@
...
+    /// @param _baseAmount The base amount.
+    /// @return shareAmount The vault shares amount.
+    function _convertToShares(
+        uint256 _baseAmount
+    ) internal view override returns (uint256) {
+        (, , uint256 totalTVL) = _restakeManager.calculateTVLs();
+        uint256 ezETHSupply = _ezETH.totalSupply();
+
+        // Price in ETH / ezETH, does not include eigenlayer points.
+        uint256 sharePrice = totalTVL.divDown(ezETHSupply);
```
jalextowle, line 128: This is one of the culprits for the loose error bounds. We should be using their function: https://github.com/Renzo-Protocol/contracts-public/blob/0530cbd20590677da5fa6fd32a0a8b250a9eb028/contracts/Oracle/RenzoOracle.sol#L111.

contracts/src/instances/ezeth/ezETHBase.sol
```diff
@@ -0,0 +1,156 @@
...
+    IERC20 internal immutable _ezETH;
+
+    /// @notice Instantiates the ezETH Hyperdrive base contract.
+    /// @param __restakeManager The Renzo Restakemanager contract.
+    constructor(IRestakeManager __restakeManager) {
+        _restakeManager = __restakeManager;
+        (, bytes memory data) = address(__restakeManager).call(
+            abi.encodeWithSignature("ezETH()")
+        );
+        _ezETH = IERC20(abi.decode(data, (address)));
```
jalextowle, line 33: Same here. We'll want to use the stubbed `ezETH` function.

contracts/src/interfaces/IRestakeManager.sol
```diff
@@ -0,0 +1,57 @@
...
+     * The collateral token specified must be pre-configured to be allowed in the protocol
+     * @param   _collateralToken  The address of the collateral ERC20 token to deposit
+     * @param   _amount The amount of the collateral token to deposit in base units
+     * @param   _referralId The referral ID to use for the deposit (can be 0 if none)
+     */
+    function deposit(
+        IERC20 _collateralToken,
+        uint256 _amount,
+        uint256 _referralId
+    ) external;
```
jalextowle, line 33: We don't use these functions, so we can get rid of them IMO.

contracts/src/interfaces/IRestakeManager.sol
```diff
@@ -0,0 +1,57 @@
...
+     */
+    function depositETH() external payable;
+
+    /**
+     * @notice  Allows a user to deposit ETH into the protocol and get back ezETH
+     * @dev     The amount of ETH sent into this function will be sent to the deposit queue to be
+     * staked later by a validator.  Once staked it will be deposited into EigenLayer.
+     * * @param   _referralId  The referral ID to use for the deposit (can be 0 if none)
+     */
+    function depositETH(uint256 _referralId) external payable;
```
jalextowle, line 47: Same here, we don't provide a referral ID, so why don't we get rid of this function?

test/instances/ezETH/EzETHHyperdrive.t.sol
```diff
@@ -0,0 +1,1081 @@
...
+        // receive LP shares totaling the amount of shares that he contributed
+        // minus the shares set aside for the minimum share reserves and the
+        // zero address's initial LP contribution.
+        assertApproxEqAbs(
+            hyperdrive.balanceOf(AssetId._LP_ASSET_ID, bob),
+            contribution.divDown(
+                hyperdrive.getPoolConfig().initialVaultSharePrice
+            ) - 2 * hyperdrive.getPoolConfig().minimumShareReserves,
+            // TODO: verify tolerance OK
+            1e6 // was 1e5
```
jalextowle, line 339: We should re-evaluate this tolerance when the conversion functions are updated.

test/instances/ezETH/EzETHHyperdrive.t.sol
```diff
@@ -0,0 +1,1081 @@
...
+        verifyFactoryEvents(
+            deployerCoordinator,
+            hyperdrive,
+            bob,
+            contribution,
+            FIXED_RATE,
+            true,
+            config.minimumShareReserves,
+            new bytes(0),
+            // NOTE: Tolerance since ezETH uses mulDivDown for share calculations.
```
jalextowle, line 366: Pretty much every yield source uses `mulDivDown` FWIW. This isn't a problem in other integrations. Let's re-evaluate this after the conversion functions are fixed in `HyperdriveBase`.

contracts/src/interfaces/IezETHHyperdriveRead.sol
sentilesdal: changed to `renzo()`

test/instances/ezETH/EzETHHyperdrive.t.sol
```diff
@@ -0,0 +1,1081 @@
...
+        // Accrue interest in Renzo. Since the share price is given by
+        // `RESTAKE_MANAGER.calculateTVLs() / EZETH.totalSupply()`, we can simulate the
+        // accrual of interest by adding to the balance of the DepositQueue contract.
+        // RestakeManager adds the balance of the DepositQueue to totalTVL in calculateTVLs()
+        uint256 adjustedVariableRate = uint256(variableRate).mulDivDown(
+            POSITION_DURATION_2_WEEKS,
+            365 days
+        );
+        uint256 ethToAdd = totalTVLBefore.mulDown(adjustedVariableRate);
+
```
jalextowle, line 1052: Nit: Remove this line or add a comment.

```suggestion
```

test/instances/ezETH/EzETHHyperdrive.t.sol
```diff
@@ -0,0 +1,1081 @@
...
+        // assertApproxEqAbs(
+        //     EZETH.balanceOf(address(hyperdrive)),
+        //     hyperdriveBalancesBefore.ezethBalance - baseProceeds,
+        //     1
+        // );
+        // assertApproxEqAbs(
+        //     EZETH.balanceOf(trader),
+        //     traderBalancesBefore.ezethBalance + baseProceeds,
+        //     1
+        // );
```
jalextowle, line 1000: Yeah, this can be removed.

test/instances/ezETH/EzETHHyperdrive.t.sol
```diff
@@ -0,0 +1,1081 @@
...
+            //     EZETH.balanceOf(address(hyperdrive)),
+            //     hyperdriveBalancesBefore.ezethBalance + basePaid,
+            //     1
+            // );
+            // ezETH doesn't have shares/balance like steth does
+            // assertApproxEqAbs(
+            //     EZETH.balanceOf(trader),
+            //     traderBalancesBefore.ezethBalance - basePaid,
+            //     1
+            // );
```
jalextowle, line 947: Same here, removing is fine.

contracts/src/interfaces/IezETHHyperdriveCore.sol
sentilesdal: ok so 
```ts
// prettier-ignore
interface IEzETHHyperdrive is
    IHyperdrive,
    IEzETHHyperdriveRead,
    IEzETHHyperdriveCore
{}
```
would become
```ts
// prettier-ignore
interface IEzETHHyperdrive is
    IHyperdrive,
    IEzETHHyperdriveRead,
    IHyperdriveCore
{}
```

test/instances/ezETH/EzETHHyperdrive.t.sol
```diff
@@ -0,0 +1,1081 @@
...
+            shortAmount,
+            false
+        );
+        (, , uint256 totalPooledEther) = RESTAKE_MANAGER.calculateTVLs();
+        uint256 totalShares = EZETH.totalSupply();
+        uint256 baseProceeds = shareProceeds.mulDivDown(
+            totalPooledEther,
+            totalShares
+        );
+        // TODO: make sure this is OK.  There is a lot of rounding that happens in calculateTVLs
```
jalextowle, line 816: We'll need to reevaluate all of the tolerances in this file once the conversion functions are updated. It would be good to comb through and reset these to the old values before another review pass.

contracts/src/interfaces/IezETHHyperdriveCore.sol
jalextowle: `IHyperdriveCore` is part of `IHyperdrive`, so you can remove that line.

contracts/src/interfaces/IDepositQueue.sol
```diff
@@ -0,0 +1,8 @@
+// SPDX-License-Identifier: BUSL-1.1
```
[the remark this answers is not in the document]
sentilesdal: Good catch, wasn't sure about this since it came from renzo.  Should we keep it @jalextowle?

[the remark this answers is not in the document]
sentilesdal: In my experience you leave it the same from where you got it if you can.

contracts/src/deployers/ezeth/EzETHHyperdriveDeployerCoordinator.sol
sentilesdal: I did a pass and I think I got all these.  I finished with the nits and now I'll start working on the updated share price and I'll do another pass to double check this.

contracts/src/interfaces/IDepositQueue.sol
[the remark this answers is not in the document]
jalextowle: We don't want a BUSL license in the repo. To avoid this, I'd say we consolidate all of this into an `IRenzo` interface that includes all of the functions. That way, they didn't write the code -- we did.

contracts/src/instances/ezeth/ezETHBase.sol
sentilesdal: I'm investigating this right now, but that function `calculateRedeemAmount` is for redeeming ezETH to ETH, which we don't do on our side so I don't think that's the right equation.  The equation above it, `calculateMintAmount` is used in `depositETH` already and I just grab the amount minted by diffing the balances.  I've been using their implementation here:
https://github.com/Renzo-Protocol/contracts-public/blob/0530cbd20590677da5fa6fd32a0a8b250a9eb028/contracts/RateProvider/BalancerRateProvider.sol#L27
as an example of how to calculate the price.  They don't use Fixedpoint math here so there could be truncation issues but that shouldn't cause the large differences we're seeing.

sentilesdal: Ah ok I found the issue!  Their calculateMintAmount adjusted for inflation percentage, which I wasn't doing in the test.  After applying to a couple spots in the tests I was able to get the values to match.  I'm updating now!

contracts/src/instances/ezeth/ezETHBase.sol
```diff
@@ -0,0 +1,175 @@
...
+            revert IHyperdrive.TransferFailed();
+        }
+
+        // If the user sent more ether than the amount specified, refund the
+        // excess ether.
+        unchecked {
+            refund = msg.value - _baseAmount;
+        }
+
+        // Submit the provided ether to Renzo to be deposited.  sharesMinted is
```
jalextowle, line 66: Nit: 
```suggestion
        // Submit the provided ether to Renzo to be deposited. `sharesMinted` is
```

contracts/src/instances/ezeth/ezETHHyperdrive.sol
```diff
@@ -0,0 +1,85 @@
...
+/// SSSSSSSS                                                                SSSSSSSS
+///
+/// @author DELV
+/// @title EzETHHyperdrive
+/// @notice A Hyperdrive instance that uses ezETH as the yield source.
+/// @custom:disclaimer The language used in this code is for coding convenience
+///                    only, and is not intended to, and does not, have any
+///                    particular legal or regulatory significance.
+contract EzETHHyperdrive is Hyperdrive, EzETHBase {
+    address constant ETH = 0xEeeeeEeeeEeEeeEeEeEeeEEEeeeeEeeeeeeeEEeE;
```
jalextowle, line 57: We already have a constant for this in `contracts/src/libraries/Constants.sol`. Let's use that instead (see the `StETHHyperdrive` implementation.

contracts/src/instances/ezeth/ezETHBase.sol
```diff
@@ -0,0 +1,175 @@
+// SPDX-License-Identifier: Apache-2.0
```
jalextowle, line 1: This filename needs to be changed. I see several others like this too. Anywhere that we have `ezETHBase` needs to be updated to `EzETHBase`. You might have to use `git mv` for this since MacOS has a case-insensitive file-system.

test/utils/HyperdriveTest.sol
```diff
@@ -31,7 +31,7 @@ contract HyperdriveTest is IHyperdriveEvents, BaseTest {
     uint256 internal constant MINIMUM_SHARE_RESERVES = ONE;
     uint256 internal constant MINIMUM_TRANSACTION_AMOUNT = 0.001e18;
     uint256 internal constant CHECKPOINT_DURATION = 1 days;
-    uint256 internal constant POSITION_DURATION = 365 days;
+    uint256 internal constant POSITION_DURATION = 15 days;
```
jalextowle, line 34: Woah, why are we changing this globally? We just want to change this for the test that you're using, not for all of the tests.

test/instances/ezETH/EzETHHyperdrive.t.sol
```diff
@@ -0,0 +1,1110 @@
...
+        uint256 hyperdriveSharesBefore = EZETH.balanceOf(address(hyperdrive));
+        openLong(bob, basePaid);
+
+        // NOTE: using the vaultSharePrice has somewhat higher tolerance.  If
+        // calculate with expectedShares(), we'll match exactly, but we're
+        // testing the vaultSharePrice here.
+        assertApproxEqAbs(
+            EZETH.balanceOf(address(hyperdrive)),
+            hyperdriveSharesBefore + basePaid.divDown(vaultSharePrice),
+            // TODO: verify tolerance ok, vaultSharePrice is pretty inaccurate
```
jalextowle, line 429: Why is it any less accurate than Lido's share price?

test/instances/ezETH/EzETHHyperdrive.t.sol
```diff
@@ -0,0 +1,1110 @@
...
+        // Return the rate
+        uint256 rate = (10 ** 18 * totalTVL) / totalSupply;
+        //*********/
+
+        // My implementation
+        // (, , totalPooledEther) = RESTAKE_MANAGER.calculateTVLs();
+        // totalShares = EZETH.totalSupply();
+        // sharePrice = totalPooledEther.divDown(totalShares);
+
+        return (rate, totalTVL, totalSupply);
```
jalextowle, line 1095: We should just call `renzoOracle.calculateMintAmount` with `ONE` provided, since that's what Hyperdrive does. I think the comments make sense as a way of getting to the answer, but I don't think we need them when we merge.

test/instances/ezETH/EzETHHyperdrive.t.sol
```diff
@@ -0,0 +1,1110 @@
...
+            false
+        );
+        (, uint256 totalPooledEther, uint256 totalShares) = getSharePrice();
+        uint256 baseProceeds = shareProceeds.mulDivDown(
+            totalPooledEther,
+            totalShares
+        );
+        // TODO: make sure this is OK.  vaultSharePrice is pretty inaccurate.
+        assertLe(baseProceeds, expectedBaseProceeds + 1e4);
+        assertApproxEqAbs(baseProceeds, expectedBaseProceeds, 1e5); // was 100
```
jalextowle, line 847: I'm not sure if this is okay. These tolerances seem pretty high, but it feels unavoidable because of the way that they do their share price calculations: https://github.com/Renzo-Protocol/contracts-public/blob/0530cbd20590677da5fa6fd32a0a8b250a9eb028/contracts/Oracle/RenzoOracle.sol#L111. 

I spent some time seeing if I could eliminate the tolerances, and I couldn't because their share price logic is inherently imprecise. I'm not sure why these chose this calculation, but it feels like we're stuck with it. 

With all this said, I think we should do some more rigorous testing of this yield source over the course of the audit. I'm fine with removing the TODOs and merging this for now though.

contracts/src/interfaces/IDepositQueue.sol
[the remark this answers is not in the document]
sentilesdal: took that approach and moved everything to IRenzo.sol

test/instances/ezETH/EzETHHyperdrive.t.sol
sentilesdal: per the following comments about tolerance, resolving this.

test/instances/ezETH/EzETHHyperdrive.t.sol
sentilesdal: per following comments, resolving this one

test/instances/ezETH/EzETHHyperdrive.t.sol
sentilesdal: This was a nice cleanup, thanks.

test/utils/HyperdriveTest.sol
sentilesdal: Oh sorry meant to undo that, was sanity checking the interest accrued.

test/instances/ezETH/EzETHHyperdrive.t.sol
sentilesdal: Are you asking me explain in the comment or here?  The reason AFAIK is that the vaultSharePrice here relys on converting one share to base, and that conversion doesn't account for the inflation like the base -> share conversion and also obviously becomes less accurate for larger conversions.

test/instances/ezETH/EzETHHyperdrive.t.sol
sentilesdal: Roger, calling the renzo oracle's calculateRedeemAmount here, which is what _pricePerVaultShare calls under the hood now.

contracts/src/instances/ezeth/ezETHBase.sol
sentilesdal: Yeah I'm running into that issue with the other renames right now...

test/instances/ezETH/EzETHHyperdrive.t.sol
sentilesdal: Ok sounds good to me, lemme know if you wanna sync up to figure out how to do that.

test/instances/ezETH/EzETHHyperdrive.t.sol
jalextowle: I was asking to explain here, but my later comment supersedes this one. 

My reasoning for why the tolerances are so much higher for this yield source is that the minting calculation is the following:

This is their mint function:

```
        // For first mint, just return the new value added.
        // Checking both current value and existing supply to guard against gaming the initial mint
        if (_currentValueInProtocol == 0 || _existingEzETHSupply == 0) {
            return _newValueAdded; // value is priced in base units, so divide by scale factor
        }
        
        // Calculate the percentage of value after the deposit 
        uint256 inflationPercentaage = SCALE_FACTOR * _newValueAdded / (_currentValueInProtocol + _newValueAdded);

        // Calculate the new supply
        uint256 newEzETHSupply = (_existingEzETHSupply * SCALE_FACTOR) / (SCALE_FACTOR - inflationPercentaage);

        // Subtract the old supply from the new supply to get the amount to mint
        uint256 mintAmount = newEzETHSupply - _existingEzETHSupply;

        // Sanity check
        if(mintAmount == 0) revert InvalidTokenAmount();

        return mintAmount;
```

To analyze this a bit, this reduces to:

$$
\begin{aligned}
    i & = \frac{\Delta z}{z_0 + \Delta z} \\
    s_1 & = \frac{s_0}{1 - i}
\end{aligned}
$$

This looked weird to me, considering that they have a normal share price calculation when redeeming, but when I thought about it, I realized that it's doing the same thing as other yield sources:

$$
s_1 = \frac{s_0}{1 - i} = \frac{s_0}{1 - \tfrac{Delta z}{z_0 + \Delta z}} = \frac{s_0}{\tfrac{z_0}{z_0 + \Delta z}} = s_0 \cdot \frac{z_0 + \Delta z}{z_0}
$$

With this in mind, I think that what's happening is that by compressing the scaling factor into a value between 0 and 1, they lose a lot of precision compared to a normal share price calculation. This leads to an unavoidable loss in precision in our tests. Now that I think about it again this morning, I think this is fine from Hyperdrive's perspective since the imprecision doesn't effect most of Hyperdrive's accounting (Hyperdrive typically operates on the output of deposit, so it would receive the correct value). The one area where we could have issues is in `openShort`: https://github.com/delvtech/hyperdrive/blob/60c789d437ba4e77ce14495f931c45cabd1a643c/contracts/src/internal/HyperdriveShort.sol#L104. Instead of taking the output of `_deposit` as gospel, `openShort` assumes that the amount deposited is correct.

I think the lowest risk/easiest solution to these problems is to disable the ETH -> Renzo minting flow. We can still support this via a zap, but it wouldn't negatively impact Hyperdrive's precision.

contracts/src/deployers/ezeth/EzETHHyperdriveDeployerCoordinator.sol
```diff
@@ -0,0 +1,126 @@
...
+            revert IHyperdrive.UnsupportedToken();
+        }
+        // Otherwise, transfer vault shares from the LP and approve the
+        // Hyperdrive pool.
+        else {
+            ezETH.transferFrom(_lp, address(this), _contribution);
+            ezETH.approve(address(_hyperdrive), _contribution);
+        }
+
+        return value;
```
jrhea, line 84: if you remove this return it will increase the coverage

image

contracts/src/instances/ezeth/EzETHTarget0.sol
```diff
@@ -0,0 +1,44 @@
...
+        IHyperdrive.PoolConfig memory _config,
+        IRestakeManager _restakeManager
+    ) HyperdriveTarget0(_config) EzETHBase(_restakeManager) {}
+
+    /// Extras ///
+
+    /// @notice Returns the Renzo contract.
+    /// @return _restakeManager The Renzo contract.
+    function renzo() external view returns (IRestakeManager) {
+        _revert(abi.encode(_restakeManager));
```
jrhea, line 36: can we add a test that covers this?

contracts/src/instances/ezeth/EzETHBase.sol
```diff
@@ -0,0 +1,152 @@
...
+                totalSupply
+            );
+    }
+
+    /// @dev Gets the total amount of base held by the pool.
+    /// @return baseAmount The total amount of base.
+    function _totalBase() internal pure override returns (uint256) {
+        // NOTE: Since ETH is the base token and can't be swept, we can safely
+        // return zero.
+        return 0;
```
jrhea, line 134: can we get a test that covers this?

contracts/src/instances/ezeth/EzETHBase.sol
```diff
@@ -0,0 +1,152 @@
...
+    /// @dev Gets the total amount of shares held by the pool in the yield
+    ///      source.
+    /// @return shareAmount The total amount of shares.
+    function _totalShares()
+        internal
+        view
+        override
+        returns (uint256 shareAmount)
+    {
+        return _ezETH.balanceOf(address(this));
```
jrhea, line 146: can we get a test that covers this too?

test/instances/ezETH/EzETHHyperdrive.t.sol
sentilesdal: Ok so after our IRL conversation, I've disabled ETH -> ezETH minting flows.  We can reopen this in the future but taking the safe approach for now.

contracts/src/deployers/ezeth/EzETHHyperdriveDeployerCoordinator.sol
sentilesdal: I get solidity warnings when I try to do this

sentilesdal: image

sentilesdal: and it has to have a return type to match the virtual fn its overriding so we're stuck on this one.

sentilesdal: I moved it to the `else` block, should work just as well.

contracts/src/instances/ezeth/EzETHTarget0.sol
sentilesdal: removing instead ;)

contracts/src/instances/ezeth/EzETHBase.sol
sentilesdal: so it turns out _totalBase and _totalShares are only used by the _sweep function, which we don't use here so these are actually safe to remove as well.

contracts/src/instances/ezeth/EzETHBase.sol
sentilesdal: removed instead.

contracts/src/deployers/ezeth/EzETHHyperdriveDeployerCoordinator.sol
```diff
@@ -0,0 +1,125 @@
...
+        if (_options.asBase) {
+            revert IHyperdrive.UnsupportedToken();
+        }
+        // Otherwise, transfer vault shares from the LP and approve the
+        // Hyperdrive pool.
+        else {
+            ezETH.transferFrom(_lp, address(this), _contribution);
+            ezETH.approve(address(_hyperdrive), _contribution);
+            return value;
+        }
```
jalextowle, line 83: Nit: We don't need the else.

```suggestion
        
        // Otherwise, transfer vault shares from the LP and approve the
        // Hyperdrive pool.

        ezETH.transferFrom(_lp, address(this), _contribution);
        ezETH.approve(address(_hyperdrive), _contribution);
        return value;
```

contracts/src/instances/ezeth/EzETHBase.sol
```diff
@@ -0,0 +1,146 @@
...
+    }
+
+    /// Yield Source ///
+
+    /// @dev Accepts a transfer from the user in base token.
+    function _depositWithBase(
+        uint256, // _baseAmount unused
+        bytes calldata // unused
+    ) internal pure override returns (uint256, uint256) {
+        // Price calculations are too inaccurate
```
jalextowle, line 50: We should update the dev comment and remove this comment. Something like this:

```
@dev This option isn't supported because the minting calculation is too imprecise.
```

I can't tell where we need to split the comment here, but make sure that we respect the 80 column width.

contracts/src/instances/ezeth/EzETHBase.sol
```diff
@@ -0,0 +1,146 @@
...
+
+    /// @dev Unused, gets the total amount of shares held by the pool in the yield
+    ///      source.
+    /// @return shareAmount The total amount of shares.
+    function _totalShares()
+        internal
+        view
+        override
+        returns (uint256 shareAmount)
+    {}
```
jalextowle, line 141: Why are these unused? We need both of these to be implemented. Refer to Cash's PRs.

test/instances/ezETH/EzETHHyperdrive.t.sol
```diff
@@ -0,0 +1,1214 @@
...
+import { AssetId } from "contracts/src/libraries/AssetId.sol";
+import { ETH } from "contracts/src/libraries/Constants.sol";
+import { FixedPointMath, ONE } from "contracts/src/libraries/FixedPointMath.sol";
+import { HyperdriveMath } from "contracts/src/libraries/HyperdriveMath.sol";
+import { ERC20ForwarderFactory } from "contracts/src/token/ERC20ForwarderFactory.sol";
+import { ERC20Mintable } from "contracts/test/ERC20Mintable.sol";
+import { HyperdriveTest } from "test/utils/HyperdriveTest.sol";
+import { HyperdriveUtils } from "test/utils/HyperdriveUtils.sol";
+import { Lib } from "test/utils/Lib.sol";
+import "forge-std/console.sol";
```
jalextowle, line 26: Nit: Remove.

jalextowle (commented):
We're going to need another pass.

contracts/src/instances/ezeth/EzETHBase.sol
sentilesdal: Sorry I just changed these because I thought they weren't used, just tracked down that they are used for `sweep()`

sentilesdal: Put these back in.

contracts/src/deployers/ezeth/EzETHHyperdriveDeployerCoordinator.sol
```diff
@@ -0,0 +1,123 @@
...
+        // Otherwise, transfer vault shares from the LP and approve the
+        // Hyperdrive pool.
+        ezETH.transferFrom(_lp, address(this), _contribution);
+        ezETH.approve(address(_hyperdrive), _contribution);
+        return value;
+    }
+
+    /// @dev Allows the contract to receive ether.
+    function _checkMessageValue() internal view override {}
```
jalextowle, line 85: This should be disabled since we disabled deposits. We should revert with `NotPayable`.

contracts/src/instances/ezeth/EzETHBase.sol
```diff
@@ -0,0 +1,152 @@
...
+        view
+        override
+        returns (uint256 shareAmount)
+    {
+        return _ezETH.balanceOf(address(this));
+    }
+
+    /// @dev We override the message value check since this integration is
+    ///      payable.
+    function _checkMessageValue() internal pure override {}
```
jalextowle, line 151: Same here. If we don't accept deposits in ETH, leaving this open is dangerous.

contracts/src/instances/ezeth/EzETHBase.sol
jalextowle: That's why we have the `Sweep.t.sol` sweep for other integrations. These are definitely used (as our recent exchange covers).

contracts/src/instances/ezeth/EzETHBase.sol
```diff
@@ -0,0 +1,152 @@
...
+        _ezETH = IERC20(__restakeManager.ezETH());
+        _renzoOracle = IRenzoOracle(__restakeManager.renzoOracle());
+    }
+
+    /// Yield Source ///
+
+    //// @dev This option isn't supported because the minting calculation is too
+    ///       imprecise.
+    function _depositWithBase(
+        uint256, // _baseAmount unused
```
jalextowle, line 48: Nit: Remove

```suggestion
        uint256, // unused
```

jalextowle (approved):
🏓 One more round of changes. I'm approving, but these need to be addressed before merging

PR #842, opened 2024-03-14
Calculate `normalized_time_remaining` from timestamps

DannyDelott:
This PR unifies the calculation of `normalized_time_remaining` into a `calculate_normalized_time_remaining` method based on current time and maturity time. 

This allows other sdk methods like `close_long()` and `close_short()` to take in current time and maturity time as arguments and return accurate price quotes.

### DIFF
--- a/crates/hyperdrive-math/src/lib.rs
+++ b/crates/hyperdrive-math/src/lib.rs
@@ -143,7 +143,11 @@ impl State {
     }
 
     /// Gets the normalized time remaining
-    fn calculate_time_remaining(&self, maturity_time: U256, current_time: U256) -> FixedPoint {
+    fn calculate_normalized_time_remaining(
+        &self,
+        maturity_time: U256,
+        current_time: U256,
+    ) -> FixedPoint {
         let latest_checkpoint = self.to_checkpoint(current_time);
         let time_remaining = if maturity_time > latest_checkpoint {
             FixedPoint::from(maturity_time - latest_checkpoint)
--- a/crates/hyperdrive-math/src/lib.rs
+++ b/crates/hyperdrive-math/src/lib.rs
@@ -142,6 +142,21 @@ impl State {
         }
     }
 
+    /// Gets the normalized time remaining
+    fn calculate_time_remaining(&self, maturity_time: U256, current_time: U256) -> FixedPoint {
+        let latest_checkpoint = self.to_checkpoint(current_time);
+        let time_remaining = if maturity_time > latest_checkpoint {
+            FixedPoint::from(maturity_time - latest_checkpoint)
+        } else {
+            fixed!(0)
+        };
+
+        // NOTE: Round down to underestimate the time remaining.
+        let time_remaining = time_remaining.div_down(self.position_duration());
+
+        time_remaining
+    }
+
     /// Config ///
 
     fn position_duration(&self) -> FixedPoint {
--- a/crates/hyperdrive-math/src/lib.rs
+++ b/crates/hyperdrive-math/src/lib.rs
@@ -27,6 +27,9 @@ impl Distribution<State> for Standard {
     // TODO: It may be better for this to be a uniform sampler and have a test
     // sampler that is more restrictive like this.
     fn sample<R: Rng + ?Sized>(&self, rng: &mut R) -> State {
+        let one_day_in_seconds = 60 * 60 * 24;
+        let one_hour_in_seconds = 60 * 60;
+
         let config = PoolConfig {
             base_token: Address::zero(),
             vault_shares_token: Address::zero(),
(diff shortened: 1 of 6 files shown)


crates/hyperdrive-math/src/long/fees.rs
```diff
@@ -46,12 +48,41 @@ impl State {
...
+        time_remaining
+    }
+
+    /// Gets the most recent checkpoint time.
+    /// @return latestCheckpoint The latest checkpoint.
+    fn get_latest_checkpoint(
+        &self,
+        current_time: FixedPoint,
+    ) -> FixedPoint {
+        current_time - (current_time % self.checkpoint_duration())
```
DannyDelott, line 86: Wasn't sure how to do modulo with fixed point. This is currently breaking at compile time.

crates/hyperdrive-math/src/long/fees.rs
```diff
@@ -46,12 +51,33 @@ impl State {
...
     }
+
+    fn calculate_time_remaining(
+        &self,
+        maturity_time: U256,
+        current_time: U256 
+    ) -> FixedPoint {
+        let latest_checkpoint = self.to_checkpoint(current_time); 
+        let time_remaining = if maturity_time > latest_checkpoint {
+            fixed!(maturity_time - latest_checkpoint)
```
DannyDelott, line 72: Wasn't sure how to convert this to FixedPoint.

image

crates/hyperdrive-math/src/long/fees.rs
DannyDelott: Fixed, h/t @jalextowle

crates/hyperdrive-math/src/long/fees.rs
DannyDelott: Fixed, h/t @jalextowle

ryangoree (approved):

crates/hyperdrive-math/src/long/close.rs
```diff
@@ -66,9 +68,16 @@ mod tests {
         for _ in 0..*FAST_FUZZ_RUNS {
             let state = rng.gen::<State>();
             let in_ = rng.gen_range(fixed!(0)..=state.effective_share_reserves());
-            let normalized_time_remaining = rng.gen_range(fixed!(0)..=fixed!(1e18));
+            let maturity_time = state.checkpoint_duration();
```
dpaiton, line 71: shouldn't this be `position_duration`?

crates/hyperdrive-math/src/lib.rs
```diff
@@ -142,6 +142,21 @@ impl State {
         }
     }
 
+    /// Gets the normalized time remaining
```
dpaiton, line 145: This seems like a slight variation of the function above it, `time_remaining_scaled`; I think we should just have one of these to avoid confusion.

crates/hyperdrive-math/src/short/close.rs
```diff
@@ -156,10 +158,18 @@ mod tests {
         for _ in 0..*FAST_FUZZ_RUNS {
             let state = rng.gen::<State>();
             let in_ = rng.gen_range(fixed!(0)..=state.bond_reserves());
-            let normalized_time_remaining = rng.gen_range(fixed!(0)..=fixed!(1e18));
+            let maturity_time = state.checkpoint_duration();
```
dpaiton, line 161: again I think we want this to be `position_duration`

dpaiton (approved):
approved pending comments 🚀

crates/hyperdrive-math/src/long/close.rs
dpaiton: assuming the current time is 0; otherwise it should be `now + position_duration`

DannyDelott: ooh yeah `position_duration` makes more sense

crates/hyperdrive-math/src/lib.rs
```diff
@@ -128,18 +134,23 @@ impl State {
         time - time % self.config.checkpoint_duration
     }
 
-    pub fn time_remaining_scaled(
+    /// Gets the normalized time remaining
+    fn calculate_normalized_time_remaining(
```
jalextowle, line 138: This will need to be public for you to be able to use it.

jalextowle (commented):
We'll have to update the visibility, and we should add tests for this `calculate_normalized_time_remaining` (I get that we didn't test this before, but we're working on improving our test coverage. You can add a differential test against `MockHyperdriveMath` which should be pretty easy). Also, let's add a description to the PR.

crates/hyperdrive-math/src/lib.rs
DannyDelott: I don't think we want this as part of the public API. This function is only intended to be called internally by other rust sdk methods.

jalextowle: Yeah makes sense. I left that comment before getting deeper into the PR

crates/hyperdrive-math/src/lib.rs
```diff
@@ -248,3 +255,30 @@ impl YieldSpace for State {
...
+
+    #[tokio::test]
+    async fn test_calculate_normalized_time_remaining() -> Result<()> {
+        // TODO: fuzz test against calculateTimeRemaining in MockHyperdrive.sol
+        let mut rng = thread_rng();
+        let mut state = rng.gen::<State>();
+
+        // Set a snapshot for the values used for calculating normalized time
+        // remaining
+        state.config.position_duration = fixed!(0.000000000028209717e18).into();
```
jalextowle, line 273: You can get rid of the e18 and remove the leading decimal point and zeros.

jalextowle (approved):
There’s one remaining nit, but otherwise LGTM

PR #846, opened 2024-03-15
Annualized Flat Fees

jalextowle:
Fixes: #844.

This PR implements a feature that annualizes the flat fee in the factory before checking it against the minimum and maximum fee parameters. This allows governance to specify reasonable flat fee parameters for all position durations simultaneously.

### DIFF
--- a/contracts/src/factory/HyperdriveFactory.sol
+++ b/contracts/src/factory/HyperdriveFactory.sol
@@ -118,8 +118,20 @@ contract HyperdriveFactory is IHyperdriveFactory {
         ///      deployments.
         uint256 maxTimeStretchAPR;
         /// @dev The lower bound on the fees that can be used in new deployments.
+        /// @dev Most of the fee parameters are used unmodified; however, the
+        ///      flat fee parameter is interpreted as the minimum annualized
+        ///      flat fee. This allows deployers to specify a smaller flat fee
+        ///      than the minimum for terms shorter than a year and ensures that
+        ///      they specify a larger flat fee than the minimum for terms
+        ///      longer than a year.
         IHyperdrive.Fees minFees;
         /// @dev The upper bound on the fees that can be used in new deployments.
+        /// @dev Most of the fee parameters are used unmodified; however, the
+        ///      flat fee parameter is interpreted as the minimum annualized
+        ///      flat fee. This ensures that deployers specify a smaller flat
+        ///      fee than the maximum for terms shorter than a year and allows
+        ///      deployers to specify a larger flat fee than the maximum for
+        ///      terms longer than a year.
         IHyperdrive.Fees maxFees;
         /// @dev The address of the linker factory.
         address linkerFactory;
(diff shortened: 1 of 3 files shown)


contracts/src/factory/HyperdriveFactory.sol
```diff
@@ -871,14 +883,20 @@ contract HyperdriveFactory is IHyperdriveFactory {
...
 
-        // Ensure that the specified fees are within the minimum and maximum fees.
+        // Ensure that the specified fees are within the minimum and maximum
+        // fees. The flat fee is annualized so that it is consistent across all
+        // term lengths.
         if (
             _config.fees.curve > _maxFees.curve ||
-            _config.fees.flat > _maxFees.flat ||
+            // NOTE: Round down to make the check stricter.
+            _config.fees.flat.mulDivDown(365 days, _config.positionDuration) >
```
jalextowle, line 892: Let's say the position duration is `365 days / 2`. If we wanted an annualized flat fee of 5 bps, we'd use `0.00025e18` as the flat fee parameter. When we annualize it here, we will compute:

```
annualizedFlatFee = 0.00025e18.mulDivDown(365 days, 365 days / 2) = 2 * 0.00025e18 = 0.0005e18 = 5 bps
```

contracts/src/factory/HyperdriveFactory.sol
```diff
@@ -118,8 +118,20 @@ contract HyperdriveFactory is IHyperdriveFactory {
...
+        /// @dev Most of the fee parameters are used unmodified; however, the
+        ///      flat fee parameter is interpreted as the minimum annualized
+        ///      flat fee. This allows deployers to specify a smaller flat fee
+        ///      than the minimum for terms shorter than a year and ensures that
+        ///      they specify a larger flat fee than the minimum for terms
+        ///      longer than a year.
         IHyperdrive.Fees minFees;
         /// @dev The upper bound on the fees that can be used in new deployments.
+        /// @dev Most of the fee parameters are used unmodified; however, the
+        ///      flat fee parameter is interpreted as the minimum annualized
```
jrhea, line 130: should this be maximum?

contracts/src/factory/HyperdriveFactory.sol
jrhea: okay so we have to use 2.5bps input for a 5bps annualized fee if the posish duration is 6 months? would it be easier to just like always enter in the annualized one and scale it according to the position duration?

jalextowle: I thought about that. The thing I didn’t love about it is that the config you pass in isn’t the config you get out, so it’s just as much of a footgun as this IMO. That said, I do see the appeal, so I can do this if you feel strongly about it. Something that might be the best of both worlds is passing the flatFeeAPR as a parameter, checking that the flat fee is 0 in the config, and overriding it with the calculated flat fee (similar to what we do with time stretch).

jrhea: it isn't a hill i wanna die on. this is something that we need to make note of though when people deploy

jrhea (approved):

contracts/src/factory/HyperdriveFactory.sol
jalextowle: 100%

contracts/src/factory/HyperdriveFactory.sol
jalextowle: Yep, good catch.

PR #847, opened 2024-03-17
Liveness Fixes

jalextowle:
This PR adds more comprehensive liveness checks throughout the codebase. This will further reduce the chance of arithmetic errors resulting in liveness failures during the checkpoint process.

### DIFF
--- a/contracts/src/interfaces/IHyperdrive.sol
+++ b/contracts/src/interfaces/IHyperdrive.sol
@@ -250,10 +250,6 @@ interface IHyperdrive is
     /// @notice Thrown when the present value calculation fails.
     error InvalidPresentValue();
 
-    /// @notice Thrown when update liquidity brings the share reserves below
-    ///         the minimum share reserves.
-    error InvalidShareReserves();
-
     /// @notice Thrown when an invalid signature is used provide permit access
     ///         to the MultiToken. A signature is considered to be invalid if
     ///         it fails to recover to the owner's address.
@@ -351,6 +347,9 @@ interface IHyperdrive is
     ///         a user attempts to sweep an invalid token. The options and sweep
     ///         targets that are supported vary between instances.
     error UnsupportedToken();
+
+    /// @notice Thrown when `LPMath.calculateUpdateLiquidity` fails.
+    error UpdateLiquidityFailed();
 
     /// Getters ///
 
--- a/contracts/src/libraries/LPMath.sol
+++ b/contracts/src/libraries/LPMath.sol
@@ -51,17 +51,64 @@ library LPMath {
             uint256 bondReserves
         )
     {
+        bool success;
+        (
+            shareReserves,
+            shareAdjustment,
+            bondReserves,
+            success
+        ) = calculateUpdateLiquiditySafe(
+            _shareReserves,
+            _shareAdjustment,
+            _bondReserves,
+            _minimumShareReserves,
+            _shareReservesDelta
+        );
+        if (!success) {
+            revert IHyperdrive.UpdateLiquidityFailed();
+        }
+    }
+
+    /// @dev Calculates the new share reserves, share adjustment, and bond
+    ///      reserves after liquidity is added or removed from the pool. This
+    ///      update is made in such a way that the pool's spot price remains
+    ///      constant.
+    /// @param _shareReserves The current share reserves.
+    /// @param _shareAdjustment The current share adjustment.
+    /// @param _bondReserves The current bond reserves.
+    /// @param _minimumShareReserves The minimum share reserves.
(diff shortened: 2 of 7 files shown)


contracts/src/interfaces/IHyperdrive.sol
```diff
@@ -192,6 +192,9 @@ interface IHyperdrive is
         SolvencyViolated
     }
 
+    // FIXME: Comb through these to see if any of these errors aren't needed
```
jrhea, line 195: can we remove this?

contracts/src/libraries/LPMath.sol
```diff
@@ -50,18 +54,72 @@ library LPMath {
...
-            revert IHyperdrive.InvalidShareReserves();
+        int256 shareReserves_;
+        (shareReserves_, success) = _shareReserves.toInt256Safe();
+        if (!success) {
+            return (0, 0, 0, false);
+        }
+        shareReserves_ += _shareReservesDelta;
+        // NOTE: This is a safe cast since the minimum share reserves must be a
+        // valid uint128.
+        if (shareReserves_ < int256(_minimumShareReserves)) {
```
jrhea, line 121: understood.  normally, it would be good to write a test that verifies that this "NOTE" holds true and the code doesn't change and invalidate this, but this might be difficult

contracts/src/libraries/LPMath.sol
```diff
@@ -76,14 +134,21 @@ library LPMath {
...
             // NOTE: Rounding up to have a larger effective share reserves.
-            shareAdjustment = -shareReserves
+            (shareAdjustment, success) = shareReserves
                 .mulDivUp(uint256(-_shareAdjustment), _shareReserves)
-                .toInt256();
+                .toInt256Safe();
+            if (!success) {
+                return (0, 0, 0, false);
+            }
+            shareAdjustment = -shareAdjustment;
```
jrhea, line 151: this is not not confusing ;)

contracts/src/libraries/YieldSpaceMath.sol
```diff
@@ -567,17 +674,60 @@ library YieldSpaceMath {
     /// @param t The time elapsed since the term's start.
     /// @param c The vault share price.
     /// @param mu The initial vault share price.
-    /// @return The modified YieldSpace Constant.
+    /// @return k The modified YieldSpace Constant, k.
```
jrhea, line 677: i'd just say:

/// @return k The modified YieldSpace Constant.

contracts/test/MockLPMath.sol
```diff
@@ -35,18 +35,22 @@ contract MockLPMath {
...
     ) external pure returns (int256) {
-        return LPMath.calculateNetFlatTrade(_params);
+        (int256 netFlatTrade, bool success) = LPMath.calculateNetFlatTradeSafe(
+            _params
+        );
+        require(!success, "MockLPMath: calculateNetFlatTradeSafe failed");
+        return netFlatTrade;
     }
 
     function calculateNetCurveTrade(
```
jrhea, line 48: should you name this `calculateNetCurveTradeSafe`?

contracts/src/libraries/LPMath.sol
jalextowle: Yeah I get it. It made more sense before when it was all on one line. I'll see if there is a better way to lay this out.

contracts/src/libraries/LPMath.sol
jalextowle: The `NOTE` will hold true since we assert that this is a valid `uint128` in the deployer coordinator. I get what you're saying about testing it though.

contracts/src/libraries/SafeCast.sol
```diff
@@ -2,10 +2,13 @@
 pragma solidity 0.8.20;
 
 import { IHyperdrive } from "../interfaces/IHyperdrive.sol";
+import { Errors } from "./Errors.sol";
```
jrhea, line 5: I know this isn't on you, but it'd be good to get tests for these methods bc rn there are 0.

contracts/src/libraries/LPMath.sol
jalextowle: I added a comment for this.

contracts/src/libraries/SafeCast.sol
jrhea: maybe create an issue. this is something that cash, matt or john could do

jalextowle: I opened an issue. I'll start adding to it over the course of this week.

jrhea (approved):

PR #853, opened 2024-03-19
Add `calculate_idle_share_reserves` to rust sdk

DannyDelott:
This calculation fixes the broken calculation for `Available Liquidity` on the frontend.

### DIFF
--- a/crates/hyperdrive-math/src/lp.rs
+++ b/crates/hyperdrive-math/src/lp.rs
@@ -12,7 +12,7 @@ impl State {
         // NOTE: Round up to underestimate the pool's idle.
         let long_exposure = self.long_exposure().div_up(self.vault_share_price());
 
-        let mut idle_shares_in_base = fixed!(0e18);
+        let mut idle_shares_in_base = fixed!(0);
         if (self.share_reserves() > long_exposure + self.minimum_share_reserves()) {
             idle_shares_in_base =
                 (self.share_reserves() - long_exposure - self.minimum_share_reserves())
--- a/crates/hyperdrive-math/src/lp.rs
+++ b/crates/hyperdrive-math/src/lp.rs
@@ -7,6 +7,21 @@ use fixed_point_macros::{fixed, int256};
 use crate::{State, YieldSpace};
 
 impl State {
+    /// Calculates the number of base that are not reserved by open positions.
+    pub fn calculate_idle_share_reserves_in_base(&self) -> FixedPoint {
+        // NOTE: Round up to underestimate the pool's idle.
+        let long_exposure = self.long_exposure().div_up(self.vault_share_price());
+
+        let mut idle_shares_in_base = fixed!(0e18);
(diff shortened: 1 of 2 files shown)


crates/hyperdrive-math/src/utils.rs
```diff
@@ -98,6 +98,31 @@ pub fn calculate_initial_bond_reserves(
...
+    // NOTE: Round up to underestimate the pool's idle.
+    let long_exposure = long_exposure.div_up(vault_share_price);
+
+    let mut idle_shares_in_base = fixed!(0e18);
+    if (share_reserves > long_exposure + minimum_share_reserves) {
+        idle_shares_in_base = (share_reserves - long_exposure - minimum_share_reserves) * vault_share_price;
+    }
+
+    idle_shares_in_base
+}
```
jalextowle, line 124: We can add this to `hyperdrive-math` since all of these values are in `PoolInfo` and `PoolConfig`. Want to add this to `creates/hyperdrive-math/src/lp.rs`?

crates/hyperdrive-math/src/utils.rs
```diff
@@ -98,6 +98,31 @@ pub fn calculate_initial_bond_reserves(
...
 }
 
+/// @dev Calculates the number of base that are not reserved by
+///      open positions.
+/// @param vault_share_price The current vault share price.
+/// @param share_reserves The current share reserves, from market state
+/// @param long_exposure The current long exposure, from market state
+/// @param minimum_share_reserves The minimum share reserves, from pool config
+/// @return idleShares The amount of shares that are available for LPs to
+///         withdraw.
```
jalextowle, line 108: We don't typically include NatSpec in Rust. When you add this to `lp.rs`, can you follow the comment convention of that file? You can probably just take out the contents of the dev and make that the whole comment.

crates/hyperdrive-math/src/lp.rs
```diff
@@ -7,6 +7,21 @@ use fixed_point_macros::{fixed, int256};
 use crate::{State, YieldSpace};
 
 impl State {
+    /// Calculates the number of base that are not reserved by open positions.
+    pub fn calculate_idle_share_reserves_in_base(&self) -> FixedPoint {
+        // NOTE: Round up to underestimate the pool's idle.
+        let long_exposure = self.long_exposure().div_up(self.vault_share_price());
+
+        let mut idle_shares_in_base = fixed!(0e18);
```
jalextowle, line 15: NIT: Just a heads up that `0e18 = 0`, so we usually just use `fixed!(0)`. 

```suggestion
        let mut idle_shares_in_base = fixed!(0);
```

jalextowle, line 15: Nit: Can we add a comment here? Something like this:

```suggestion
        // Calculate the idle base reserves.
        let mut idle_shares_in_base = fixed!(0e18);
```

jalextowle (approved):
Two nits, but LGTM

PR #855, opened 2024-03-19
Add `renzo()` and `ezETH()` getters

sentilesdal:
Add `renzo()` and `ezETH` getters so that the main entrypoint contract is publicly accessible from the ezETH implementation of hyperdrive.  Note that for this implementation, the vault differs from that of lido in that the token `ezETH` is separate from the vault, `renzo`

### DIFF
--- a/contracts/src/instances/ezeth/EzETHTarget0.sol
+++ b/contracts/src/instances/ezeth/EzETHTarget0.sol
@@ -36,6 +36,12 @@ contract EzETHTarget0 is HyperdriveTarget0, EzETHBase {
         _revert(abi.encode(_restakeManager));
     }
 
+    /// @notice Gets the ezETH token contract.
+    /// @return The ezETH token contract.
+    function ezETH() external view returns (IERC20) {
+        _revert(abi.encode(_restakeManager.ezETH()));
+    }
+
     /// @notice Returns the MultiToken's decimals.
     /// @return The MultiToken's decimals.
     function decimals() external pure override returns (uint8) {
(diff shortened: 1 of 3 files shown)


ryangoree (approved):
LGTM

contracts/src/instances/ezeth/EzETHTarget0.sol
```diff
@@ -30,6 +30,12 @@ contract EzETHTarget0 is HyperdriveTarget0, EzETHBase {
 
     /// Extras ///
 
+    /// @notice Returns the Renzo contract.
+    /// @return _restakeManager The Renzo contract.
+    function renzo() external view returns (IRestakeManager) {
```
jalextowle, line 35: We should add ezETH too IMO. This would publicly display the immutable that is being used in the contract.

contracts/src/instances/ezeth/EzETHTarget0.sol
```diff
@@ -30,6 +30,12 @@ contract EzETHTarget0 is HyperdriveTarget0, EzETHBase {
 
     /// Extras ///
 
+    /// @notice Returns the Renzo contract.
```
jalextowle, line 33: We should add a test for this.

contracts/src/instances/ezeth/EzETHTarget0.sol
sentilesdal: Added ezETH

contracts/src/instances/ezeth/EzETHTarget0.sol
sentilesdal: Added tests for both getters

PR #857, opened 2024-03-19
Add `renzoOracle()` getter

sentilesdal:
- return immutable value
- add renzoOracle getter

### DIFF
--- a/contracts/src/instances/ezeth/EzETHTarget0.sol
+++ b/contracts/src/instances/ezeth/EzETHTarget0.sol
@@ -42,7 +42,7 @@ contract EzETHTarget0 is HyperdriveTarget0, EzETHBase {
         _revert(abi.encode(_ezETH));
     }
 
-    /// @notice Gets the Renzo Oracle contract.
+    /// @notice Gets the RenzoOracle contract.
     /// @return The RenzoOracle contract.
     function renzoOracle() external view returns (IRenzoOracle) {
         _revert(abi.encode(_renzoOracle));
--- a/contracts/src/instances/ezeth/EzETHTarget0.sol
+++ b/contracts/src/instances/ezeth/EzETHTarget0.sol
@@ -37,9 +37,9 @@ contract EzETHTarget0 is HyperdriveTarget0, EzETHBase {
     }
 
     /// @notice Gets the ezETH token contract.
-    /// @return The ezETH token contract.
+    /// @return IERC20 The ezETH token contract.
     function ezETH() external view returns (IERC20) {
-        _revert(abi.encode(_restakeManager.ezETH()));
+        _revert(abi.encode(_ezETH));
     }
 
     /// @notice Returns the MultiToken's decimals.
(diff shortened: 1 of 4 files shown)


contracts/src/instances/ezeth/EzETHTarget0.sol
```diff
@@ -37,9 +37,15 @@ contract EzETHTarget0 is HyperdriveTarget0, EzETHBase {
     }
 
     /// @notice Gets the ezETH token contract.
-    /// @return The ezETH token contract.
+    /// @return IERC20 The ezETH token contract.
```
jalextowle, line 40: IERC20 is a type and shouldn’t be in the NatSpec.

contracts/src/instances/ezeth/EzETHTarget0.sol
```diff
@@ -37,9 +37,15 @@ contract EzETHTarget0 is HyperdriveTarget0, EzETHBase {
...
     function ezETH() external view returns (IERC20) {
-        _revert(abi.encode(_restakeManager.ezETH()));
+        _revert(abi.encode(_ezETH));
+    }
+
+    /// @notice Gets the Renzo Oracle contract.
+    /// @return IRenzoOracle The RenzoOracle contract.
```
jalextowle, line 46: Same here

contracts/src/instances/ezeth/EzETHTarget0.sol
sentilesdal: should this be `address` instead?

jalextowle: If you check out other comments in the repo, you'll see what I mean. There isn't a type in any other `@return` statement.

contracts/src/instances/ezeth/EzETHTarget0.sol
```diff
@@ -31,15 +31,21 @@ contract EzETHTarget0 is HyperdriveTarget0, EzETHBase {
...
 
     /// @notice Gets the ezETH token contract.
     /// @return The ezETH token contract.
     function ezETH() external view returns (IERC20) {
-        _revert(abi.encode(_restakeManager.ezETH()));
+        _revert(abi.encode(_ezETH));
+    }
+
+    /// @notice Gets the Renzo Oracle contract.
+    /// @return The RenzoOracle contract.
```
jalextowle, line 46: Why is it `RenzoOracle` here and `Renzo Oracle` above? Let's pick one?

contracts/src/interfaces/IEzETHHyperdriveRead.sol
```diff
@@ -13,4 +13,8 @@ interface IEzETHHyperdriveRead is IHyperdriveRead {
     /// @notice Gets the ezETH token contract.
     /// @return The ezETH token contract.
     function ezETH() external view returns (IERC20);
+
+    /// @notice Gets the Renzo Oracle contract.
+    /// @return The RenzoOracle contract.
```
jalextowle, line 18: Same here.

PR #863, opened 2024-03-19
Add getters for immutables for LsETH and rETH

cashd:
- Added getter functions for immutables in LsETH/rETH integrations. 
- Added test for getter functions for code coverage.


contracts/src/interfaces/IRethHyperdriveRead.sol
```diff
@@ -0,0 +1,16 @@
+// SPDX-License-Identifier: Apache-2.0
+pragma solidity 0.8.20;
+
+import { IHyperdriveRead } from "./IHyperdriveRead.sol";
+import { IRocketStorage } from "./IRocketStorage.sol";
+import { IRocketTokenRETH } from "./IRocketTokenRETH.sol";
+
+interface IRethHyperdriveRead is IHyperdriveRead {
```
jalextowle, line 8: We should add a `IRethHyperdrive` interface that consumes this if we don't have one already. Also, looking at this again, I'm wondering if `RETH` would be more consistent in these names than `Reth`. We should consider changing this everywhere.

contracts/src/interfaces/lseth/ILsETHHyperdriveRead.sol
```diff
@@ -0,0 +1,11 @@
+// SPDX-License-Identifier: Apache-2.0
+pragma solidity 0.8.20;
+
+import { IHyperdriveRead } from "../IHyperdriveRead.sol";
+import { IRiverV1 } from "./IRiverV1.sol";
+
+interface ILsETHHyperdriveRead is IHyperdriveRead {
```
jalextowle, line 7: Same here, we should add a `ILsETHHyperdrive` interface that consumes this. Check out `IStETHHyperdrive` for an example.

contracts/src/interfaces/IRethHyperdriveRead.sol
```diff
@@ -0,0 +1,16 @@
+// SPDX-License-Identifier: Apache-2.0
+pragma solidity 0.8.20;
+
+import { IHyperdriveRead } from "./IHyperdriveRead.sol";
+import { IRocketStorage } from "./IRocketStorage.sol";
+import { IRocketTokenRETH } from "./IRocketTokenRETH.sol";
+
+interface IRETHHyperdriveRead is IHyperdriveRead {
```
jalextowle, line 8: This name looks right, but the filename is still `IRethHyperdriveRead`. The MacOS filesystem is case insensitive, so you might need to use `git mv` to actually update the file name.

jalextowle (approved):
LGTM once the remaining file is renamed.

contracts/src/interfaces/IRethHyperdriveRead.sol
cashd: good catch, I had it updated but git did not register!!

Issue #775, opened 2024-02-23
Implement `get_targeted_long`


dpaiton:
[This overleaf document](https://www.overleaf.com/read/jkngspgyqzkf#1de6ce) outlines the logic flow for calculateMaxLong in section 1 and proposes a new algorithm for calculateTargetedLong in section 2. I'm posting this issue to seek feedback on the calculations.

The document or PR #882 should be referenced for an up-to-date full derivation and context.

PR #882, opened 2024-03-22
add get_targeted_long

dpaiton:
addresses https://github.com/delvtech/hyperdrive/issues/775

The solution first attempts to directly compute targeted reserve levels:
For some target rate, $r_t$, the pool share reserves, $z_t$, must be:

$$
z_t = \frac{1}{\mu} \left(
  \frac{k}{\frac{c}{\mu} + \left(
    (r_t t + 1)^{\frac{1}{T}}
  \right)^{1 - T}}
\right)^{\tfrac{1}{1 - T}}
$$

and the pool bond reserves, $y_t$, must be:

$$
y_t = \left(
  \frac{k}{ \frac{c}{\mu} +  \left(
    \left( r_t t + 1 \right)^{\frac{1}{T}}
  \right)^{1-T}}
\right)^{1-T} \left( r_t t + 1 \right)^{\frac{1}{T}}
$$

where $c$ is the vault share price, $\mu$ is the initial vault share price, $t$ is the annualized position duration, and $T$ is the time stretch constant. The corresponding deltas are then:

$$
\Delta x = c * (z_t - z_e)
$$

$$
\Delta y = y - y_t - c(\Delta x)
$$

If that is not successful, then it performs newtons method to minimize

$$
l(x) = r(x) - r_t
$$

for some realized fixed rate, $r(x) = (1 - p(x)) / (p(x) t)$ after a long was opened with $x$ base, where $p(x)$ is the spot price after the long was opened.

### DIFF
--- a/crates/hyperdrive-math/src/long/max.rs
+++ b/crates/hyperdrive-math/src/long/max.rs
@@ -340,7 +340,7 @@ impl State {
     /// It's possible that the pool is insolvent after opening a long. In this
     /// case, we return `None` since the fixed point library can't represent
     /// negative numbers.
-    pub fn solvency_after_long(
+    pub(super) fn solvency_after_long(
         &self,
         base_amount: FixedPoint,
         bond_amount: FixedPoint,
@@ -380,7 +380,10 @@ impl State {
     /// This derivative is negative since solvency decreases as more longs are
     /// opened. We use the negation of the derivative to stay in the positive
     /// domain, which allows us to use the fixed point library.
-    pub fn solvency_after_long_derivative(&self, base_amount: FixedPoint) -> Option<FixedPoint> {
+    pub(super) fn solvency_after_long_derivative(
+        &self,
+        base_amount: FixedPoint,
+    ) -> Option<FixedPoint> {
         let maybe_derivative = self.long_amount_derivative(base_amount);
         maybe_derivative.map(|derivative| {
             (derivative
@@ -419,7 +422,7 @@ impl State {
     /// $$
     /// c'(x) = \phi_{c} \cdot \left( \tfrac{1}{p} - 1 \right)
     /// $$
-    pub fn long_amount_derivative(&self, base_amount: FixedPoint) -> Option<FixedPoint> {
+    pub(super) fn long_amount_derivative(&self, base_amount: FixedPoint) -> Option<FixedPoint> {
         let share_amount = base_amount / self.vault_share_price();
         let inner =
             self.initial_vault_share_price() * (self.effective_share_reserves() + share_amount);
--- a/crates/hyperdrive-math/src/long/max.rs
+++ b/crates/hyperdrive-math/src/long/max.rs
@@ -340,7 +340,7 @@ impl State {
     /// It's possible that the pool is insolvent after opening a long. In this
     /// case, we return `None` since the fixed point library can't represent
     /// negative numbers.
-    fn solvency_after_long(
+    pub fn solvency_after_long(
         &self,
         base_amount: FixedPoint,
         bond_amount: FixedPoint,
@@ -380,7 +380,7 @@ impl State {
     /// This derivative is negative since solvency decreases as more longs are
     /// opened. We use the negation of the derivative to stay in the positive
     /// domain, which allows us to use the fixed point library.
-    fn solvency_after_long_derivative(&self, base_amount: FixedPoint) -> Option<FixedPoint> {
+    pub fn solvency_after_long_derivative(&self, base_amount: FixedPoint) -> Option<FixedPoint> {
         let maybe_derivative = self.long_amount_derivative(base_amount);
         maybe_derivative.map(|derivative| {
             (derivative
@@ -419,7 +419,7 @@ impl State {
     /// $$
     /// c'(x) = \phi_{c} \cdot \left( \tfrac{1}{p} - 1 \right)
     /// $$
-    fn long_amount_derivative(&self, base_amount: FixedPoint) -> Option<FixedPoint> {
+    pub fn long_amount_derivative(&self, base_amount: FixedPoint) -> Option<FixedPoint> {
         let share_amount = base_amount / self.vault_share_price();
         let inner =
             self.initial_vault_share_price() * (self.effective_share_reserves() + share_amount);
--- a/crates/hyperdrive-math/src/long/max.rs
+++ b/crates/hyperdrive-math/src/long/max.rs
@@ -624,7 +624,7 @@ mod tests {
             let spot_price_after_long = bob
                 .get_state()
                 .await?
-                .calculate_spot_price_after_long(max_long, None);
+                .calculate_spot_price_after_long(max_long, None)?;
             bob.open_long(max_long, None, None).await?;
 
             // One of three things should be true after opening the long:
--- a/crates/hyperdrive-math/src/long/max.rs
+++ b/crates/hyperdrive-math/src/long/max.rs
@@ -83,7 +83,7 @@ impl State {
             self.max_long_guess(absolute_max_base_amount, checkpoint_exposure);
         let mut maybe_solvency = self.solvency_after_long(
             max_base_amount,
-            self.calculate_open_long(max_base_amount),
+            self.calculate_open_long(max_base_amount).unwrap(),
             checkpoint_exposure,
         );
         if maybe_solvency.is_none() {
@@ -121,7 +121,7 @@ impl State {
             let possible_max_base_amount = max_base_amount + solvency / maybe_derivative.unwrap();
             maybe_solvency = self.solvency_after_long(
                 possible_max_base_amount,
-                self.calculate_open_long(possible_max_base_amount),
+                self.calculate_open_long(possible_max_base_amount).unwrap(),
                 checkpoint_exposure,
             );
             if let Some(s) = maybe_solvency {
--- a/crates/hyperdrive-math/src/long/open.rs
+++ b/crates/hyperdrive-math/src/long/open.rs
@@ -38,8 +38,6 @@ impl State {
         let ending_spot_price =
             self.calculate_spot_price_after_long(base_amount, long_amount.into());
         let max_spot_price = self.calculate_max_spot_price();
-        println!("ending_spot_price {:#?}", ending_spot_price);
-        println!("max_spot_price {:#?}", max_spot_price);
         if ending_spot_price > max_spot_price {
             return Err(eyre!(
                 "calculate_open_long: InsufficientLiquidity: Negative Interest",
@@ -55,17 +53,17 @@ impl State {
         &self,
         base_amount: FixedPoint,
         bond_amount: Option<FixedPoint>,
-    ) -> FixedPoint {
+    ) -> Result<FixedPoint> {
         let bond_amount = match bond_amount {
             Some(bond_amount) => bond_amount,
-            None => self.calculate_open_long(base_amount),
+            None => self.calculate_open_long(base_amount)?,
         };
         let mut state: State = self.clone();
         state.info.bond_reserves -= bond_amount.into();
         state.info.share_reserves += (base_amount / state.vault_share_price()
             - self.open_long_governance_fee(base_amount) / state.vault_share_price())
         .into();
-        state.calculate_spot_price()
+        Ok(state.calculate_spot_price())
     }
 
     /// Calculate the spot rate after a long has been opened.
@@ -74,11 +72,11 @@ impl State {
         &self,
         base_amount: FixedPoint,
         bond_amount: Option<FixedPoint>,
-    ) -> FixedPoint {
-        calculate_rate_given_fixed_price(
-            self.calculate_spot_price_after_long(base_amount, bond_amount),
+    ) -> Result<FixedPoint> {
+        Ok(calculate_rate_given_fixed_price(
+            self.calculate_spot_price_after_long(base_amount, bond_amount)?,
             self.position_duration(),
-        )
+        ))
     }
 }
 
--- a/crates/hyperdrive-math/src/long/open.rs
+++ b/crates/hyperdrive-math/src/long/open.rs
@@ -1,3 +1,4 @@
+use eyre::{eyre, Result};
 use fixed_point::FixedPoint;
 
 use crate::{calculate_rate_given_fixed_price, State, YieldSpace};
@@ -22,7 +23,7 @@ impl State {
     ///                \right) \right)^{1 - t_s}
     ///            \right)^{\tfrac{1}{1 - t_s}}
     /// $$
-    pub fn calculate_open_long<F: Into<FixedPoint>>(&self, base_amount: F) -> FixedPoint {
+    pub fn calculate_open_long<F: Into<FixedPoint>>(&self, base_amount: F) -> Result<FixedPoint> {
         let base_amount = base_amount.into();
 
         if base_amount < self.config.minimum_transaction_amount.into() {
@@ -37,12 +38,15 @@ impl State {
         let ending_spot_price =
             self.calculate_spot_price_after_long(base_amount, long_amount.into());
         let max_spot_price = self.calculate_max_spot_price();
+        println!("ending_spot_price {:#?}", ending_spot_price);
+        println!("max_spot_price {:#?}", max_spot_price);
         if ending_spot_price > max_spot_price {
-            // TODO would be nice to return a `Result` here instead of a panic.
-            panic!("InsufficientLiquidity: Negative Interest");
+            return Err(eyre!(
+                "calculate_open_long: InsufficientLiquidity: Negative Interest",
+            ));
         }
 
-        long_amount - self.open_long_curve_fees(base_amount)
+        Ok(long_amount - self.open_long_curve_fees(base_amount))
     }
 
     /// Calculates the spot price after opening a Hyperdrive long.
--- a/crates/hyperdrive-math/src/long/open.rs
+++ b/crates/hyperdrive-math/src/long/open.rs
@@ -36,7 +36,7 @@ impl State {
 
         // Throw an error if opening the long would result in negative interest.
         let ending_spot_price =
-            self.calculate_spot_price_after_long(base_amount, long_amount.into());
+            self.calculate_spot_price_after_long(base_amount, long_amount.into())?;
         let max_spot_price = self.calculate_max_spot_price();
         if ending_spot_price > max_spot_price {
             return Err(eyre!(
@@ -82,7 +82,6 @@ impl State {
 
 #[cfg(test)]
 mod tests {
-    use eyre::Result;
     use fixed_point_macros::fixed;
     use rand::{thread_rng, Rng};
     use test_utils::{
@@ -126,7 +125,7 @@ mod tests {
             let expected_spot_price = bob
                 .get_state()
                 .await?
-                .calculate_spot_price_after_long(base_paid, None);
+                .calculate_spot_price_after_long(base_paid, None)?;
 
             // Open the long.
             bob.open_long(base_paid, None, None).await?;
@@ -190,7 +189,7 @@ mod tests {
             let expected_spot_rate = bob
                 .get_state()
                 .await?
-                .calculate_spot_rate_after_long(base_paid, None);
+                .calculate_spot_rate_after_long(base_paid, None)?;
 
             // Open the long.
             bob.open_long(base_paid, None, None).await?;
--- a/crates/hyperdrive-math/src/long/targeted.rs
+++ b/crates/hyperdrive-math/src/long/targeted.rs
@@ -1,0 +1,233 @@
+use ethers::{providers::maybe, types::I256};
+use fixed_point::FixedPoint;
+use fixed_point_macros::fixed;
+
+use crate::{State, YieldSpace};
+
+impl State {
+    /// Gets a target long that can be opened given a budget to achieve a desired fixed rate.
+    pub fn get_targeted_long<F: Into<FixedPoint>, I: Into<I256>>(
+        &self,
+        budget: F,
+        target_rate: F,
+        checkpoint_exposure: I,
+        maybe_max_iterations: Option<usize>,
+    ) -> FixedPoint {
+        let budget = budget.into();
+        let target_rate = target_rate.into();
+        let checkpoint_exposure = checkpoint_exposure.into();
+
+        // Estimate the long that achieves a target rate
+        let (absolute_target_base_amount, absolute_target_bond_amount) =
+            self.absolute_targeted_long(target_rate);
+        // Get the maximum long that brings the spot price to 1.
+        let max_base_amount = self.get_max_long(budget, checkpoint_exposure, maybe_max_iterations);
+        // Ensure that the target is less than the max.
+        let target_base_amount = absolute_target_base_amount.min(max_base_amount);
+        // Verify solvency.
+        if self
+            .solvency_after_long(
+                absolute_target_base_amount,
+                absolute_target_bond_amount,
+                checkpoint_exposure,
+            )
+            .is_some()
+        {
+            return target_base_amount.min(budget);
+        } else {
+            // TODO: Refine using an iterative method
+            panic!("Initial guess in `get_targeted_long` is insolvent.");
+        }
+    }
+
+    /// Calculates the long that should be opened to hit a target interest rate.
+    /// This calculation does not take Hyperdrive's solvency constraints into account and shouldn't be used directly.
+    fn absolute_targeted_long<F: Into<FixedPoint>>(
+        &self,
+        target_rate: F,
+    ) -> (FixedPoint, FixedPoint) {
+        //
+        // TODO: Docstring
+        //
+        let target_rate = target_rate.into();
+        let c_over_mu = self
+            .vault_share_price()
+            .div_up(self.initial_vault_share_price());
+        let scaled_rate = (target_rate.mul_up(self.position_duration()) + fixed!(1e18))
+            .pow(fixed!(1e18) / self.time_stretch());
+        let inner = (self.k_down()
+            / (c_over_mu + scaled_rate.pow(fixed!(1e18) - self.time_stretch())))
+        .pow(fixed!(1e18) / (fixed!(1e18) - self.time_stretch()));
+        let target_share_reserves = inner / self.initial_vault_share_price();
+
+        // Now that we have the target share reserves, we can calculate the
+        // target bond reserves using the formula:
+        //
+        // TODO: docstring
+        //
+        let target_bond_reserves = inner * scaled_rate;
+
+        // The absolute max base amount is given by:
+        //
+        // absolute_target_base_amount = c * (z_t - z)
+        let absolute_target_base_amount =
+            (target_share_reserves - self.effective_share_reserves()) * self.vault_share_price();
+
+        // The absolute max bond amount is given by:
+        //
+        // absolute_target_bond_amount = (y - y_t) - c(x)
+        let absolute_target_bond_amount = (self.bond_reserves() - target_bond_reserves)
+            - self.open_long_curve_fees(absolute_target_base_amount);
+
+        (absolute_target_base_amount, absolute_target_bond_amount)
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use eyre::Result;
+    use rand::{thread_rng, Rng};
+    use test_utils::{
+        agent::Agent,
+        chain::{Chain, TestChain},
+        constants::FUZZ_RUNS,
+    };
+    use tracing_test::traced_test;
+
+    use super::*;
+
+    #[traced_test]
+    #[tokio::test]
+    async fn test_get_targeted_long() -> Result<()> {
+        // Spawn a test chain and create three agents -- Alice, Bob, and Claire. Alice
+        // is funded with a large amount of capital so that she can initialize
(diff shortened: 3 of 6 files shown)


dpaiton:
A detailed derivation of the solution can be found in section 2 of the attached document.
[get_targeted_long.pdf](https://github.com/delvtech/hyperdrive/files/14830017/get_targeted_long.pdf)

crates/hyperdrive-math/src/long/open.rs
```diff
@@ -36,21 +37,29 @@ impl State {
         };
         let max_spot_price = self.get_max_spot_price();
         if ending_spot_price > max_spot_price {
-            // TODO would be nice to return a `Result` here instead of a panic.
-            panic!("InsufficientLiquidity: Negative Interest");
+            return Err(eyre!(
+                "calculate_open_long: InsufficientLiquidity: Negative Interest",
+            ));
```
jalextowle, line 42: 🙏

crates/hyperdrive-math/src/long/open.rs
```diff
@@ -36,21 +37,29 @@ impl State {
...
         }
 
-        long_amount - self.open_long_curve_fees(base_amount)
+        Ok(long_amount - self.open_long_curve_fees(base_amount))
     }
 
     #[deprecated(since = "0.4.0", note = "please use `calculate_open_long` instead")]
     pub fn get_long_amount<F: Into<FixedPoint>>(&self, base_amount: F) -> FixedPoint {
-        self.calculate_open_long(base_amount)
+        self.calculate_open_long(base_amount).unwrap()
```
jalextowle, line 50: Any reason not to have this return a `Result` also? We should at least leave a `TODO`.

crates/hyperdrive-math/src/long/targeted.rs
```diff
@@ -0,0 +1,518 @@
...
+        let base_delta =
+            (share_reserves - self.effective_share_reserves()) * self.vault_share_price();
+        let bond_delta =
+            (self.bond_reserves() - bond_reserves) - self.open_long_curve_fees(base_delta);
+        (base_delta, bond_delta)
+    }
+
+    /// Calculates the long that should be opened to hit a target interest rate.
+    /// This calculation does not take Hyperdrive's solvency constraints or exposure
+    /// into account and shouldn't be used directly.
```
jalextowle, line 324: In addition to not taking Hyperdrive's solvency constraints and exposure into account, it also doesn't consider fees. With this in mind, this could be a source of our error. I haven't looked at the tests yet, but on a first glance, this seems like it will cause problems.

As we spoke about during the initial call about the targeted long whitepaper, we can't actually solve directly for this in the same way that we can solve directly for the max long and max short. Let me know if I'm missing something, but it seems more reasonable to me to use this as a guess and proceed directly to Newton's method.

crates/hyperdrive-math/src/long/targeted.rs
```diff
@@ -0,0 +1,518 @@
...
+        // should always be greater than the spot price.
+        if target_rate > resulting_rate {
+            return Err(eyre!("get_targeted_long: We overshot the zero-crossing.",));
+        }
+        let rate_error = resulting_rate - target_rate;
+
+        // Verify solvency and target rate.
+        if self
+            .solvency_after_long(target_base_delta, target_bond_delta, checkpoint_exposure)
+            .is_some()
```
jalextowle, line 69: If we are insolvent here and the rate is greater than the desired rate, shouldn't we be using `get_max_long`? Since this isn't using solvency as part of the objective function, it looks like this will be susceptible to failing due to solvency constraints. We should consider having a heuristic that we use to invoke `get_max_long`. Since we're frankly not performance constrained at this point, we could just start with that and see if we have hit the rate yet. I see the merits of optimizing this, so I'll let you make that call.

crates/hyperdrive-math/src/long/targeted.rs
```diff
@@ -0,0 +1,518 @@
...
+    /// is the spot price after a long of `delta_base`$= x$ was opened and $t$
+    /// is the annualized position duration, the rate derivative is:
+    ///
+    /// $$
+    /// r'(x) = \frac{(-p'(x) p(x) t - (1 - p(x)) (p'(x) t))}{(p(x) t)^2} //
+    /// r'(x) = \frac{-p'(x)}{t p(x)^2}
+    /// $$
+    ///
+    /// We return $-r'(x)$ because negative numbers cannot be represented by FixedPoint.
+    fn negative_rate_after_long_derivative(
```
jalextowle, line 200: This may just be me, but I read this as "we are computing the derivative of the negative rate," which is a little confusing. This name `rate_after_long_derivative_negation` would be clearer to me, but I'll let you make this call.

crates/hyperdrive-math/src/long/open.rs
dpaiton: If you'd like I can change all the Options to Results, that makes more sense to me I was just trying to follow patterns I found elsewhere.

crates/hyperdrive-math/src/long/targeted.rs
dpaiton: > it also doesn't consider fees.

I've been confused about this, so I might have something wrong here. However, your comment brought to my attention that the docstring itself is inaccurate. This function returns the reserve levels (shares & bonds) for a pool to hit the target rate. I believe it is accurate and would not need to account for fees.

That being said, the output of this function is immediately passed into `trade_deltas_from_reserves`, which converts the values into base & bond _deltas_. These delta amounts do take into account fees, but I suspect the two things must be considered jointly to truly get the long amount? I'd appreciate any guidance for how to communicate this.

> it seems more reasonable to me to use this as a guess and proceed directly to Newton's method.

That's what we are doing -- we take this guess and if it is within the error tolerance then we return, else we use it as the starting point for Newton's method.

crates/hyperdrive-math/src/long/targeted.rs
dpaiton: I'd rather this function throws an error if they do not achieve the target instead of giving the max long. If you'd like, I can modify `get_targeted_long_with_budget` to also include a call to `get_max_long` so that it always returns one of the long to hit the target, the budget balance, or the max long.

crates/hyperdrive-math/src/long/max.rs
```diff
@@ -340,7 +340,7 @@ impl State {
     /// It's possible that the pool is insolvent after opening a long. In this
     /// case, we return `None` since the fixed point library can't represent
     /// negative numbers.
-    fn solvency_after_long(
+    pub fn solvency_after_long(
```
jalextowle, line 343: This comment goes for the other functions in this file. My guess is that you want to use these in the `targeted` long module, but these shouldn't be accessible to someone that is using the SDK because it will clutter up the modules that they will see. A way to accomplish both goals is to add a modifier to `pub`. In this case, I think that something like `pub(super)` is appropriate. This allows the function to be imported within the `long` module, but it is considered private outside of that module.

crates/hyperdrive-math/src/long/targeted.rs
```diff
@@ -0,0 +1,503 @@
...
+    ) -> Result<FixedPoint> {
+        let budget = budget.into();
+        match self.calculate_targeted_long(
+            target_rate,
+            checkpoint_exposure,
+            maybe_max_iterations,
+            maybe_allowable_error,
+        ) {
+            Ok(long_amount) => Ok(long_amount.min(budget)),
+            Err(error) => Err(error),
```
jalextowle, line 31: I'm not sure we want to bubble this up. Let's say that I have `10_000e18` base, and the target long can't be calculated due to solvency constraints. If the max long is `100_000e18`, it would be cumbersome to get an error back from this function and have to call `calculate_max_long` manually. 

What about calling `calculate_max_long` if we hit the error case?

crates/hyperdrive-math/src/long/targeted.rs
```diff
@@ -0,0 +1,503 @@
...
+            let possible_target_bond_delta = self
+                .calculate_open_long(possible_target_base_delta)
+                .unwrap();
+            let resulting_rate =
+                self.rate_after_long(possible_target_base_delta, Some(possible_target_bond_delta))?;
+            if target_rate > resulting_rate {
+                return Err(eyre!("get_targeted_long: We overshot the zero-crossing.",));
+            }
+            let loss = resulting_rate - target_rate;
+            if !(loss < allowable_error) {
```
jalextowle, line 149: IMO it's clearer to use `loss >= allowable_error`.

crates/hyperdrive-math/src/long/targeted.rs
```diff
@@ -0,0 +1,503 @@
...
+                return Err(eyre!(
+                    "get_targeted_long: Unable to find an acceptible loss. Final loss = {}.",
```
jalextowle, line 151: Nit: 

```suggestion
                    "get_targeted_long: Unable to find an acceptable loss. Final loss = {}.",
```

crates/hyperdrive-math/src/long/targeted.rs
```diff
@@ -0,0 +1,503 @@
...
+        base_amount: FixedPoint,
+        bond_amount: FixedPoint,
+    ) -> Result<FixedPoint> {
+        let price = self.calculate_spot_price_after_long(base_amount, Some(bond_amount))?;
+        let price_derivative = self.price_after_long_derivative(base_amount, bond_amount)?;
+        // The actual equation we want to represent is:
+        // r' = -p' / (t p^2)
+        // We can do a trick to return a positive-only version and
+        // indicate that it should be negative in the fn name.
+        Ok(price_derivative / (self.annualized_position_duration() * price.pow(fixed!(2e18))))
```
jalextowle, line 205: Since we're squaring `price`, it would be more accurate and efficient to do something like `price * price`. That way, you don't invoke the `pow` machinery, which can introduce a non-negligible amount of error.

crates/hyperdrive-math/src/long/targeted.rs
```diff
@@ -0,0 +1,503 @@
...
+    /// The price after a long that moves shares by $\Delta z$ and bonds by $\Delta y$
+    /// is equal to $P(\Delta z) = \frac{\mu (z_e + \Delta z)}{y - \Delta y}^T$,
+    /// where $T$ is the time stretch constant and $z_e$ is the initial effective share reserves.
+    /// Equivalently, for some amount of `delta_base`$= x$ provided to open a long,
+    /// we can write:
+    ///
+    /// $$
+    /// p(x) = \frac{\mu (z_e + \frac{x}{c} - g(x) - \zeta)}{y_0 - Y(x)}^{T}
+    /// $$
+    /// where $g(x)$ is the [open_long_governance_fee](long::fees::open_long_governance_fee),
```
jalextowle, line 219: Nit:

```suggestion
    ///
    /// where $g(x)$ is the [open_long_governance_fee](long::fees::open_long_governance_fee),
```

crates/hyperdrive-math/src/long/targeted.rs
```diff
@@ -0,0 +1,503 @@
...
+    /// $Y(x)$ is the [long_amount](long::open::calculate_open_long), and $\zeta$ is the
+    /// zeta adjustment.
+    ///
+    /// To compute the derivative, we first define some auxiliary variables:
+    /// $$
```
jalextowle, line 224: Nit:

```suggestion
    ///
    /// $$
```

crates/hyperdrive-math/src/long/targeted.rs
```diff
@@ -0,0 +1,503 @@
...
+    /// is equal to $P(\Delta z) = \frac{\mu (z_e + \Delta z)}{y - \Delta y}^T$,
+    /// where $T$ is the time stretch constant and $z_e$ is the initial effective share reserves.
+    /// Equivalently, for some amount of `delta_base`$= x$ provided to open a long,
+    /// we can write:
+    ///
+    /// $$
+    /// p(x) = \frac{\mu (z_e + \frac{x}{c} - g(x) - \zeta)}{y_0 - Y(x)}^{T}
+    /// $$
+    /// where $g(x)$ is the [open_long_governance_fee](long::fees::open_long_governance_fee),
+    /// $Y(x)$ is the [long_amount](long::open::calculate_open_long), and $\zeta$ is the
```
jalextowle, line 220: We're opening a long, so we know that the zeta adjustment will be zero. I'd suggest removing this here and in the following calculations to simply the comments.

jalextowle: I'd also say that $\zeta$ is already factored into $z_e$. Maybe I'm missing something, but I don't expect to see $\zeta$ here. $\Delta \zeta$ would make more sense, but it will be zero.

crates/hyperdrive-math/src/long/targeted.rs
```diff
@@ -0,0 +1,503 @@
...
+    /// zeta adjustment.
+    ///
+    /// To compute the derivative, we first define some auxiliary variables:
+    /// $$
+    /// a(x) = \mu (z_e + \frac{x}{c} - g(x) - \zeta) \\
+    /// b(x) = y_0 - Y(x) \\
```
jalextowle, line 226: We don't use $z_{e, 0}$ (we just use $z_e$ in these equations), so it feels somewhat inconsistent to use $y_0$ instead of $y$. My preference would be to include the naught subscript in both places, but we should be consistent regardless.

crates/hyperdrive-math/src/long/targeted.rs
```diff
@@ -0,0 +1,503 @@
...
+    /// v(x) = \frac{a(x)}{b(x)}
+    /// $$
+    ///
+    /// and thus $p(x) = v(x)^T$. Given these, we can write out intermediate derivatives:
+    ///
+    /// $$
+    /// a'(x) = \frac{\mu}{c} - g'(x) \\
+    /// b'(x) = -Y'(x) \\
+    /// v'(x) = \frac{b a' - a b'}{b^2}
```
jalextowle, line 235: Nit: I get what you're saying by eliding the arguments, but I think it would be clearer to use the following

```suggestion
    /// v'(x) = \frac{b(x) a'(x) - a(x) b'(x)}{b(x)^2}
```

crates/hyperdrive-math/src/long/targeted.rs
```diff
@@ -0,0 +1,503 @@
...
+    /// $$
+    ///
+    /// And finally, the price after long derivative is:
+    ///
+    /// $$
+    /// p'(x) = v'(x) T v(x)^(T-1)
```
jalextowle, line 241: IMO these formulas are clearer when we explicitly add multiplication with `\cdot` ($\cdot$). This is also more consistent with the documentation in most of the other significant functions. I also think this makes ascii math clearer. 

I'd say this goes doubly when we have a value like $T$, since it is easily confused with transpose (I get that none of these quantities are vectors or matrices, but it collides in my brain).

crates/hyperdrive-math/src/long/targeted.rs
```diff
@@ -0,0 +1,503 @@
...
+        &self,
+        base_amount: FixedPoint,
+        bond_amount: FixedPoint,
+    ) -> Result<FixedPoint> {
+        // g'(x)
+        let gov_fee_derivative = self.governance_lp_fee()
+            * self.curve_fee()
+            * (fixed!(1e18) - self.calculate_spot_price());
+
+        // a(x) = u (z_e + x/c - g(x) - zeta)
```
jalextowle, line 254: Elsewhere in the code and in the contracts, we use `mu` to represent $\mu$.

crates/hyperdrive-math/src/long/targeted.rs
```diff
@@ -0,0 +1,503 @@
...
+        let inner_numerator = self.mu()
+            * (self.ze() + base_amount / self.vault_share_price()
+                - self.open_long_governance_fee(base_amount)
+                - self.zeta().into());
```
jalextowle, line 258: I could be missing something, but as I said above, I think that this zeta term shouldn't be here since $z_e = z - \zeta$. The tests are probably passing because we don't allow positions to mature prior to using the targeted long. 

We should add some of these tests IMO. You'll want to manually checkpoint to automatically mature these positions since the Rust SDK doesn't handle checkpoints well right now.

crates/hyperdrive-math/src/long/targeted.rs
```diff
@@ -0,0 +1,503 @@
...
+
+        // a'(x) = u / c - g'(x)
+        let inner_numerator_derivative = self.mu() / self.vault_share_price() - gov_fee_derivative;
+
+        // b(x) = y_0 - Y(x)
+        let inner_denominator = self.bond_reserves() - bond_amount;
+
+        // b'(x) = Y'(x)
```
jalextowle, line 266: Isn't $b'(x) = -Y'(x)$? I think it would be good to be explicit about this and make a note that we're not using the negative value for a particular reason.

crates/hyperdrive-math/src/long/targeted.rs
```diff
@@ -0,0 +1,503 @@
...
+        let long_amount_derivative = match self.long_amount_derivative(base_amount) {
+            Some(derivative) => derivative,
+            None => return Err(eyre!("long_amount_derivative failure.")),
+        };
+
+        // v(x) = a(x) / b(x)
+        // v'(x) = ( b(x) * a'(x) + a(x) * b'(x) ) / b(x)^2
```
jalextowle, line 273: The fact that we are adding here is subtle and relates to the above comment. I think it would be good to call out explicitly that we are factoring the negative sign out of $b'(x)$.

crates/hyperdrive-math/src/long/targeted.rs
```diff
@@ -0,0 +1,503 @@
...
+        let inner_derivative = (inner_denominator * inner_numerator_derivative
+            + inner_numerator * long_amount_derivative)
+            / inner_denominator.pow(fixed!(2e18));
```
jalextowle, line 276: Same comment here as elsewhere. It would be more efficient and accurate to do something like `inner_denominator * inner_denominator`.

crates/hyperdrive-math/src/long/targeted.rs
```diff
@@ -0,0 +1,503 @@
...
+    /// constant. By setting these equal we can solve for the pool reserve levels
+    /// as a function of a target rate.
+    ///
+    /// For some target rate, $r_t$, the pool share reserves, $z_t$, must be:
+    ///
+    /// $$
+    /// z_t = \frac{1}{\mu} \left(
+    ///   \frac{k}{\frac{c}{\mu} + \left(
+    ///     (r_t t + 1)^{\frac{1}{T}}
+    ///   \right)^{1 - T}}
```
jalextowle, line 327: IMO, it would be clearer to simplify the exponents to $\frac{1 - T}{T}$ (or $\frac{1 - t_s}{t_s}$ if we switch to the usual notation).

jalextowle: This also minimizes the invocations of `pow`, which reduces error and makes the calculation more efficient.

crates/hyperdrive-math/src/long/targeted.rs
```diff
@@ -0,0 +1,503 @@
...
+        // First get the target share reserves
+        let c_over_mu = self
+            .vault_share_price()
+            .div_up(self.initial_vault_share_price());
+        let scaled_rate = (target_rate.mul_up(self.annualized_position_duration()) + fixed!(1e18))
+            .pow(fixed!(1e18) / self.time_stretch());
+        let target_base_reserves = (self.k_down()
+            / (c_over_mu + scaled_rate.pow(fixed!(1e18) - self.time_stretch())))
+        .pow(fixed!(1e18) / (fixed!(1e18) - self.time_stretch()));
+        let target_share_reserves = target_base_reserves / self.initial_vault_share_price();
```
jalextowle, line 355: This is confusing to me since we can't convert between base and shares using the initial share price. I see why we do this (and I am glad that we don't recompute the bond reserves from scratch!). This is a case in which I'd advocate for calling the `target_base_reserves` something like `inner`. I see the appeal of this naming, but it isn't accurate IMO.

crates/hyperdrive-math/src/long/targeted.rs
```diff
@@ -0,0 +1,503 @@
...
+        // the pool. Bob is funded with a random amount of capital so that we
+        // can test `calculate_targeted_long` when budget is the primary constraint
+        // and when it is not.
+
+        let allowable_solvency_error = fixed!(1e5);
+        let allowable_budget_error = fixed!(1e5);
+        let allowable_rate_error = fixed!(1e10);
+        let num_newton_iters = 3;
+
+        // Initialize a test chain; don't need mocks because we want state updates.
```
jalextowle, line 390:
```suggestion
        // Initialize a test chain. We don't need mocks because we want state updates.
```

crates/hyperdrive-math/src/long/targeted.rs
```diff
@@ -0,0 +1,503 @@
...
+                    target_rate,
+                    new_rate,
+                    abs_error,
+                    allowable_rate_error
+                );
+
+            // Else, we should have undershot,
+            // or by some coincidence the budget was the perfect amount
+            // and we hit the rate exactly.
+            } else {
```
jalextowle, line 486: This is just personal preference, so feel free to disregard. To comment `else` blocks, I think this style is pretty nice:

```
// Comment for if
if blah {
    // blah
}
// Comment for else
else {
    // blah
}
```

crates/hyperdrive-math/src/long/targeted.rs
```diff
@@ -0,0 +1,503 @@
...
+        chain::{Chain, TestChain},
+        constants::FUZZ_RUNS,
+    };
+    use tracing_test::traced_test;
+
+    use super::*;
+
+    #[traced_test]
+    #[tokio::test]
+    async fn test_calculate_targeted_long_with_budget() -> Result<()> {
```
jalextowle, line 378: This test is great. To improve the robustness, we could make some trades before the targeted long is opened. It would be especially helpful to make some trades that could result in a non-trivial $\zeta$ value. If you open a random long or short and advance a full term, you'll have a non-trivial $\zeta$ value.

crates/hyperdrive-math/src/long/targeted.rs
jalextowle: Resolving this since it's outdated. Consider my new review comments to be my current thoughts.

jalextowle (commented):
Excellent job on this! Overall, I found the code to be really well-documented and clear. I went through all of the math, and I worked through the denser areas (like solving for the share and bond reserves implied by a target rate) on the whiteboard. It's fairly late, so it's possible that I missed something, but the implementation looks spot on for the most part. 

There were some stylistic nits and a few areas where I think the documentation can be improved. There were also a few areas that could be optimized (mostly by using `pow` less). I think there's only one real bug that I found, but I could be missing something.

crates/hyperdrive-math/src/long/targeted.rs
dpaiton: Thanks for pointing out this potential source of confusion, I added a comment clarifying $\Delta \zeta$ and I updated the targeted long PDF.

I just went over it again and I believe these equations are correct. The price is computed from the effective share reserves, and the equations are derived wrt the change in shares, $\Delta z$, which is not taking into account the zeta adjustment. Therefore, the equations for $p(x)$ and $p'(x)$ will want to require $z_{e,1} = z_{e,0} + (\Delta z - (\zeta + \Delta \zeta))$. In other words,

$$
z_1 = z_0 + \frac{
    \Delta x - \phi_g \cdot \phi_c \cdot \left( 1 - p \right) \cdot \Delta x
}{c}
$$

$$
\zeta_1 = \zeta_0
$$

and $z_{e,1} = z_{1} - \zeta_{1}$. Therefore,

$$
z_{e,1} = z_0 + \frac{
    \Delta x - \phi_g \cdot \phi_n \cdot \left( 1 - p \right) \cdot \Delta x
}{c} - \zeta_{1}
$$



I agree keeping $\zeta$ around is cumbersome because we're not changing it. However, if a user follows the `calculate_spot_price_after_long` math all the way into `yield_space::calculate_spot_price` they will see that we use `self.ze()`, which computes $z - \zeta$. So leaving it out would be inaccurate.

crates/hyperdrive-math/src/long/targeted.rs
dpaiton: Thanks for pointing this out!

crates/hyperdrive-math/src/long/targeted.rs
dpaiton: This is what I'm doing, I think. Are you suggesting I remove the newline before the else comment?

jalextowle: Oh, I was suggesting that instead of “} else {“ you put the first brace on the line above the comment

crates/hyperdrive-math/src/long/targeted.rs
dpaiton: good catch!

crates/hyperdrive-math/src/long/targeted.rs
dpaiton: I definitely could be missing something, but I think the number of `pow` invocations stays the same because the `target_bond_reserves` do not want that simplification.

Here's a modified version using the simplified exponent, which still has to have 3 `pow` calls.

```rs
        let c_over_mu = self
            .vault_share_price()
            .div_up(self.initial_vault_share_price());
        let scaled_rate = target_rate.mul_up(self.annualized_position_duration()) + fixed!(1e18);
        let inner = (self.k_down()
            / (c_over_mu
                + scaled_rate.pow((fixed!(1e18) - self.time_stretch()) / self.time_stretch())))
        .pow(fixed!(1e18) / (fixed!(1e18) - self.time_stretch()));
        let target_share_reserves = inner / self.initial_vault_share_price();

        // Then get the target bond reserves.
        let target_bond_reserves = inner * scaled_rate.pow(fixed!(1e18) / self.time_stretch());
```

You can see that here:

$$
z_{t} = \tfrac{1}{\mu} \left( \tfrac{k}{ \tfrac{c}{\mu} + \left( (r_{t} t_{d}+1)^{\tfrac{1}{t_{s}}} \right)^{1-t_{s}}} \right)^{\tfrac{1}{1-t_{s}}}
$$

$$
y_{t} = \left( \tfrac{k}{ \tfrac{c}{\mu} +  \left( \left( r_{t} t_{d} + 1 \right)^{\tfrac{1}{t_{s}}} \right)^{1-t_{s}}} \right)^{1-t_{s}} \left( r_{t} t_{d} + 1 \right)^{\tfrac{1}{t_{s}}}
$$

The $y_{t}$ term wants the exponent to be $\tfrac{1}{t_s}$ and the $z_{t}$ term wants the exponent to be $\frac{1 - t_s}{t_s}$. By keeping them separated, I can simplify the `target_bond_reserves` equation.

I think the difference is minor, and I'm happy to go with your suggestion if you still want it.  Unless I missed a way to reduce the `pow`s; if so please let me know.

crates/hyperdrive-math/src/long/targeted.rs
dpaiton: Ahh I think I just got it. This:

$$
z_{e,1} = z_0 + \frac{
    \Delta x - \phi_g \cdot \phi_n \cdot \left( 1 - p \right) \cdot \Delta x
}{c} - \zeta_{1}
$$

is correct, but not what I'm implementing in code. The $- \zeta$ part happens inside `self.ze()`, as I said, and if I do it again here then we subtract it twice.

crates/hyperdrive-math/src/long/targeted.rs
jalextowle: I'm talking about the inner exponents. This term:

$$
\left( \left( 1 + r_t \cdot t_d \right)^{\tfrac{1}{t_s}} \right)^{1 - t_s}
$$

We can combine the two exponents into $\tfrac{1 - t_s}{t_s}$ in both.

crates/hyperdrive-math/src/long/targeted.rs
dpaiton: AHHH ok got it!

crates/hyperdrive-math/src/long/targeted.rs
```diff
@@ -0,0 +1,538 @@
...
+        // g'(x)
+        let gov_fee_derivative = self.governance_lp_fee()
+            * self.curve_fee()
+            * (fixed!(1e18) - self.calculate_spot_price());
+
+        // a(x) = mu * (z_{e,0} + x/c - g(x))
+        let inner_numerator = self.mu()
+            * (self.ze() + base_amount / self.vault_share_price()
+                - self.open_long_governance_fee(base_amount));
+        // - self.zeta().into());
```
jalextowle, line 268: Ok, I don't see the $\zeta$ term here, but it's still in the comments above. Unless there's something wrong with $self.ze()$ we shouldn't need it.

jalextowle (approved):

crates/hyperdrive-math/src/long/targeted.rs
dpaiton: made an issue to support this with a higher-level fn: https://github.com/delvtech/hyperdrive/issues/949

crates/hyperdrive-math/src/long/targeted.rs
dpaiton: resolved in a phone call; same number of pows so I'll leave it as is.

PR #883, opened 2024-03-22
Integration Test Suite

cashd:
# Description

Creates a new testing suite called `IntegrationTest` that generalizes instance setups and deployments. The testing suite is an abstract contract that is implemented in instance test. The implementor overrides two virtual functions and the testing suite properly configures the forked environment. The suite then provides testing coverage for the deployment and initializing steps of the market creation process. 

- test__deployAndInitialize_asShares
- test__deployAndInitialize_asBase

The testing suite is implemented in `StETH`, `LsETH`, `rETH`, and `EzETH` instance test. This step is the first in a series to generalize the entire instance testing process. As of now, the generalization has reduced the amount of code by about 1.5k lines 🙌

# Review Checklists

be removed.

### Solidity

- [x] **Tokens**
    - [x] Do all `approve` calls use `forceApprove`?
    - [x] Do all `transfer` calls use `safeTransfer`?
    - [x] Do all `transferFrom` calls use `msg.sender` as the `from` address?
        - [x] If not, is the function access restricted to prevent unauthorized
              token spend?
    - [x] Is the returned `success` boolean checked to handle failed calls?
    - [x] If using `delegatecall`, are there strict access controls on the
- [x] **Reentrancy**
    - [x] Are functions that make external calls or transfer ether marked as `nonReentrant`?
        - [x] If not, is there documentation that explains why reentrancy is
- [x] **Gas Optimizations**
    - [x] Is the logic as simple as possible?
    - [x] Are the storage values that are used repeatedly cached in stack or
          memory variables?
    - [x] If loops are used, are there guards in place to avoid out-of-gas
          issues?
- [x] **Visibility**
    - [x] Are all `payable` functions restricted to avoid stuck ether?
- [x] **Math**
    - [x] Is all of the arithmetic checked or guarded by if-statements that will
          catch underflows?
    - [x] If `Safe` functions are altered, are potential underflows and
    - [x] Are all of the rounding directions clearly documented?
- [x] **Testing**


test/instances/ezETH/EzETHHyperdrive.t.sol
```diff
@@ -960,7 +603,7 @@ contract EzETHHyperdriveTest is HyperdriveTest {
             totalShares
         );
         assertLe(baseProceeds, expectedBaseProceeds + 1e4);
-        assertApproxEqAbs(baseProceeds, expectedBaseProceeds, 1e5);
+        assertApproxEqAbs(baseProceeds, expectedBaseProceeds, 1e6);
```
cashd, line 606: @sentilesdal I had to up this delta value by one sd 🤔

test/utils/IntegrationTest.sol
```diff
@@ -0,0 +1,415 @@
...
+
+/// @author DELV
+/// @title IntegrationTest
+/// @notice The base contract for the integration testing suite.
+/// @dev A testing suite that provides a foundation to setup, deploy, and
+/// test common cases for Hyperdrive instances.
+/// @custom:disclaimer The language used in this code is for coding convenience
+///                    only, and is not intended to, and does not, have any
+///                    particular legal or regulatory significance.
+abstract contract IntegrationTest is HyperdriveTest {
```
jalextowle, line 23: It would probably be better to call this `InstanceTest` to disambiguate it from the integration tests in `test/integration`.

test/utils/IntegrationTest.sol
```diff
@@ -0,0 +1,415 @@
...
+    using Lib for *;
+    using FixedPointMath for uint256;
+
+    // Fixed rate used to configure market.
+    uint256 internal constant FIXED_RATE = 0.05e18;
+
+    // ETH whale account.
+    address internal ETH_WHALE = 0x00000000219ab540356cBB839Cbe05303d7705Fa;
```
jalextowle, line 31: It doesn't look like this is used anywhere.

test/utils/IntegrationTest.sol
```diff
@@ -0,0 +1,415 @@
...
+        uint256 minTransactionAmount;
+        uint256 positionDuration;
+        bool enableBaseDeposits;
+    }
+
+    IntegrationConfig internal config;
+    IHyperdrive.PoolDeployConfig internal poolConfig;
+    HyperdriveFactory factory;
+    address deployerCoordinator;
+    bool internal immutable isBaseETH;
```
jalextowle, line 48: Let's document these. Should the `poolConfig` be passed in the constructor? I see it referenced below, but it isn't defined in this file. Furthermore, anything that isn't supposed to be accessed outside of this file should be marked `private`.

test/utils/IntegrationTest.sol
```diff
@@ -0,0 +1,415 @@
...
+        deployFactory();
+
+        // Set the deployer coordinator address and add to the factory.
+        deployerCoordinator = deployCoordinator();
+        factory.addDeployerCoordinator(deployerCoordinator);
+
+        // Deploy all Hyperdrive targets using  deployer coordinator contract.
+        deployTargets(
+            bytes32(uint256(0xdeadbeef)), // Deployment Id
+            bytes32(uint256(0xdeadbabe)), // Deployment Salt
```
jalextowle, line 88: We might as well make these constants.

test/utils/IntegrationTest.sol
```diff
@@ -0,0 +1,415 @@
...
+        vm.recordLogs();
+    }
+
+    /// @dev Deploys all Hyperdrive contracts using the
+    ///      deployer coordinator contract.
+    /// @param deploymentId The deployment id.
+    /// @param deploymentSalt The deployment salt for create2.
+    /// @param contribution The amount to initialize the market.
+    /// @param asBase Initialize the market with base token.
+    function deployTargets(
```
jalextowle, line 103: This isn't just deploying the targets. It's also deploying the pool. Why don't we call it `deploy` or `deployHyperdrive`?

test/utils/IntegrationTest.sol
```diff
@@ -0,0 +1,415 @@
...
+        }
+
+        // Deploy the Hyperdrive Factory contract.
+        deployFactory();
+
+        // Set the deployer coordinator address and add to the factory.
+        deployerCoordinator = deployCoordinator();
+        factory.addDeployerCoordinator(deployerCoordinator);
+
+        // Deploy all Hyperdrive targets using  deployer coordinator contract.
```
jalextowle, line 85: This comment is misleading since we also deploy the instance. I'm slightly confused about why we are deploying this considering that we don't do anything with the return value.

test/utils/IntegrationTest.sol
```diff
@@ -0,0 +1,415 @@
...
+            FIXED_RATE,
+            4,
+            deploymentSalt
+        );
+
+        // Alice gives approval to the deployer coordinator to fund the market.
+        config.token.approve(deployerCoordinator, 100_000e18);
+
+        // We expect the deployAndInitialize to fail with an
+        // UnSupported token error if depositing as base is not supported.
```
jalextowle, line 168: Nit: `UnSupported` => `Unsupported`.

test/utils/IntegrationTest.sol
```diff
@@ -0,0 +1,415 @@
...
+            contribution,
+            FIXED_RATE,
+            FIXED_RATE,
+            IHyperdrive.Options({
+                asBase: asBase,
+                destination: alice,
+                extraData: new bytes(0)
+            }),
+            deploymentSalt
+        );
```
jalextowle, line 196: In the current tests, we make assertions about the LP shares received by the LPs and the initial quantity of LP shares. I think it would still be good to do this.

test/utils/IntegrationTest.sol
jalextowle: Ok, I see that the pool config is set using the factory's configuration. Let's make it private in that case.

test/utils/IntegrationTest.sol
```diff
@@ -0,0 +1,415 @@
...
+                    flat: ONE,
+                    governanceLP: ONE,
+                    governanceZombie: ONE
+                }),
+                linkerFactory: address(forwarderFactory),
+                linkerCodeHash: forwarderFactory.ERC20LINK_HASH()
+            })
+        );
+
+        poolConfig = IHyperdrive.PoolDeployConfig({
```
jalextowle, line 240: Nit: Missing comment.

test/utils/IntegrationTest.sol
```diff
@@ -0,0 +1,415 @@
...
+        // This function reverts if base deposits are disabled.
+        deployTargets(
+            bytes32(uint256(0xbeefbabe)), // Deployment Id
+            bytes32(uint256(0xdeadfade)), // Deployment Salt
+            contribution, // Contribution
+            true // asBase
+        );
+
+        // If base deposits are enabled we verify some assertions.
+        if (config.enableBaseDeposits && isBaseETH) {
```
jalextowle, line 303: Shouldn't we have assertions if the base token isn't ETH?

test/utils/IntegrationTest.sol
```diff
@@ -0,0 +1,415 @@
...
+            // If the base token is ETH we assert the ETH balance is correct.
+            if (isBaseETH) {
```
jalextowle, line 305: This is implied by the preceding `&&`. Maybe we should remove the `&& isBaseETH`?

test/utils/IntegrationTest.sol
```diff
@@ -0,0 +1,415 @@
...
+        uint256 contribution = 5_000e18;
+
+        // Get protocol state information used for calculating correct LP shares.
+        (uint256 totalBase, uint256 totalShare, ) = getProtocolSharePrice();
+
+        // Contribution in terms of shares.
+        uint256 contributionShares = contribution.mulDivDown(
+            totalShare,
+            totalBase
+        );
```
jalextowle, line 291: In general, it might be better to use an abstract share conversion function here. It seems like the tolerances work out for now, but some share conversion methods are less precise than others (e.g. Renzo's `calculateMintAmount`).

test/utils/IntegrationTest.sol
jalextowle: It feels like this is mostly done for historical reasons, and unless it's doing something that I'm not picking up on, we could consider getting rid of it.

test/instances/ezETH/EzETHHyperdrive.t.sol
```diff
@@ -54,187 +52,78 @@ contract EzETHHyperdriveTest is HyperdriveTest {
     // mocking.  To test with their deployed code we use a shorter position
     // duration.
     uint256 internal constant POSITION_DURATION_2_WEEKS = 15 days;
```
jalextowle, line 54: Nit: I know you didn't write this, but we may as well rename this to `POSITION_DURATION_15_DAYS`.

test/instances/ezETH/EzETHHyperdrive.t.sol
```diff
@@ -54,187 +52,78 @@ contract EzETHHyperdriveTest is HyperdriveTest {
...
+
+    constructor() IntegrationTest(__testConfig) {}
 
     function setUp() public override __mainnet_fork(STARTING_BLOCK) {
+        // Depositing with ETH is not allowed for this pool so we need to get
+        // some ezETH for alice first.
+        vm.startPrank(EZETH_WHALE);
+
+        vm.deal(EZETH_WHALE, 50_000e18);
+        RESTAKE_MANAGER.depositETH{ value: 40_000e18 }();
```
jalextowle, line 81: What's up with these amounts? Why are we using 50_000e18 for one and 40_000e18 for the other?

test/instances/ezETH/EzETHHyperdrive.t.sol
```diff
@@ -54,187 +52,78 @@ contract EzETHHyperdriveTest is HyperdriveTest {
...
+        vm.stopPrank();
+
         super.setUp();
```
jalextowle, line 84: Nit: Missing comment.

test/instances/ezETH/EzETHHyperdrive.t.sol
```diff
@@ -54,187 +52,78 @@ contract EzETHHyperdriveTest is HyperdriveTest {
...
+        );
+
+    constructor() IntegrationTest(__testConfig) {}
 
     function setUp() public override __mainnet_fork(STARTING_BLOCK) {
+        // Depositing with ETH is not allowed for this pool so we need to get
+        // some ezETH for alice first.
+        vm.startPrank(EZETH_WHALE);
+
+        vm.deal(EZETH_WHALE, 50_000e18);
```
jalextowle, line 80: Nit: Missing comment.

test/instances/reth/RETHHyperdrive.t.sol
```diff
@@ -28,224 +28,78 @@ import { RETHTarget3Deployer } from "contracts/src/deployers/reth/RETHTarget3Dep
...
+            POSITION_DURATION,
+            false
+        );
 
-    HyperdriveFactory factory;
-    address deployerCoordinator;
+    constructor() IntegrationTest(__testConfig) {}
 
     function setUp() public override __mainnet_fork(19_429_100) {
+        // Give the rETH contract ETH to mimic adequate withdrawable liquidity.
```
jalextowle, line 67: This comment doesn't describe the following `super.setUp();`. We should either update this comment to describe both operations or split them out.

test/instances/reth/RETHHyperdrive.t.sol
```diff
@@ -28,224 +28,78 @@ import { RETHTarget3Deployer } from "contracts/src/deployers/reth/RETHTarget3Dep
...
+    /// @dev Fetches share price information about rETH.
+    function getProtocolSharePrice()
+        internal
+        view
+        override
+        returns (uint256, uint256, uint256)
+    {
+        return (
+            rocketTokenRETH.getExchangeRate(),
+            rocketTokenRETH.getExchangeRate(),
```
jalextowle, line 83: Shouldn't these be the total supply / total amount of ETH?

test/instances/steth/StETHHyperdrive.t.sol
```diff
@@ -39,182 +37,64 @@ contract StETHHyperdriveTest is HyperdriveTest {
...
+            1e15,
+            POSITION_DURATION,
+            true
+        );
 
-    HyperdriveFactory factory;
-    address deployerCoordinator;
+    constructor() IntegrationTest(__testConfig) {}
 
     function setUp() public override __mainnet_fork(17_376_154) {
```
jalextowle, line 58: We're missing comments in this function. I know that the tests previously violated the style we've discussed, but let's set good habits and ensure that the style that we're using is consistent between our tests and the source code. I'm not asking you to do anything we don't already do in `contracts/src/`.

test/utils/IntegrationTest.sol
```diff
@@ -0,0 +1,415 @@
...
+import { FixedPointMath, ONE } from "contracts/src/libraries/FixedPointMath.sol";
+import { HyperdriveTest } from "test/utils/HyperdriveTest.sol";
+import { HyperdriveUtils } from "test/utils/HyperdriveUtils.sol";
+import { Lib } from "test/utils/Lib.sol";
+
+/// @author DELV
+/// @title IntegrationTest
+/// @notice The base contract for the integration testing suite.
+/// @dev A testing suite that provides a foundation to setup, deploy, and
+/// test common cases for Hyperdrive instances.
```
jalextowle, line 19: Nit: 

```suggestion
/// @dev A testing suite that provides a foundation to setup, deploy, and
///           test common cases for Hyperdrive instances.
```

test/utils/IntegrationTest.sol
```diff
@@ -0,0 +1,415 @@
...
+    /// @param shareSupply Amount of share tokens in circulation.
+    /// @param sharePrice Price of of one share in terms of base.
+    function getProtocolSharePrice()
+        internal
+        virtual
+        returns (uint256 baseSupply, uint256 shareSupply, uint256 sharePrice);
+
+    /// @dev A virtual function that defines the deployer coordinator
+    ///      contract that will be used to deploy all the instance targets.
+    function deployCoordinator() internal virtual returns (address);
```
jalextowle, line 274: Right now, these functions are sort of buried between non-abstract functions in the test. It would be better IMO to create a section of the functions that need to be overrided after setup. It would be good to separate these sections with comments like:

```
/// Overrides ///
```

and 

```
/// Tests ///
```

or something like that.

test/utils/IntegrationTest.sol
```diff
@@ -0,0 +1,415 @@
...
+                hyperdrive.getPoolConfig().minimumShareReserves,
+                new bytes(0),
+                config.shareTolerance
+            );
+        }
+    }
+
+    /// @dev Test to verify a market can be deployed and initialized funded
+    ///      by the share token.
+    function test__deployAndInitialize__asShares() external {
```
jalextowle, line 356: For the sake of completeness, I'm wondering if we should have a config option about whether or not shares are supported. Right now, all of our yield sources support shares. That said, this may not always be the case, and it would be easier to implement this now than down the road.

test/instances/ezETH/EzETHHyperdrive.t.sol
sentilesdal: Dang ok.

test/utils/InstanceTest.sol
```diff
@@ -0,0 +1,451 @@
...
+        // We expect the deployAndInitialize to fail with an
+        // Unsupported token error if depositing with shares is not supported.
+        if (!config.enableShareDeposits && !asBase) {
+            vm.expectRevert(IHyperdrive.UnsupportedToken.selector);
+        }
+
+        // Deploy and initialize the market. If the base token is ETH we pass the
+        // contribution through the call.
+        hyperdrive = factory.deployAndInitialize{
+            value: asBase && isBaseETH ? contribution : 0
```
jalextowle, line 219: It looks like we aren't testing refunds here. I think it would be good to add a check for this since each deployer coordinator has to implement their own refunds. Handling ETH can be a bit of a footgun.

jalextowle (approved):
Looks great! Nice work @cashd! I just have one more comment about refunds, and then it looks good to go.

PR #884, opened 2024-03-26
Integrations Boilerplate Codegen Tool

sentilesdal:
Adds a codegen tool to generate much of boilerplate code necessary for new integrations of hyperdrive.  The overall goal is to be to just define a simple `config.yaml` file and then be able to run:

```bash
hyperdrive-codegen --config path/to/config.yaml --out contracts/src
```

This will generate a slew of files:

```
deployers/
    testeth/
      TestETHHyperdriveCoreDeployer.sol
      TestETHHyperdriveDeployerCoordinator.sol
      TestETHTarget0Deployer.sol
      TestETHTarget1Deployer.sol
      TestETHTarget2Deployer.sol
      TestETHTarget3Deployer.sol
      TestETHTarget4Deployer.sol

instances/
    testeth/
      TestETHBase.sol
      TestETHHyperdrive.sol
      TestETHTarget0.sol
      TestETHTarget1.sol
      TestETHTarget2.sol
      TestETHTarget3.sol
      TestETHTarget4.sol

interfaces/
      ITestETH.sol
      ITestETHHyperdrive.sol
```

The configuration yaml will allow us to create certain preconfigured defaults, such as disabling base deposits and stubbing out default functions for those.

I've also added a github workflow to test the different configurations by making sure all the generated code compiles.


codegen/templates/interfaces/IHyperdrive.sol.jinja
```diff
@@ -0,0 +1,14 @@
+// SPDX-License-Identifier: Apache-2.0
+pragma solidity 0.8.20;
+
+import { IHyperdrive } from "../IHyperdrive.sol";
+
+// prettier-ignore
```
cashd, line 6: I don't think this prettier ignore is needed

cashd (approved):
Great job this is going to be soo useful! can't wait to use it in the next integration. I ran through the readme and got every working very smoothly.

codegen/templates/interfaces/IYieldSource.sol.jinja
```diff
@@ -0,0 +1,17 @@
+// SPDX-License-Identifier: GPL-2.0-or-later
```
jalextowle, line 1: We should always use Apache-2.0. We can write interfaces by hand if we need to. It's important that we don't bring any copy-left code into the Hyperdrive codebase.

jalextowle (commented):
This looks great, really nice job dude! Aside from the licensing comment, I have one other high-level comment. I think it would be good to keep the `interfaces/` directory flattened for consistency. If we're going to add folders, we should add folders everywhere. Since that isn't something that we've done for the old yield sources, I'd prefer updating this to output the autogenerated interface into the `contracts/src/interfaces`.

codegen/templates/interfaces/IYieldSource.sol.jinja
sentilesdal: roger, this was copied from our IER4626.sol, should we update there as well?

sentilesdal:
@jalextowle flattened interfaces/ files and updated the license.  left a comment about the GPL license in IERC20.sol

jalextowle (approved):
Looks great! I think the last thing you should do is add yourself and cash as codeowners for the `codegen/` folder so that you can make changes to the tool without needing a review from me or Jonny. We'll still need to review new instances for now.

PR #888, opened 2024-03-27
[Don't Merge] Minimal Parameter Rust SDK Fn's

mcclurejt:
# Resolved Issues
- Fixes https://github.com/delvtech/hyperdrive-rs/issues/32

# Description
- Exposes all hyperdrive-math functionality as functions with minimal parameters
- Use these functions to power the existing `State` method implementations

# Review Checklists

### Rust

- [x] **Testing**
    - Shouldn't be needed since it leverages existing tests and doesn't add any "new" functionality
- [ ] **Semantics**
    - [x] Are the naming conventions clear and consistent?
    - [x] Are exports scoped logically?
    - [ ] Some functions have an excessive amount of arguments, do you see opportunities for reduction?
      -  EX: Can any arguments be derived from some combination of the others (w/o significant perf impact)?

### Rust

- [ ] **Testing**
    - Shouldn't be needed since it leverages existing tests and doesn't add any "new" functionality
- [ ] **Semantics**
    - [ ] Are the naming conventions clear and consistent?
    - [ ] Are exports scoped logically?
    - [ ] Many functions have an excessive amount of arguments, do you see opportunities for reduction?
      -  EX: Can any arguments be derived from some combination of the others (w/o significant perf impact)?

### DIFF
--- a/crates/hyperdrive-math/src/lib.rs
+++ b/crates/hyperdrive-math/src/lib.rs
@@ -126,9 +126,9 @@ impl State {
     /// Gets the pool's spot rate.
     pub fn get_spot_rate(&self) -> FixedPoint {
         get_spot_rate(
-            self.ze(),
-            self.y(),
-            self.mu(),
+            self.effective_share_reserves(),
+            self.bond_reserves(),
+            self.initial_vault_share_price(),
             self.t(),
             self.position_duration(),
         )
--- a/crates/hyperdrive-math/src/lib.rs
+++ b/crates/hyperdrive-math/src/lib.rs
@@ -15,6 +15,7 @@ use rand::{
 };
 pub use short::*;
 pub use utils::*;
+use yield_space::get_spot_price;
 pub use yield_space::YieldSpace;
 
 #[derive(Clone, Debug)]
@@ -123,15 +124,18 @@ impl State {
 
     /// Gets the pool's spot rate.
     pub fn get_spot_rate(&self) -> FixedPoint {
-        let annualized_time =
-            self.position_duration() / FixedPoint::from(U256::from(60 * 60 * 24 * 365));
-        let spot_price = self.get_spot_price();
-        (fixed!(1e18) - spot_price) / (spot_price * annualized_time)
+        get_spot_rate(
+            self.ze(),
+            self.y(),
+            self.mu(),
+            self.t(),
+            self.position_duration(),
+        )
     }
 
     /// Converts a timestamp to the checkpoint timestamp that it corresponds to.
     pub fn to_checkpoint(&self, time: U256) -> U256 {
-        time - time % self.config.checkpoint_duration
+        to_checkpoint(self.checkpoint_duration(), time)
     }
 
     /// Gets the normalized time remaining
@@ -197,6 +201,7 @@ impl State {
         self.info.share_reserves.into()
     }
 
+    #[allow(dead_code)]
     fn effective_share_reserves(&self) -> FixedPoint {
         get_effective_share_reserves(self.share_reserves(), self.share_adjustment())
     }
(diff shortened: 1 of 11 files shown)


crates/hyperdrive-math/src/long/close.rs
```diff
@@ -1,32 +1,73 @@
...
+fn calculate_close_long_flat_plus_curve<F: Into<FixedPoint>>(
+    ze: F,
+    y: F,
+    c: F,
+    mu: F,
+    t: F,
+    dy: F,
+    tr: F,
+) -> FixedPoint {
+    let [ze, y, c, mu, t, dy, tr] = [ze, y, c, mu, t, dy, tr].map(|v| v.into());
```
jalextowle, line 18: One thing to watch out for with this kind of trait declaration is that `F` is bound to _one_ kind of `Into<FixedPoint>`. So if you have `a: I256` and `b: U256`, you won't be able to call `calculate_close_long_flat_plus_curve(a, b)` since they aren't the same `F`. It's gross, but if we want to preserve this functionality, we'd need `F1`, `F2`, etc. 

Here's an example so that you can see what I'm talking about: https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=a97ab399342a14ed0e66d748b5be57d9.

crates/hyperdrive-math/src/lib.rs
```diff
@@ -15,6 +17,7 @@ use rand::{
 };
 pub use short::*;
 pub use utils::*;
+use yield_space::get_spot_price;
```
jalextowle, line 20: We might want to do a `pub use yield_space::*;` so that all of the yield space functions are publicly accessible.

jalextowle: This would replace this line and the line below.

crates/hyperdrive-math/src/long/max.rs
```diff
@@ -2,7 +2,623 @@ use ethers::types::I256;
...
+    z_min: FixedPoint,
+    time_stretch: FixedPoint,
+    c_initial: FixedPoint,
+    flat_fee: FixedPoint,
+    curve_fee: FixedPoint,
+    governance_lp_fee: FixedPoint,
+    long_exposure: FixedPoint,
+    budget: F,
+    checkpoint_exposure: I,
+    maybe_max_iterations: Option<usize>,
```
jalextowle, line 70: I see what you mean now, these are pretty big and gnarly. I do think that some structs would clean this up. Having a struct that contains all of the yield space stuff seems like it would get us pretty far.

jalextowle: I'll leave it to you if you want to make this change. It would be good to get a review from @dpaiton or @slundqui since they are the people who requested this and are the most likely consumers.

dpaiton:
While the use of math variables (`ze`, `mu`) makes it easier to compare to the derivations, I nevertheless recommend that we switch them to readable nouns (for example `effective_share_reserves` instead of `ze`). I think at this stage of the code, the most likely users referencing these functions are going to want to know what variables they need to pass in from pool info, and are not going to be checking it against the math.

I think @jalextowle has had a differing opinion from me on this, so @mcclurejt lets wait for his input before proceeding. Alex, if you disagree then I'm fine with whatever you prefer.

dpaiton:
> I'll leave it to you if you want to make this change. It would be good to get a review from @dpaiton or @slundqui since they are the people who requested this and are the most likely consumers.

I did a quick survey of what functions we are using now. I started making a list, but decided that was dumb because it was most of them. The ones we presently make available via the Python interface are [here](https://github.com/delvtech/agent0/blob/main/src/agent0/ethpy/hyperdrive/interface/_mock_contract.py).

Another useful piece of information is why we are asking for this, which is outlined in [agent0/1355](https://github.com/delvtech/agent0/issues/1355). tl;dr is we want to mock operations in a loop without having to `deepcopy(pool_state)`, which is a slow and unnecessary operation for e.g. calculating the fixed rate after delta bonds & delta shares have been applied.

Ok, so given all of this I still agree that 16 input variables feels way too cumbersome. However I think putting it into a struct is too much like the original fn (passing the state). I would recommend that we set a cutoff (8 seems like a reasonable upper-bound. `pylint` and `ruff` python linters suggest 5. I'm fine with you making that call @mcclurejt), and anything over the cutoff still requires the original `pool_info` and `pool_config` arguments.

crates/hyperdrive-math/src/long/close.rs
jalextowle: It's to make it easier to call the function. If you have a `U256` and the argument is a `FixedPoint`, you have to know that and cast it yourself. This pattern allows you to just pass the parameter in directly without worrying about conversions.

This pattern was initially used in all of the publicly accessible functions. We've made more and more of the Rust SDK public, and I would agree that we've dropped some consistency. We should recover that. FWIW, I think a lot of the functions that are public are poorly named/documented, which is something that we'll want to keep an eye on.

mcclurejt:
Currently on Do-Not-Merge because the changes from https://github.com/delvtech/hyperdrive/commit/12af6c0088d796a81809b4d9ddb314e1eb877ccc have not been accomodated, will do so shortly but wanted to open up review as the changes won't be too dramatic

dpaiton:
I would prefer using the existing checkpoint, PoolConfig, and Poolnfo structs, as needed, over adding a new one. The new one creates a decent amount of overhead in the wrappers and downstream uses. I think an acceptable outcome is that functions with a reasonable number of arguments use them and the rest use checkpoint, config, & info structs.

PR #889, opened 2024-03-27
Generalize `open_long` & `open_short` test cases into `InstanceTest` suite

### LINKED ISSUE #885: Generalize `open_long` instance test cases
Build on #883 and add testing coverage for `open_long` test cases.

- add `test_open_long_with_base` to `InstanceTestingSuite` contract.
- add `test_open_long_with_shares` to `InstanceTestingSuite` contract.

cashd:
# Resolved Issues
Resolves #885 #890 

# Description
Build on #883 and adds generalized testing coverage for open_long test cases.

- adds `test_open_long_with_base` to `InstanceTest` contract.
- adds `test_open_long_with_shares` to `InstanceTest` contract.
- adds `test_open_short_with_base` to `InstanceTest` contract.
- adds `test_open_short_with_shares` to `InstanceTest` contract.

# Review Checklists

be removed.

### Solidity

- [ ] **Tokens**
              token spend?
- [ ] **Reentrancy**
- [ ] **Gas Optimizations**
          memory variables?
          issues?
- [ ] **Visibility**
- [ ] **Math**
          catch underflows?
- [ ] **Testing**

### Rust

- [ ] **Testing**


test/utils/InstanceTest.sol
```diff
@@ -458,4 +496,128 @@ abstract contract InstanceTest is HyperdriveTest {
...
+
+    /// @dev Fuzz Test to ensure deposit accounting is correct for opening longs
+    ///      with the share token.
+    /// @param basePaid Amount in terms of base to open a long.
+    function test_open_long_with_shares(uint256 basePaid) external {
+        // Get balance information before opening a long.
+        (
+            uint256 totalBaseSupplyBefore,
+            uint256 totalSharesSupplyBefore
+        ) = getSupply();
```
jalextowle, line 508: This looks like the wrong order (above `shares` come before `base`). FWIW, this order is more intuitive to me (`base` and then `shares`), but regardless it looks like this needs to be updated.

jalextowle: Why did this and the below test pass with this order?

test/utils/InstanceTest.sol
```diff
@@ -244,8 +244,7 @@ abstract contract InstanceTest is HyperdriveTest {
         }
     }
 
-    /// @dev Deploys the Hyperdrive Factory contract and
-    ///      sets the default pool configuration.
```
jalextowle, line 248: This comment was shorter to stay within 80 chars. We'll want to do this everywhere to stay consistent.

test/utils/InstanceTest.sol
```diff
@@ -458,4 +496,128 @@ abstract contract InstanceTest is HyperdriveTest {
...
+            vm.expectRevert(
+                isBaseETH
+                    ? IHyperdrive.NotPayable.selector
+                    : IHyperdrive.UnsupportedToken.selector
+            );
+        }
+
+        // Bob opens a long by depositing the base token.
+        vm.startPrank(bob);
+        hyperdrive.openLong{ value: isBaseETH ? basePaid : 0 }(
```
jalextowle, line 578: It would be good to check the return value.

test/utils/InstanceTest.sol
```diff
@@ -458,4 +496,128 @@ abstract contract InstanceTest is HyperdriveTest {
...
+        basePaid = basePaid.normalizeToRange(
+            2 * hyperdrive.getPoolConfig().minimumTransactionAmount,
+            maxLongAmount > maxShareAmount ? maxShareAmount : maxLongAmount
+        );
+
+        // Convert the amount to deposit in terms of the share token.
+        uint256 sharesPaid = convertToShares(basePaid);
+
+        // Bob opens a long by depositing the share token.
+        openLong(bob, sharesPaid, false);
```
jalextowle, line 530: Same here, it would be good to check the return.

test/instances/steth/StETHHyperdrive.t.sol
```diff
@@ -92,6 +92,100 @@ contract StETHHyperdriveTest is InstanceTest {
...
+
+    /// @dev Fetches the token balance information of an account.
+    function getTokenBalances(
+        address account
+    ) internal view override returns (uint256, uint256) {
+        return (LIDO.sharesOf(account), LIDO.balanceOf(account));
+    }
+
+    /// @dev Verifies that deposit accounting is correct when opening positions.
+    function verifyDeposit(
```
jalextowle, line 108: Any reason not to abstract this further? It seems like we could define some functions like `getSupply` and `getTokenBalances` that we could use to define this. Maybe I'm missing something, but it seems like the kind of thing that we shouldn't need to implement from scratch for each yield source.

test/utils/InstanceTest.sol
cashd: Agreed on base then shares, I confused my own self haha. This is actually how I implemented the override functions so the test cases are passing.

test/utils/InstanceTest.sol
cashd: Is the style requirement only comment part or entire line (including natspec)? Do you use a plugin to automatically do this for you?

test/instances/steth/StETHHyperdrive.t.sol
cashd: I've considered this also, but there is some divergence between what values we assert depending on the instance. The biggest difference is with stETH and the other instances. For example stETH is checking the underlying balance and shares when depositing as shares. With yield bearing tokens it's just a share token balance check. Given the divergence I plan to tackle the lowest hanging fruit first, open/close positions test cases.

cashd:
I added `open_short` test cases to this PR

test/utils/InstanceTest.sol
```diff
@@ -315,12 +315,59 @@ abstract contract InstanceTest is HyperdriveTest {
...
+    /// @param account The account to fetch token balances of.
+    /// @return sharesBalance The shares token balance of the account.
+    /// @return baseBalance The base token balance of the account.
+    function getTokenBalances(
+        address account
+    )
+        internal
+        view
+        virtual
+        returns (uint256 sharesBalance, uint256 baseBalance);
```
jalextowle, line 361: Should this be base and then shares? That would be more consistent with the below function.

test/utils/InstanceTest.sol
```diff
@@ -458,4 +505,340 @@ abstract contract InstanceTest is HyperdriveTest {
...
+        // Ensure that Bob received the correct amount of bonds.
+        assertEq(
+            hyperdrive.balanceOf(
+                AssetId.encodeAssetId(AssetId.AssetIdPrefix.Long, maturityTime),
+                bob
+            ),
+            bondAmount
+        );
+
+        // If base deposits are supported we ensure the deposit accounting is correct.
```
jalextowle, line 642: Nit:

```suggestion
        // If base deposits are supported, we ensure the deposit accounting is correct.
```

test/utils/InstanceTest.sol
jalextowle: It's all of it. I don't have a plugin do it, but I do have a marker in my editor that tells me where 80 characters is. A plugin would be great.

test/instances/steth/StETHHyperdrive.t.sol
jalextowle: Ok, that makes sense. As long as we get to this eventually, I'm happy.

jalextowle (approved):
LGTM once the last few comments are addressed

PR #893, opened 2024-03-27
adds a variation of the get_spot_price fn to include fees

dpaiton:
# Resolved Issues
The `get_spot_price` rust fn does not include fees, I need one that does for the targeted long fn.

# Description
Simple change that uses `calculate_open_long` instead of `calculate_bonds_out_given_shares_in_down` to get the delta bonds.

# Review Checklists

be removed.

### Solidity

- [ ] **Tokens**
              token spend?
- [ ] **Reentrancy**
- [ ] **Gas Optimizations**
          memory variables?
          issues?
- [ ] **Visibility**
- [ ] **Math**
          catch underflows?
- [ ] **Testing**

### Rust

- [ ] **Testing**

### DIFF
--- a/crates/hyperdrive-math/src/long/open.rs
+++ b/crates/hyperdrive-math/src/long/open.rs
@@ -43,6 +43,12 @@ impl State {
         self.calculate_open_long(base_amount)
     }
 
+    /// Gets the spot price after opening the long on the YieldSpace curve with fees.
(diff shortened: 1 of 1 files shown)


dpaiton:
closed in favor of https://github.com/delvtech/hyperdrive/pull/833

PR #906, opened 2024-04-01
fix unsafe routes in _distributeExcessIdleSafe

mcclurejt:
# Resolved Issues

- fixes https://github.com/spearbit-audits/delv-review/issues/22

# Description

Some paths in `_distributeExcessIdleSafe` are not safe, they should be. See https://github.com/spearbit-audits/delv-review/issues/22 for more details.

# Review Checklists

be removed.

### Solidity

- [x] **Tokens**
              token spend?
- [x] **Reentrancy**
- [x] **Gas Optimizations**
          memory variables?
          issues?
- [x] **Visibility**
- [x] **Math**
          catch underflows?
- [x] **Testing**

### Solidity

- [ ] **Tokens**
              token spend?
- [ ] **Reentrancy**
- [ ] **Gas Optimizations**
          memory variables?
          issues?
- [ ] **Visibility**
- [ ] **Math**
          catch underflows?
- [ ] **Testing**

### DIFF
--- a/contracts/src/internal/HyperdriveLP.sol
+++ b/contracts/src/internal/HyperdriveLP.sol
@@ -552,7 +552,7 @@ abstract contract HyperdriveLP is
     function _updateLiquiditySafe(
         int256 _shareReservesDelta
     ) internal returns (bool) {
-        // Calculate the updated reserves.
+        // Calculate the updated reserves, return false if the calculation fails
         uint256 shareReserves_ = _marketState.shareReserves;
         int256 shareAdjustment_ = _marketState.shareAdjustment;
         uint256 bondReserves_ = _marketState.bondReserves;
@@ -572,7 +572,7 @@ abstract contract HyperdriveLP is
             return false;
         }
 
-        // Update the market state.
+        // Update the market state and return true since the update was successful
         if (updatedShareReserves != shareReserves_) {
             _marketState.shareReserves = updatedShareReserves.toUint128();
         }
@@ -582,7 +582,6 @@ abstract contract HyperdriveLP is
         if (updatedBondReserves != bondReserves_) {
             _marketState.bondReserves = updatedBondReserves.toUint128();
         }
-
         return true;
     }
 }
--- a/contracts/src/internal/HyperdriveLP.sol
+++ b/contracts/src/internal/HyperdriveLP.sol
@@ -540,7 +540,7 @@ abstract contract HyperdriveLP is
     /// @dev Updates the pool's liquidity and holds the pool's spot price constant.
     /// @param _shareReservesDelta The delta that should be applied to share reserves.
     function _updateLiquidity(int256 _shareReservesDelta) internal {
-        // attempt updating the pool's liquidity, revert if the update fails
+        // Attempt updating the pool's liquidity, revert if the update fails
         if (!_updateLiquiditySafe(_shareReservesDelta)) {
             revert IHyperdrive.UpdateLiquidityFailed();
         }
--- a/contracts/src/internal/HyperdriveLP.sol
+++ b/contracts/src/internal/HyperdriveLP.sol
@@ -525,7 +525,7 @@ abstract contract HyperdriveLP is
             .calculateDistributeExcessIdle(params);
 
         // Remove the withdrawal pool proceeds from the reserves.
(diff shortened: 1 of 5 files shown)


contracts/src/internal/HyperdriveLP.sol
```diff
@@ -564,4 +567,40 @@ abstract contract HyperdriveLP is
             _marketState.bondReserves = updatedBondReserves.toUint128();
         }
     }
+
+    function _updateLiquiditySafe(int256 _shareReservesDelta) internal returns (bool) {
```
mcclurejt, line 571: Is there a preferred way to do this over simply re-implementing the logic w/ `calculateUpdateLiquiditySafe`?

contracts/src/internal/HyperdriveLP.sol
```diff
@@ -564,4 +567,42 @@ abstract contract HyperdriveLP is
             _marketState.bondReserves = updatedBondReserves.toUint128();
         }
     }
+
+    function _updateLiquiditySafe(
```
jalextowle, line 571: Two things:

1. This needs `NatSpec` comments. You can copy the NatSpec from the function above, and we'll need to add a `@return` field for the failure flag. You can copy the failure flag comment that we have on other `Safe` functions.
2. We shouldn't duplicate logic like this. Let's use this function implement `_updateLiquidity`. As part of that change, we may be eliminating the only logic that calls `calculateUpdateLiquidity`, and we should eliminate that.

contracts/src/internal/HyperdriveLP.sol
```diff
@@ -564,4 +567,42 @@ abstract contract HyperdriveLP is
...
+                shareReserves_,
+                shareAdjustment_,
+                bondReserves_,
+                _minimumShareReserves,
+                _shareReservesDelta
+            );
+
+        if (!success) {
+            return false;
+        }
```
jalextowle, line 593: I know this sounds nit-picky, but the style we use in this codebase is to avoid having "naked" code like this. In this case, I'd suggest removing the newline, so that it is part of the above code block. Otherwise, we'll need to add a comment explaining what this is doing.

contracts/src/libraries/LPMath.sol
```diff
@@ -756,9 +756,17 @@ library LPMath {
                 // proceeds couldn't be calculated.
                 return 0;
             }
-            uint256 presentValue = calculatePresentValue(
+
```
jalextowle, line 759: Nit: Remove this newline. This falls into the broader theme of commenting code blocks. I see the value of separating this, so you can also remove the part of the above comment that pertains to the new block and comment this separately.

jalextowle (commented):
I'm leaving a few nits. Since you're still onboarding onto our style, I'll give this another look, but it's ready to approve after that. It would be nice to have tests, but as discussed offline, I think that is a stretch when we use the `Safe` pattern.

contracts/src/internal/HyperdriveLP.sol
```diff
@@ -524,36 +524,55 @@ abstract contract HyperdriveLP is
...
+    }
+
+    /// @dev Updates the pool's liquidity and holds the pool's spot price constant.
+    /// @param _shareReservesDelta The delta that should be applied to share reserves.
+    /// @return A flag indicating if the update succeeded.
+    function _updateLiquiditySafe(
+        int256 _shareReservesDelta
+    ) internal returns (bool) {
+        // Calculate the updated reserves, return false if the calculation fails
```
sentilesdal, line 555: Add period to sentence.

contracts/src/internal/HyperdriveLP.sol
```diff
@@ -524,36 +524,55 @@ abstract contract HyperdriveLP is
...
                 bondReserves_,
                 _minimumShareReserves,
                 _shareReservesDelta
             );
+        if (!success) {
+            return false;
+        }
 
-        // Update the market state.
+        // Update the market state and return true since the update was successful
```
sentilesdal, line 575: Add period :P

sentilesdal (commented):
Nit:  style guide is to use proper sentence structure in comments.  Couple more I didn't comment on.

jalextowle (approved):
LGTM. Nice work dude!

PR #909, opened 2024-04-01
use msg.sender,deploymentId in target deployment addresses

mcclurejt:
# Resolved Issues

- fixes https://github.com/spearbit-audits/delv-review/issues/26

# Description

The effective salt provided for deploying targets 1-4 is currently not hashed with other parameter to prevent-front running DoS/griefing attacks. Include msg.sender and deploymentId in the "salt".

# Review Checklists

be removed.

### Solidity

- [x] **Tokens**
    - [x] Do all `approve` calls use `forceApprove`?
    - [x] Do all `transfer` calls use `safeTransfer`?
    - [x] Do all `transferFrom` calls use `msg.sender` as the `from` address?
        - [x] If not, is the function access restricted to prevent unauthorized
              token spend?
    - [x] Is the returned `success` boolean checked to handle failed calls?
    - [x] If using `delegatecall`, are there strict access controls on the
- [x] **Reentrancy**
    - [x] Are functions that make external calls or transfer ether marked as `nonReentrant`?
        - [x] If not, is there documentation that explains why reentrancy is
- [x] **Gas Optimizations**
    - [x] Is the logic as simple as possible?
    - [x] Are the storage values that are used repeatedly cached in stack or
          memory variables?
    - [x] If loops are used, are there guards in place to avoid out-of-gas
          issues?
- [x] **Visibility**
    - [x] Are all `payable` functions restricted to avoid stuck ether?
- [x] **Math**
    - [x] Is all of the arithmetic checked or guarded by if-statements that will
          catch underflows?
    - [x] If `Safe` functions are altered, are potential underflows and
    - [x] Are all of the rounding directions clearly documented?
- [ ] **Testing**

### Solidity

- [ ] **Tokens**
              token spend?
- [ ] **Reentrancy**
- [ ] **Gas Optimizations**
          memory variables?
          issues?
- [ ] **Visibility**
- [ ] **Math**
          catch underflows?
- [ ] **Testing**

### DIFF
--- a/test/integrations/deployers/DeployerCoordinator.t.sol
+++ b/test/integrations/deployers/DeployerCoordinator.t.sol
@@ -320,31 +320,8 @@ contract DeployerCoordinatorTest is HyperdriveTest {
         assertEq(deployment.target4, targets[3]);
     }
 
-    function test_deployTarget_frontrun() external {
-        // Deploy a target0 instance as deployer (should succeed).
-        coordinator.deployTarget(DEPLOYMENT_ID, config, new bytes(0), 0, SALT);
-
-        vm.stopPrank();
-        vm.startPrank(bob);
-        // Try to frontrun the target1 deployment with bob
-        vm.expectRevert(
-            IHyperdriveDeployerCoordinator.DeploymentDoesNotExist.selector
-        );
-        coordinator.deployTarget(DEPLOYMENT_ID, config, new bytes(0), 1, SALT);
-        vm.stopPrank();
-    }
-
-    function test_deployTarget_frontrun_success() external {
-        // Deploy a target0 instance.
-        address target0 = coordinator.deployTarget(
-            DEPLOYMENT_ID,
-            config,
-            new bytes(0),
-            0,
-            SALT
-        );
-
-        // Deploy another target0 as the frontrunner (bob) with same deployment_id
+    function test_deployTarget_frontRun_success() external {
+        // Deploy all target instances as the frontrunner (bob)
         vm.stopPrank();
         vm.startPrank(bob);
         for (uint256 i = 0; i < 5; i++) {
@@ -357,6 +334,19 @@ contract DeployerCoordinatorTest is HyperdriveTest {
             );
         }
         vm.stopPrank();
+
+        // Deploy all target instances as the intended deployer (alice)
+        vm.startPrank(alice);
+        address[] memory targets = new address[](5);
+        for (uint256 i = 0; i < 5; i++) {
+            targets[i] = coordinator.deployTarget(
+                DEPLOYMENT_ID,
+                config,
+                new bytes(0),
+                i,
+                SALT
+            );
+        }
 
         // Ensure that the deployment was configured correctly.
         HyperdriveDeployerCoordinator.Deployment memory deployment = coordinator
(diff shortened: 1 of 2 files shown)


jalextowle (commented):
This looks good, but it would be nice to have some test cases that verify that these can't be front-run anymore.

mcclurejt:
> This looks good, but it would be nice to have some test cases that verify that these can't be front-run anymore.

added here: a50065f654110e506b526cc0c1b08e5fc0fcc433

test/integrations/deployers/DeployerCoordinator.t.sol
```diff
@@ -320,6 +320,85 @@ contract DeployerCoordinatorTest is HyperdriveTest {
...
+        coordinator.deployTarget(DEPLOYMENT_ID, config, new bytes(0), 0, SALT);
+
+        vm.stopPrank();
+        vm.startPrank(bob);
+        // Try to frontrun the target1 deployment with bob
+        vm.expectRevert(
+            IHyperdriveDeployerCoordinator.DeploymentDoesNotExist.selector
+        );
+        coordinator.deployTarget(DEPLOYMENT_ID, config, new bytes(0), 1, SALT);
+        vm.stopPrank();
```
jalextowle, line 334: This would have passed before. What I think would make a better test is to have `bob` try to deploy `target0` with `DEPLOYMENT_ID` and `SALT`. Next, we should have `alice` deploy `target0` with the same `DEPLOYMENT_ID` and `SALT` to illustrate that Alice can't be front-run by Bob.

test/integrations/deployers/DeployerCoordinator.t.sol
```diff
@@ -320,6 +320,85 @@ contract DeployerCoordinatorTest is HyperdriveTest {
         assertEq(deployment.target4, targets[3]);
     }
 
+    function test_deployTarget_frontrun() external {
+        // Deploy a target0 instance as deployer (should succeed).
+        coordinator.deployTarget(DEPLOYMENT_ID, config, new bytes(0), 0, SALT);
+
+        vm.stopPrank();
+        vm.startPrank(bob);
+        // Try to frontrun the target1 deployment with bob
```
jalextowle, line 329: Nit: Let's keep comments at the top of the code block (in this case, we'd move it above `vm.stopPrank()`).

test/integrations/deployers/DeployerCoordinator.t.sol
```diff
@@ -320,6 +320,85 @@ contract DeployerCoordinatorTest is HyperdriveTest {
...
+        vm.expectRevert(
+            IHyperdriveDeployerCoordinator.DeploymentDoesNotExist.selector
+        );
+        coordinator.deployTarget(DEPLOYMENT_ID, config, new bytes(0), 1, SALT);
+        vm.stopPrank();
+    }
+
+    function test_deployTarget_frontrun_success() external {
```
jalextowle, line 337: This isn't testing anything for target1-4. Additionally, Bob isn't the front-runner. Alice is front-running Bob since she submits her deployment first.

We can fix this by having Alice first deploy all of the targets (as the front-runner), and then leave Bob's loop below. This will demonstrate that Alice can't front-run Bob.

jalextowle, line 337: Nit: Let's use camel casing here.

```suggestion
    function test_deployTarget_frontRun_success() external {
```

test/integrations/deployers/DeployerCoordinator.t.sol
jalextowle: Since the below test can be used to test that front-running will fail, we can get rid of this test case.

jalextowle (approved):

PR #913, opened 2024-04-01
allow sweeping of base tokens

mcclurejt:
# Resolved Issues

- fixes https://github.com/spearbit-audits/delv-review/issues/35

# Description

The Hyperdrive pool should only custody vault shares (deposited base tokens are converted to shares, base tokens withdrawn from the vault are paid out immediately). Any lingering base token balance of the contract is by mistake and sweep should therefore be able to transfer out these tokens.

# Review Checklists

be removed.

### Solidity

- [x] **Tokens**
              token spend?
- [x] **Reentrancy**
- [x] **Gas Optimizations**
          memory variables?
          issues?
- [x] **Visibility**
- [x] **Math**
          catch underflows?
- [x] **Testing**

### Solidity

- [ ] **Tokens**
              token spend?
- [ ] **Reentrancy**
- [ ] **Gas Optimizations**
          memory variables?
          issues?
- [ ] **Visibility**
- [ ] **Math**
          catch underflows?
- [ ] **Testing**

### DIFF
--- a/contracts/src/internal/HyperdriveAdmin.sol
+++ b/contracts/src/internal/HyperdriveAdmin.sol
@@ -138,7 +138,7 @@ abstract contract HyperdriveAdmin is IHyperdriveEvents, HyperdriveBase {
             revert IHyperdrive.Unauthorized();
         }
 
-        // Gets the Hyperdrive's balance of base and vault shares prior to
+        // Gets the Hyperdrive's balance of vault shares prior to
         // sweeping.
         uint256 shareBalance = _totalShares();
 
@@ -147,7 +147,7 @@ abstract contract HyperdriveAdmin is IHyperdriveEvents, HyperdriveBase {
         uint256 balance = _target.balanceOf(address(this));
         ERC20(address(_target)).safeTransfer(sweepCollector, balance);
 
-        // Ensure that the base and vault shares balance hasn't changed.
+        // Ensure that the vault shares balance hasn't changed.
         if (_totalShares() != shareBalance) {
             revert IHyperdrive.SweepFailed();
         }
--- a/contracts/src/internal/HyperdriveAdmin.sol
+++ b/contracts/src/internal/HyperdriveAdmin.sol
@@ -140,7 +140,6 @@ abstract contract HyperdriveAdmin is IHyperdriveEvents, HyperdriveBase {
 
         // Gets the Hyperdrive's balance of base and vault shares prior to
         // sweeping.
-        uint256 baseBalance = _totalBase();
         uint256 shareBalance = _totalShares();
 
         // Transfer the entire balance of the sweep target to the sweep
@@ -149,7 +148,7 @@ abstract contract HyperdriveAdmin is IHyperdriveEvents, HyperdriveBase {
         ERC20(address(_target)).safeTransfer(sweepCollector, balance);
 
         // Ensure that the base and vault shares balance hasn't changed.
-        if (_totalBase() != baseBalance || _totalShares() != shareBalance) {
+        if (_totalShares() != shareBalance) {
             revert IHyperdrive.SweepFailed();
         }
 
--- a/test/instances/erc4626/Sweep.t.sol
+++ b/test/instances/erc4626/Sweep.t.sol
@@ -133,7 +133,7 @@ contract SweepTest is BaseTest, IHyperdriveEvents {
         vm.stopPrank();
         vm.startPrank(celine);
 
-        // Trying to sweep the base token should succeed since any lingering amount is a mistake
+        // Trying to sweep the base token should succeed since any lingering amount is a mistake.
         address baseToken = address(hyperdrive.baseToken());
         hyperdrive.sweep(IERC20(baseToken));
     }
@@ -142,7 +142,7 @@ contract SweepTest is BaseTest, IHyperdriveEvents {
         vm.stopPrank();
         vm.startPrank(celine);
 
-        // Trying to sweep the base token should succeed since any lingering amount is a mistake
+        // Trying to sweep the base token should succeed since any lingering amount is a mistake.
         hyperdrive.sweep(IERC20(address(baseForwarder)));
     }
 
--- a/test/instances/erc4626/Sweep.t.sol
+++ b/test/instances/erc4626/Sweep.t.sol
@@ -1,22 +1,22 @@
 // SPDX-License-Identifier: Apache-2.0
 pragma solidity 0.8.20;
 
-import { ERC4626Hyperdrive } from "contracts/src/instances/erc4626/ERC4626Hyperdrive.sol";
-import { ERC4626Target0 } from "contracts/src/instances/erc4626/ERC4626Target0.sol";
-import { ERC4626Target1 } from "contracts/src/instances/erc4626/ERC4626Target1.sol";
-import { ERC4626Target2 } from "contracts/src/instances/erc4626/ERC4626Target2.sol";
-import { ERC4626Target3 } from "contracts/src/instances/erc4626/ERC4626Target3.sol";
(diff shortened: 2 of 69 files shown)


mcclurejt:
The recommendation suggests considering the removal of the `_totalBase()` function from `HyperdriveBase` which seems fine to me. The only place I see it being used is in the [MockHyperdriveBase](https://github.com/delvtech/hyperdrive/blob/5ff6cce248c7bdcac557080966aabcb6d1efa008/contracts/test/MockHyperdrive.sol#L242:L256)... any thoughts here?

Also, any idea why the [`_totalShares()`](https://github.com/delvtech/hyperdrive/blob/5ff6cce248c7bdcac557080966aabcb6d1efa008/contracts/src/internal/HyperdriveBase.sol#L221) function exists on `HyperdriveBase`? Couldn't you just use the [totalShares storage variable](https://github.com/delvtech/hyperdrive/blob/5ff6cce248c7bdcac557080966aabcb6d1efa008/contracts/test/MockHyperdrive.sol#L41)? Having both seems confusing.

contracts/src/internal/HyperdriveAdmin.sol
```diff
@@ -147,7 +146,7 @@ abstract contract HyperdriveAdmin is IHyperdriveEvents, HyperdriveBase {
         ERC20(address(_target)).safeTransfer(_sweepCollector, balance);
 
         // Ensure that the base and vault shares balance hasn't changed.
```
sentilesdal, line 148: Update comment?

jalextowle:
> The recommendation suggests considering the removal of the `_totalBase()` function from `HyperdriveBase` which seems fine to me. The only place I see it being used is in the [MockHyperdriveBase](https://github.com/delvtech/hyperdrive/blob/5ff6cce248c7bdcac557080966aabcb6d1efa008/contracts/test/MockHyperdrive.sol#L242:L256)... any thoughts here?
> 
> Also, any idea why the [`_totalShares()`](https://github.com/delvtech/hyperdrive/blob/5ff6cce248c7bdcac557080966aabcb6d1efa008/contracts/src/internal/HyperdriveBase.sol#L221) function exists on `HyperdriveBase`? Couldn't you just use the [totalShares storage variable](https://github.com/delvtech/hyperdrive/blob/5ff6cce248c7bdcac557080966aabcb6d1efa008/contracts/test/MockHyperdrive.sol#L41)? Having both seems confusing.

We use both `_totalBase` and `_totalShares` in the `sweep` function in `HyperdriveAdmin`. The `totalShares` storage variable is only used in `MockHyperdrive` and isn't available for any of the other instances found in `contracts/src/instances`. With that in mind, we can't assume that `totalShares` will be available, and the point of the `_totalShares()` function is to abstract this out.

contracts/src/internal/HyperdriveAdmin.sol
```diff
@@ -138,7 +138,6 @@ abstract contract HyperdriveAdmin is IHyperdriveEvents, HyperdriveBase {
 
         // Gets the Hyperdrive's balance of base and vault shares prior to
         // sweeping.
```
jalextowle, line 140: We should update this comment too.

mcclurejt:
> > The recommendation suggests considering the removal of the `_totalBase()` function from `HyperdriveBase` which seems fine to me. The only place I see it being used is in the [MockHyperdriveBase](https://github.com/delvtech/hyperdrive/blob/5ff6cce248c7bdcac557080966aabcb6d1efa008/contracts/test/MockHyperdrive.sol#L242:L256)... any thoughts here?
> > Also, any idea why the [`_totalShares()`](https://github.com/delvtech/hyperdrive/blob/5ff6cce248c7bdcac557080966aabcb6d1efa008/contracts/src/internal/HyperdriveBase.sol#L221) function exists on `HyperdriveBase`? Couldn't you just use the [totalShares storage variable](https://github.com/delvtech/hyperdrive/blob/5ff6cce248c7bdcac557080966aabcb6d1efa008/contracts/test/MockHyperdrive.sol#L41)? Having both seems confusing.
> 
> We use both `_totalBase` and `_totalShares` in the `sweep` function in `HyperdriveAdmin`. The `totalShares` storage variable is only used in `MockHyperdrive` and isn't available for any of the other instances found in `contracts/src/instances`. With that in mind, we can't assume that `totalShares` will be available, and the point of the `_totalShares()` function is to abstract this out.

ahhh gotcha, not sure what I was thinking with totalShares being a storage variable on `HyperdriveBase` 🤦  thank you

contracts/src/internal/HyperdriveAdmin.sol
```diff
@@ -136,18 +136,17 @@ abstract contract HyperdriveAdmin is IHyperdriveEvents, HyperdriveBase {
             revert IHyperdrive.Unauthorized();
         }
 
-        // Gets the Hyperdrive's balance of base and vault shares prior to
+        // Gets the Hyperdrive's balance of vault shares prior to
         // sweeping.
-        uint256 baseBalance = _totalBase();
```
jalextowle, line 141: Shouldn't we get rid of `_totalBase` entirely? It's not used anywhere else.

test/instances/erc4626/Sweep.t.sol
```diff
@@ -1,22 +1,22 @@
...
+import {IERC20} from "contracts/src/interfaces/IERC20.sol";
+import {IERC4626} from "contracts/src/interfaces/IERC4626.sol";
+import {IHyperdrive} from "contracts/src/interfaces/IHyperdrive.sol";
+import {IHyperdriveEvents} from "contracts/src/interfaces/IHyperdriveEvents.sol";
+import {ONE} from "contracts/src/libraries/FixedPointMath.sol";
+import {HyperdriveMath} from "contracts/src/libraries/HyperdriveMath.sol";
+import {ERC20Mintable} from "contracts/test/ERC20Mintable.sol";
+import {MockERC4626} from "contracts/test/MockERC4626.sol";
+import {BaseTest} from "test/utils/BaseTest.sol";
+import {HyperdriveUtils} from "test/utils/HyperdriveUtils.sol";
```
jalextowle, line 19: Was this done by prettier? The former is more consistent with our style, so we should revert this.

mcclurejt: ahh crap just caught this, fixing now

jalextowle (approved):
Approving this, but it looks like you'll need to run prettier. We should also add a case that demonstrates that base tokens *can* be swept.

test/instances/erc4626/Sweep.t.sol
```diff
@@ -110,34 +110,42 @@ contract SweepTest is BaseTest, IHyperdriveEvents {
...
         // Trying to sweep the vault token via the forwarding token should fail.
         vm.expectRevert(IHyperdrive.SweepFailed.selector);
         hyperdrive.sweep(IERC20(address(vaultForwarder)));
     }
 
+    function test_sweep_success_direct_baseToken() external {
+        vm.stopPrank();
+        vm.startPrank(celine);
+
+        // Trying to sweep the base token should succeed since any lingering amount is a mistake
```
jrhea, line 136:
```suggestion
        // Trying to sweep the base token should succeed since any lingering amount is a mistake.
```

test/instances/erc4626/Sweep.t.sol
```diff
@@ -110,34 +110,42 @@ contract SweepTest is BaseTest, IHyperdriveEvents {
...
+        address baseToken = address(hyperdrive.baseToken());
+        hyperdrive.sweep(IERC20(baseToken));
+    }
+
+    function test_sweep_success_indirect_baseToken() external {
+        vm.stopPrank();
+        vm.startPrank(celine);
+
+        // Trying to sweep the base token should succeed since any lingering amount is a mistake
```
jrhea, line 145:
```suggestion
        // Trying to sweep the base token should succeed since any lingering amount is a mistake.
```

Issue #874, opened 2024-03-20
`calculate_redeem_withdrawal_shares`


ryangoree:
Add a function to the Rust SDK that calculates the amount of proceeds from redeeming withdrawal shares. Traders and UIs could use this to preview their returns without hitting the chain.

Issue #873, opened 2024-03-20
`calculate_add_liquidity`


ryangoree:
Add a function to the Rust SDK that calculates the amount of LP shares received when adding liquidity. Traders and UIs could use this to preview their position without hitting the chain.

sentilesdal:
Fixed in https://github.com/delvtech/hyperdrive/pull/934

Issue #872, opened 2024-03-20
`calculate_remove_liquidity`


ryangoree:
Add a function to the Rust SDK that calculates the amount of `proceeds` and `withdrawal_shares` received for removing liquidity. Traders and UIs could use this to preview their returns without hitting the chain.

jalextowle:
This is related to #862. Both of these will require `_distributeExcessIdle` to be ported over.

sentilesdal:
Several PRs in flight for this:

https://github.com/delvtech/hyperdrive-rs/pull/17
https://github.com/delvtech/hyperdrive-rs/pull/11
One landed:
https://github.com/delvtech/hyperdrive-rs/pull/7

Will need to wait for `updateLiquidity` update due to k-inflation changes, then I'll put up a PR for `calculate_distribute_excess_idle_share_proceeds` and `calculate_distribue_excess_idle`.  The functions are written already, but need to add testing and wait for these PRs to land.

Issue #871, opened 2024-03-19
Review `calculate_max_long` and `calculate_max_short` for accuracy


jalextowle:
A lot has changed in Hyperdrive over the last few months. We should make sure that the math used in these functions is actually correct and fix any discrepancies that we find. This should address any of the lingering known issues in the SDK.

dpaiton:
While doing this it would be good to verify that #797 and #732 are resolved. This is also related to #894.

PR #916, opened 2024-04-01
Add error handling to rust trade logic

slundqui:
# Resolved Issues
https://github.com/delvtech/hyperdrive/issues/743 (partially)

# Description
Adds the following error checks in rust trades:
- Minimum transaction checks for all long and short trades.
- Adds negative interest checks for open and close shorts.

# Review Checklists

be removed.

### Rust

- [ ] **Testing**
          
### Rust

- [x] **Testing**

### DIFF
--- a/crates/hyperdrive-math/src/long/close.rs
+++ b/crates/hyperdrive-math/src/long/close.rs
@@ -111,7 +111,11 @@ mod tests {
         let mut rng = thread_rng();
         let state = rng.gen::<State>();
         let result = std::panic::catch_unwind(|| {
-            state.calculate_close_long(state.config.minimum_transaction_amount - 10, 0.into())
+            state.calculate_close_long(
+                state.config.minimum_transaction_amount - 10,
+                0.into(),
+                0.into(),
+            )
         });
         assert!(result.is_err());
         Ok(())
--- a/crates/hyperdrive-math/src/long/close.rs
+++ b/crates/hyperdrive-math/src/long/close.rs
@@ -110,12 +110,9 @@ mod tests {
     async fn test_close_long_min_txn_amount() -> Result<()> {
         let mut rng = thread_rng();
         let state = rng.gen::<State>();
-        let result = std::panic::catch_unwind(||
-            state.calculate_close_long(
-                state.config.minimum_transaction_amount - 10,
-                0.into(),
-            )
-        );
+        let result = std::panic::catch_unwind(|| {
+            state.calculate_close_long(state.config.minimum_transaction_amount - 10, 0.into())
+        });
         assert!(result.is_err());
         Ok(())
     }
--- a/crates/hyperdrive-math/src/long/close.rs
+++ b/crates/hyperdrive-math/src/long/close.rs
@@ -104,4 +104,19 @@ mod tests {
 
         Ok(())
     }
+
+    // Tests close long with an amount smaller than the minimum.
+    #[tokio::test]
+    async fn test_close_long_min_txn_amount() -> Result<()> {
+        let mut rng = thread_rng();
+        let state = rng.gen::<State>();
+        let result = std::panic::catch_unwind(||
+            state.calculate_close_long(
+                state.config.minimum_transaction_amount - 10,
+                0.into(),
+            )
+        );
+        assert!(result.is_err());
+        Ok(())
+    }
 }
--- a/crates/hyperdrive-math/src/long/close.rs
+++ b/crates/hyperdrive-math/src/long/close.rs
@@ -41,6 +41,11 @@ impl State {
     ) -> FixedPoint {
         let bond_amount = bond_amount.into();
 
+        if bond_amount < self.config.minimum_transaction_amount.into() {
+            // TODO would be nice to return a `Result` here instead of a panic.
+            panic!("MinimumTransactionAmount: Input amount too low");
+        }
+
         // Subtract the fees from the trade
         self.calculate_close_long_flat_plus_curve(bond_amount, maturity_time, current_time)
             - self.close_long_curve_fee(bond_amount, maturity_time, current_time)
--- a/crates/hyperdrive-math/src/short/close.rs
+++ b/crates/hyperdrive-math/src/short/close.rs
@@ -65,6 +65,22 @@ impl State {
         }
     }
 
+    /// Since traders pay a curve fee when they close shorts on Hyperdrive,
+    /// it is possible for traders to receive a negative interest rate even
+    /// if curve's spot price is less than or equal to 1.
+    //
+    /// Given the curve fee `phi_c` and the starting spot price `p_0`, the
+    /// maximum spot price is given by:
+    /// $$
+    /// p_max = 1 - phi_c * (1 - p_0)
+    /// $$
+    fn calculate_close_max_short_price(&self) -> FixedPoint {
+        fixed!(1e18)
+            - self
+                .curve_fee()
+                .mul_up(fixed!(1e18) - self.calculate_spot_price())
+    }
+
     /// Calculates the amount of shares the trader will receive after fees for closing a short
     pub fn calculate_close_short<F: Into<FixedPoint>>(
         &self,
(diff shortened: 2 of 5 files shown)


slundqui:
This PR still needs tests to make sure these checks are being hit in these cases. Would love some help coming up with known trades that result in insufficient liquidity

crates/hyperdrive-math/src/short/open.rs
```diff
@@ -34,19 +34,35 @@ impl State {
...
         // If the open share price hasn't been set, we use the current share
         // price, since this is what will be set as the checkpoint share price
         // in the next transaction.
         if open_vault_share_price == fixed!(0) {
             open_vault_share_price = self.vault_share_price();
         }
 
+        // TODO solidity uses mulUp here, should we do that here as well?
+        let share_reserves_delta_in_shares =
+            self.vault_share_price() * self.short_principal(short_amount)?;
```
ryangoree, line 51: Looks like it was changed to round up a couple months ago. We should probably change this and the calc below to round up to remain at parity and overestimate the proceeds for the negative interest check.

https://github.com/delvtech/hyperdrive/pull/736/files#diff-114831fb0435b556f3b41af29e9f3b9e808cbd0dbd8d3ca504492a09ed4fd6fcR425

Also, I think the name of this variable is a typo, because the number is denominated in base.

crates/hyperdrive-math/src/long/close.rs
```diff
@@ -36,6 +36,13 @@ impl State {
         normalized_time_remaining: F,
     ) -> FixedPoint {
         let bond_amount = bond_amount.into();
+
+        if bond_amount < self.config.minimum_transaction_amount.into() {
+            // TODO would be nice to return a `Result` here instead of a panic.
```
ryangoree, line 41: Agreed. If we don't implement them in this PR, can we add an issue for these so it can be prioritized in the backlog?

dpaiton: Agreed; leave it as a TODO & make an issue. I'll handle it, I already fixed a few in targeted_long PR and I think I'm going to just do all of them.

dpaiton:
Since you're testing these "unhappy paths" & verifying error handling, I think it makes sense to include a set of tests that assert that there is an error if one tries to trade more than the maximum amount.

dpaiton:
This doesn't fully resolve #743, right? We now have coverage for open/close long/short but not for add/remove liquidity.

dpaiton:
Part of the checklist asks if we have sufficient testing against the solidity contracts. One way we could do this is to fuzz a large number of open & close trade amounts, with fuzz bounds that are outside the range of allowable trades (too big & too small). Then the test confirms that if rust allows it, solidity allows it and vice versa. 

There's a python example of this in #743 . I think it would be good to add these tests in this PR.

slundqui:
> This doesn't fully resolve #743, right? We now have coverage for open/close long/short but not for add/remove liquidity.

We currently don't have an add/remove liquidity in rust:
https://github.com/delvtech/hyperdrive/issues/873
https://github.com/delvtech/hyperdrive/issues/874
https://github.com/delvtech/hyperdrive/issues/872

slundqui:
> > > This doesn't fully resolve #743, right? We now have coverage for open/close long/short but not for add/remove liquidity.
> > 
> > 
> > We currently don't have an add/remove liquidity in rust: #873 #874 #872
> 
> Got it; I updated #743. What about the max fns? Should we move that to a new issue or will you address it here?

Up to you if you want to keep these in that issue, but I was going to group error handling while tackling this issue:
https://github.com/delvtech/hyperdrive/issues/871

dpaiton:
> > > > This doesn't fully resolve #743, right? We now have coverage for open/close long/short but not for add/remove liquidity.
> > > 
> > > 
> > > We currently don't have an add/remove liquidity in rust: #873 #874 #872
> > 
> > 
> > Got it; I updated #743. What about the max fns? Should we move that to a new issue or will you address it here?
> 
> Up to you if you want to keep these in that issue, but I was going to group error handling while tackling this issue: #871

sounds good to me, I'll move the task over.

crates/hyperdrive-math/src/short/close.rs
```diff
@@ -75,12 +91,30 @@ impl State {
...
         let share_reserves_delta = self
             .calculate_close_short_flat_plus_curve(bond_amount, normalized_time_remaining)
             + self.close_short_curve_fee(bond_amount, normalized_time_remaining)
             + self.close_short_flat_fee(bond_amount, normalized_time_remaining);
 
+        // Throw an error if closing the short would result in negative interest.
+        let ending_spot_price = {
+            let mut state: State = self.clone();
+            state.info.bond_reserves -= bond_amount.into();
+            state.info.share_reserves += share_reserves_delta.into();
```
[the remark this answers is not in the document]
slundqui: Reworked to include the two negative interest checks, one of which requires accounting for governance fee

slundqui:
> Part of the review checklist asks if we have sufficient testing against the solidity contracts. One way we could do this is to fuzz a large number of open & close trade amounts, with fuzz bounds that are outside the range of allowable trades (too big & too small). Then the test confirms that if rust allows it, solidity allows it and vice versa.
> 
> There's a python example of this in #743 . I think it would be good to add these tests in this PR.

Digging down into this, this is a pretty big undertaking, as our current integration tests (1) doesn't exist for open calls, and (2) only tests against the underlying hyperdrive math functions (instead of the long/short interface), which bypasses both solidity's and rust's fee and error check calculations. Is this something we can split up into a separate PR, and is this critical for our deadlines?

dpaiton:
> > Part of the review checklist asks if we have sufficient testing against the solidity contracts. One way we could do this is to fuzz a large number of open & close trade amounts, with fuzz bounds that are outside the range of allowable trades (too big & too small). Then the test confirms that if rust allows it, solidity allows it and vice versa.
> > There's a python example of this in #743 . I think it would be good to add these tests in this PR.
> 
> Digging down into this, this is a pretty big undertaking, as our current integration tests (1) doesn't exist for open calls, and (2) only tests against the underlying hyperdrive math functions (instead of the long/short interface), which bypasses both solidity's and rust's fee and error check calculations. Is this something we can split up into a separate PR, and is this critical for our deadlines?

Agreed, sounds like a new PR to me. Can you make the issue?

slundqui:
> > > Part of the review checklist asks if we have sufficient testing against the solidity contracts. One way we could do this is to fuzz a large number of open & close trade amounts, with fuzz bounds that are outside the range of allowable trades (too big & too small). Then the test confirms that if rust allows it, solidity allows it and vice versa.
> > > There's a python example of this in #743 . I think it would be good to add these tests in this PR.
> > 
> > 
> > Digging down into this, this is a pretty big undertaking, as our current integration tests (1) doesn't exist for open calls, and (2) only tests against the underlying hyperdrive math functions (instead of the long/short interface), which bypasses both solidity's and rust's fee and error check calculations. Is this something we can split up into a separate PR, and is this critical for our deadlines?
> 
> Agreed, sounds like a new PR to me. Can you make the issue?

Issue made here:
https://github.com/delvtech/hyperdrive/issues/937

crates/hyperdrive-math/src/short/close.rs
```diff
@@ -78,16 +113,58 @@ impl State {
...
+            state.info.share_reserves += share_curve_delta_with_fees.into();
+            state.calculate_spot_price()
+        };
+        if share_curve_delta_with_fees_spot_price > fixed!(1e18) {
+            // TODO would be nice to return a `Result` here instead of a panic.
+            panic!("InsufficientLiquidity: Negative Interest");
+        }
+
+        // Now calculate short proceeds
+        // TODO we've already calculated a couple of internal variables needed by this function,
```
dpaiton, line 155: agreed; can you make a high priority issue for this so we keep track of it?

crates/hyperdrive-math/src/short/fees.rs
```diff
@@ -40,6 +40,19 @@ impl State {
             * bond_amount.mul_div_down(normalized_time_remaining, self.vault_share_price())
     }
 
+    // Calculate the curve portion of the governance fee for close shorts
+    // NOTE: Round down to underestimate the governance curve fee
+    // TODO: avoid duplicate calculation of close short curve fee
```
dpaiton, line 45: you can include this in the other issue on making `calculate_close_short_flat_plus_curve` more efficient

dpaiton:
I think #743 is not resolved yet. Correct me if I'm wrong, but I think we are still missing:

- [ ] upper-bound check & test on close long
  - what if I try to close for more than I have? Is this just a balanceOf check? What about closing more than the pool has? 
- [ ] upper-bound check & test on close short
- [ ] open short negative interest test (check is there; but the unhappy path test is not)

Apologies if you've already said why you didn't do these 😬 

It can be a pain to handle all of the tasks in a monolith issue like that, so I'm fine with breaking out smaller tasks, closing #743, and merging this PR before continuing with the remaining ones.

dpaiton (approved):

PR #918, opened 2024-04-01
standardize checks for msg.value

mcclurejt:
# Resolved Issues

- fixes https://github.com/spearbit-audits/delv-review/issues/28

# Description

- use the same logic for checking msg.value across all contracts

# Review Checklists

be removed.

### Solidity

- [x] **Tokens**
              token spend?
- [x] **Reentrancy**
- [x] **Gas Optimizations**
          memory variables?
          issues?
- [x] **Visibility**
- [x] **Math**
          catch underflows?
- [x] **Testing**

### Solidity

- [ ] **Tokens**
              token spend?
- [ ] **Reentrancy**
- [ ] **Gas Optimizations**
          memory variables?
          issues?
- [ ] **Visibility**
- [ ] **Math**
          catch underflows?
- [ ] **Testing**

### DIFF
--- a/codegen/templates/instances/Base.sol.jinja
+++ b/codegen/templates/instances/Base.sol.jinja
@@ -188,7 +188,7 @@ abstract contract {{ name.capitalized }}Base is HyperdriveBase {
     /// @dev We override the message value check since this integration is
     ///      not payable.
     function _checkMessageValue() internal view override {
-        if (msg.value > 0) {
+        if (msg.value != 0) {
             revert IHyperdrive.NotPayable();
         }
     }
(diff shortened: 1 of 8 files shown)


codegen/templates/instances/Base.sol.jinja
```diff
@@ -188,7 +188,7 @@ abstract contract {{ name.capitalized }}Base is HyperdriveBase {
     /// @dev We override the message value check since this integration is
     ///      not payable.
     function _checkMessageValue() internal view override {
-        if (msg.value > 0) {
+        if (msg.value != 0) {
```
jalextowle, line 191: 💯

jalextowle (approved):

sentilesdal (approved):

PR #919, opened 2024-04-02
Fix `calculate_open_long` negative interest check, fix `get_spot_price_after_short`

ryangoree:
# Description

This PR changes `calculate_open_long` to use the internal `spot_price_after_long` for it's negative interest check to ensure the governance fee is accounted for. It also updates the `get_spot_price_after_short` to use `calculate_open_short` similar to `get_spot_price_after_long` so that fees will be taken into account.

Related to #833 

# Review Checklists

be removed.

### Rust

- [ ] **Testing**


crates/hyperdrive-math/src/short/open.rs
```diff
@@ -50,11 +49,15 @@ impl State {
...
-        let shares_amount = self.calculate_shares_out_given_bonds_in_down(bond_amount);
-        self.spot_price_after_short(shares_amount * self.vault_share_price(), bond_amount)
+    /// Gets the spot price after opening a Hyperdrive short.
+    pub fn get_spot_price_after_short(
+        &self,
+        short_amount: FixedPoint,
+        spot_price: FixedPoint,
+        open_vault_share_price: FixedPoint,
+    ) -> Result<FixedPoint> {
+        self.calculate_open_short(short_amount, spot_price, open_vault_share_price)
```
slundqui, line 59: If we're using this function to check for negative interest, do we not want to call the `calculate_open_short` function here? This would result in infinite recursion.

crates/hyperdrive-math/src/short/open.rs
```diff
@@ -98,7 +101,7 @@ impl State {
 mod tests {
     use std::panic;
 
-    use ethers::types::U256;
+    use ethers::types::{I256, U256};
```
slundqui, line 104: I don't see this import being used, was this residual?

crates/hyperdrive-math/src/short/open.rs
ryangoree: This is the `pub` function `get_spot_price_after_short` which calls `calculate_open_short` to determine the `base_amount`/`short_deposit`.

`calculate_open_short` uses the internal `spot_price_after_short` which takes a specified `base_amount`.

However, @dpaiton has a PR in which these can be consolidated and `calculate_open_short` will only be called if no `base_amount` id provided that I'll probably copy here.

crates/hyperdrive-math/src/short/open.rs
ryangoree: It's being used in the test below and was moved from the file's imports at the top, to the test's imports.

slundqui (approved):
Existing tests should cover these changes. LGTM

crates/hyperdrive-math/src/short/open.rs
```diff
@@ -98,7 +101,7 @@ impl State {
...
     use eyre::Result;
```
mcclurejt, line 105: I think this import is redundant with the import of the same on line 1

mcclurejt (approved):
LGTM other than the redundant import

crates/hyperdrive-math/src/short/open.rs
```diff
@@ -85,75 +88,77 @@ impl State {
...
-                .call()
-                .await
-            {
-                Ok(expected) => {
-                    assert_eq!(actual.unwrap(), FixedPoint::from(expected));
-                }
-                Err(_) => assert!(actual.is_err()),
-            }
+            // TODO: Why can't this pass with a tolerance of 1e9?
+            let tolerance = fixed!(1e10);
```
ryangoree, line 149: The spot price after long test uses `1e9`, but this one can't pass with the same tolerance... 🤔

slundqui: Maybe the source of the increase in error comes from the variable yield from the short?

PR #920, opened 2024-04-02
replace `get_*` with `calculate_*`

dpaiton:
# Resolved Issues
naming convention

# Description
I replaced any fn where we calculate something to have the prefix `calculate_` instead of `get_` per @jrhea 's request.
Fns that return state (`get_state()`; `get_pool_info()`; etc) are left alone.

# Review Checklists

be removed.

### Solidity

- [ ] **Tokens**
              token spend?
- [ ] **Reentrancy**
- [ ] **Gas Optimizations**
          memory variables?
          issues?
- [ ] **Visibility**
- [ ] **Math**
          catch underflows?
- [ ] **Testing**

### Rust

- [ ] **Testing**


crates/hyperdrive-math/src/short/max.rs
```diff
@@ -48,7 +48,7 @@ impl State {
     /// on the realized price that the short will pay. This is used to help the
```
jrhea, line 48: A few lines up change to

Calculates the max short that can be opened with the given budget.

crates/hyperdrive-math/src/short/max.rs
```diff
@@ -68,7 +68,7 @@ impl State {
         // Get the spot price and the open share price. If the open share price
```
jrhea, line 68:
```suggestion
        // Calculate the spot price and the open share price. If the open share price
```

crates/hyperdrive-math/src/short/open.rs
```diff
@@ -52,7 +52,7 @@ impl State {
 
     /// Gets the spot price after opening the short on the YieldSpace curve and
```
jrhea, line 53:
```suggestion
    /// Calculates the spot price after opening the short on the YieldSpace curve and
```

crates/hyperdrive-math/src/short/max.rs
dpaiton: good catch!

crates/test-utils/src/agent.rs
```diff
@@ -953,27 +953,30 @@ impl Agent<ChainClient, ChaCha8Rng> {
             .await?;
         state.calculate_open_short(
             short_amount,
-            state.get_spot_price(),
+            state.calculate_spot_price(),
             open_vault_share_price.into(),
         )
     }
 
     /// Gets the max long that can be opened in the current checkpoint.
```
jrhea, line 961:
```suggestion
    /// Calculates the max long that can be opened in the current checkpoint.
```

crates/test-utils/src/agent.rs
```diff
@@ -953,27 +953,30 @@ impl Agent<ChainClient, ChaCha8Rng> {
...
         let state = self.get_state().await?;
         let checkpoint_exposure = self
             .hyperdrive
             .get_checkpoint_exposure(state.to_checkpoint(self.now().await?))
             .await?;
-        Ok(state.get_max_long(self.wallet.base, checkpoint_exposure, maybe_max_iterations))
+        Ok(state.calculate_max_long(self.wallet.base, checkpoint_exposure, maybe_max_iterations))
     }
 
     /// Gets the max short that can be opened in the current checkpoint.
```
jrhea, line 974:
```suggestion
    /// Calculates the max short that can be opened in the current checkpoint.
```

crates/hyperdrive-math/src/short/max.rs
dpaiton: I found a bunch of these; I'll do a global search & fix them.

jrhea (approved):
im going to go ahead and approve. the only issues are with comments. maybe just grep for get in the comments.

PR #924, opened 2024-04-02
Fixed a couple of checkpointing bugs

jalextowle:
# Description

While I was working on the fix to [Spearbit Issue #34](https://github.com/spearbit-audits/delv-review/issues/34) yesterday, I noticed two issues with our checkpointing logic. First, we were using the provided `_vaultSharePrice` to `_applyCheckpoint` to mint the checkpoint and to perform all of our accounting logic. This doesn't pass the sniff test since our accounting logic uses the open vault share price, the close vault share price, and the current vault share price in general. I originally noticed this when thinking about the vault share price being provided to `_distributeExcessIdleSafe`, and then I noticed some even more egregious cases involving the accounting logic used to decide the base proceeds to reserve for matured longs and shorts. Second, we weren't performing a linear search to get the next share price in `_applyCheckpoint`, which could have caused cases of negative interest for checkpoints created using `closeLong` and `closeShort`. These functions create checkpoints in the past, and they passed the current `_pricePerVaultShare()` value into `_applyCheckpoint` when minting these checkpoints.

# Review Checklists

be removed.

### Solidity

- [x] **Tokens**
              token spend?
- [x] **Reentrancy**
- [x] **Gas Optimizations**
          memory variables?
          issues?
- [x] **Visibility**
- [x] **Math**
          catch underflows?
- [x] **Testing**

### Rust

- [x] **Testing**

### Solidity

- [x] **Tokens**
              token spend?
- [x] **Reentrancy**
- [x] **Gas Optimizations**
          memory variables?
          issues?
- [x] **Visibility**
- [x] **Math**
          catch underflows?
- [x] **Testing**

### Rust

- [ ] **Testing**

### DIFF
--- a/contracts/src/internal/HyperdriveCheckpoint.sol
+++ b/contracts/src/internal/HyperdriveCheckpoint.sol
@@ -46,28 +46,8 @@ abstract contract HyperdriveCheckpoint is
             revert IHyperdrive.InvalidCheckpointTime();
         }
 
-        // If the checkpoint time is the latest checkpoint, we use the current
-        // vault share price. Otherwise, we use a linear search to find the
-        // closest vault share price and use that to perform the checkpoint.
-        if (_checkpointTime == latestCheckpoint) {
-            _applyCheckpoint(latestCheckpoint, _pricePerVaultShare());
-        } else {
-            for (
-                uint256 time = _checkpointTime;
-                ;
-                time += _checkpointDuration
-            ) {
-                uint256 closestVaultSharePrice = _checkpoints[time]
-                    .vaultSharePrice;
-                if (time == latestCheckpoint && closestVaultSharePrice == 0) {
-                    closestVaultSharePrice = _pricePerVaultShare();
-                }
-                if (closestVaultSharePrice != 0) {
-                    _applyCheckpoint(_checkpointTime, closestVaultSharePrice);
-                    break;
-                }
-            }
-        }
+        // Apply the checkpoint.
+        _applyCheckpoint(_checkpointTime, _pricePerVaultShare());
     }
 
     /// @dev Creates a new checkpoint if necessary.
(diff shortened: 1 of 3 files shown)


contracts/src/internal/HyperdriveCheckpoint.sol
```diff
@@ -238,28 +248,32 @@ abstract contract HyperdriveCheckpoint is
...
     /// @return shareProceeds The proceeds of the holders in shares.
     /// @return governanceFee The fee paid to governance in shares.
     function _calculateMaturedProceeds(
         uint256 _bondAmount,
-        uint256 _vaultSharePrice,
         uint256 _openVaultSharePrice,
+        uint256 _closeVaultSharePrice,
+        uint256 _vaultSharePrice,
         bool _isLong
     ) internal view returns (uint256 shareProceeds, uint256 governanceFee) {
```
mcclurejt, line 278: Why does the [shareProceeds](https://github.com/delvtech/hyperdrive/pull/924/files#diff-1f921bd076bd7a60928e5d64b14a74f3f8245316baf175a7f7cfe12ae3d846c8R285) calculation uses `_vaultSharePrice` instead of `_closeVaultSharePrice`?

Logically I would expect proceeds to be calculated based on the price at the maturation checkpoint, rather than the current price, for a mature position.

mcclurejt (approved):
🚢

contracts/src/internal/HyperdriveCheckpoint.sol
```diff
@@ -238,28 +248,32 @@ abstract contract HyperdriveCheckpoint is
...
         );
 
-        return _vaultSharePrice;
+        return checkpointVaultSharePrice;
     }
 
     /// @dev Calculates the proceeds of the holders of a given position at
     ///      maturity.
     /// @param _bondAmount The bond amount of the position.
     /// @param _vaultSharePrice The current vault share price.
```
jrhea, line 264: These natspec params are out of order. i think it should be:

        @param  _bondAmount,
        @param  _openVaultSharePrice,
        @param  _closeVaultSharePrice,
        @param  _vaultSharePrice,    <--- this one is out of order in the code
        @param  _isLong

jrhea (changes requested):
I think just one nit with natspec and add that test we talked about

contracts/src/internal/HyperdriveCheckpoint.sol
```diff
@@ -89,8 +69,33 @@ abstract contract HyperdriveCheckpoint is
...
+        // If the checkpoint time is the latest checkpoint, we use the current
+        // vault share price. Otherwise, we use a linear search to find the
+        // closest vault share price and use that to perform the checkpoint.
+        uint256 checkpointVaultSharePrice;
+        uint256 latestCheckpoint = _latestCheckpoint();
+        if (_checkpointTime == latestCheckpoint) {
+            checkpointVaultSharePrice = _vaultSharePrice;
+        } else {
+            for (
+                uint256 time = _checkpointTime;
```
MrToph, line 81: `time` could start at `_checkpointTime + _checkpointDuration` already because we know that `_checkpoints[_checkpointTime].vaultSharePrice == 0` due to the early return above and `_checkpointTime  < latestCheckpoint ` because we're in the `else`.

MrToph (commented):
what was wrong was that when we apply a checkpoint in the past, we didn't actually do the linear search (it was only done in `_checkpoint`) but used the current share price instead. From code like this.

```solidity
uint256 vaultSharePrice = _pricePerVaultShare();
if (block.timestamp < _maturityTime) {
    _applyCheckpoint(_latestCheckpoint(), vaultSharePrice);
} else {
    _applyCheckpoint(_maturityTime, vaultSharePrice);
}
```

jrhea (approved):

PR #925, opened 2024-04-02
`distributeExcessIdle` fixes

jalextowle:
# Resolved Issues

[Spearbit Issue #34](https://github.com/spearbit-audits/delv-review/issues/34)

# Description

This PR adds two checks to the `calculateDistributeExcessIdleShareProceeds` function. First, a check was added that verifies that the ending LP share price is within 1 basis point (`1e14`) of the starting LP share price. If the check fails, idle won't be distributed. Secondly, a check was added that each step of Newton's method is monotonically improving the calculations error. If a step begins to diverge, we break out of the loop to avoid pathological cases caused by rounding issues. 

Enforcing additional checks in `_distributeExcessIdleSafe` can hurt the system's liveness, so in addition to enforcing the checks, a `_maxIterations` parameter was added to `_distributeExcessIdleSafe` that configures the amount of iterations that `calculateDistributeExcessIdleShareProceeds` will use in Newton's method. This argument was also added to `checkpoint` so that checkpointers can specify the iterations used by `_distributeExcessIdleSafe`. `checkpoint` was updated to always distribute excess idle, which will allow LPs or other stakeholders to distribute excess idle with more iterations in the event that idle becomes stuck.

# Review Checklists

be removed.

### Solidity

- [ ] **Tokens**
              token spend?
- [ ] **Reentrancy**
- [ ] **Gas Optimizations**
          memory variables?
          issues?
- [ ] **Visibility**
- [ ] **Math**
          catch underflows?
- [ ] **Testing**

### Rust

- [ ] **Testing**

### Solidity

- [ ] **Tokens**
              token spend?
- [ ] **Reentrancy**
- [ ] **Gas Optimizations**
          memory variables?
          issues?
- [ ] **Visibility**
- [ ] **Math**
          catch underflows?
- [ ] **Testing**

### Rust

- [ ] **Testing**


contracts/src/libraries/LPMath.sol
```diff
@@ -887,15 +904,23 @@ library LPMath {
...
+            // since Newton's method has begun to diverge.
             int256 delta = presentValue.mulDown(lpTotalSupply).toInt256() -
                 _params
                     .startingPresentValue
                     .mulUp(_params.activeLpTotalSupply)
                     .toInt256();
+            if (lastDelta != 0 && delta.abs() > lastDelta.abs()) {
+                break;
+            }
+            lastDelta = delta;
```
jalextowle, line 918: This change is intended to improve our security assumptions so that people don't pass a huge `maxIterations` that leads to divergence and a larger LP share price deviation than we expect.

contracts/src/external/HyperdriveTarget4.sol
```diff
@@ -30,6 +30,39 @@ abstract contract HyperdriveTarget4 is
...
+    /// @param _minLpSharePrice The minimum LP share price the LP is willing
+    ///        to accept for their shares. LPs incur negative slippage when
+    ///        adding liquidity if there is a net curve position in the market,
+    ///        so this allows LPs to protect themselves from high levels of
+    ///        slippage. The units of this quantity are either base or vault
+    ///        shares, depending on the value of `_options.asBase`.
+    /// @param _minApr The minimum APR at which the LP is willing to supply.
+    /// @param _maxApr The maximum APR at which the LP is willing to supply.
+    /// @param _options The options that configure how the operation is settled.
+    /// @return lpShares The number of LP tokens created.
```
jrhea, line 48: this isn't a named return value

contracts/src/libraries/FixedPointMath.sol
```diff
@@ -376,4 +376,11 @@ library FixedPointMath {
     function max(int256 a, int256 b) internal pure returns (int256) {
         return a > b ? a : b;
     }
+
+    /// @dev Calculates the absolute value of an int256.
+    /// @param a The value.
+    /// @return The absolute value of the input.
+    function abs(int256 a) internal pure returns (int256) {
+        return a < 0 ? -a : a;
```
jrhea, line 384: maybe i'm missing it, but is there a unit test for this?

contracts/src/libraries/LPMath.sol
```diff
@@ -730,7 +743,11 @@ library LPMath {
         // Newton's method will terminate as soon as the current iteration is
         // within the minimum tolerance or the maximum number of iterations has
         // been reached.
-        for (uint256 i = 0; i < SHARE_PROCEEDS_MAX_ITERATIONS; ) {
+        int256 lastDelta;
+        if (_maxIterations < SHARE_PROCEEDS_MAX_ITERATIONS) {
```
jrhea, line 747: am i incorrect in assuming that most of the time `SHARE_PROCEEDS_MAX_ITERATIONS` will be the max iterations?  if so, is there a more gas efficient implementation?

contracts/src/libraries/LPMath.sol
```diff
@@ -838,7 +855,7 @@ library LPMath {
                             params.presentValueParams.shareAdjustment,
                             params.presentValueParams.bondReserves,
                             params.presentValueParams.minimumShareReserves,
-                            ONE - _params.presentValueParams.timeStretch,
+                            ONE - params.presentValueParams.timeStretch,
```
jrhea, line 858: what happened here?

contracts/src/libraries/FixedPointMath.sol
mcclurejt: not seeing it either

contracts/src/libraries/LPMath.sol
```diff
@@ -925,6 +950,48 @@ library LPMath {
...
+        }
+        uint256 presentValue_ = calculatePresentValue(
+            _params.presentValueParams
+        );
+
+        // Verify that the LP share price was conserved within a reasonable
+        // tolerance.
+        if (
+            // NOTE: Round up to make the check stricter.
+            presentValue_.divUp(_params.activeLpTotalSupply) +
```
mcclurejt, line 982: Thinking this is a similar issue to [here](https://github.com/spearbit-audits/delv-review/issues/30)....

here's how I think about it:

- increased strictness == more invalid values
- truthy check result == invalid value

Therefore `more truthy check results => more invalid values => increased strictness`

For this check, using `divDown` on the lhs and `divUp` on the rhs would result in the most truthy values

contracts/src/libraries/LPMath.sol
```diff
@@ -925,6 +950,48 @@ library LPMath {
...
+                SHARE_PROCEEDS_TOLERANCE <
+            _params.startingPresentValue.divDown(lpTotalSupply) ||
+            // NOTE: Round down to make the check stricter.
+            presentValue_.divDown(_params.activeLpTotalSupply) >=
```
mcclurejt, line 986: same as above but `divUp` on lhs and `divDown` on rhs

mcclurejt (changes requested):
left a couple comments regarding rounding for the `_updateLiquidity` checks

contracts/src/libraries/LPMath.sol
jalextowle: We ran into a `stack-too-deep` issue. We already had a new `params` value, and I just missed this value in previous passes.

contracts/src/libraries/LPMath.sol
jalextowle: That's correct. 

Assuming we want to verify the value, there isn't a more efficient implementation. We could just use the provided `_maxIterations`, but I find that to be sketchy since future code changes could erroneously pass a `_maxIterations` value that is smaller than `LPMath.SHARE_PROCEEDS_MAX_ITERATIONS`.

mcclurejt (approved):
LGTM

jrhea (approved):

jrhea (approved):
approved, but i think i found one nit on a named return type in natspec

contracts/src/libraries/LPMath.sol
```diff
@@ -887,15 +906,25 @@ library LPMath {
...
+            int256 delta = presentValue.mulDown(lpTotalSupply).toInt256() -
+                params
+                    .startingPresentValue
+                    .mulUp(params.activeLpTotalSupply)
+                    .toInt256();
+            if (smallestDelta == 0 || delta.abs() < smallestDelta.abs()) {
+                smallestDelta = delta;
+                closestShareProceeds = shareProceeds;
+                closestPresentValue = presentValue;
+            }
```
MrToph, line 922: just adding my explanation here: this updates the solution candidate whenever `|F(dz)|` gets smaller, meaning `F(dz)` gets closer to `0`, the exact solution.

contracts/src/libraries/LPMath.sol
```diff
@@ -925,7 +954,60 @@ library LPMath {
...
+            closestPresentValue = presentValue_;
+        }
+
+        // Verify that the LP share price was conserved within a reasonable
+        // tolerance.
+        if (
+            // NOTE: Round down to make the check stricter.
+            closestPresentValue.divDown(params.activeLpTotalSupply) +
+                SHARE_PROCEEDS_TOLERANCE <
+            params.startingPresentValue.divUp(lpTotalSupply) ||
```
MrToph, line 999: the `shouldShortCircuitDistributeExcessIdleShareProceeds` does not allow a tolerance for a new lower value at all. I thought this was to always ensure the invariant `newLpPrice >= oldLpPrice`

MrToph: But I see how this makes more sense, if the Newton approaches it from the "wrong side" and ends up with a slightly smaller value, it shouldn't reject the update.
Question is if `shouldShortCircuitDistributeExcessIdleShareProceeds` should also add this.

contracts/src/libraries/LPMath.sol
```diff
@@ -925,7 +954,60 @@ library LPMath {
...
+            // NOTE: Round up to make the check stricter.
+            closestPresentValue.divUp(params.activeLpTotalSupply) >=
```
MrToph, line 1001: technically should use `>` instead of `>=` such that if you are _at_ the tolerance, it's still accepted. That's how all the other code here works.

contracts/src/libraries/LPMath.sol
```diff
@@ -925,7 +954,60 @@ library LPMath {
...
+            params.startingPresentValue.divDown(lpTotalSupply) +
+                SHARE_PROCEEDS_TOLERANCE
```
MrToph, line 1003: you're using a different tolerance check here (additive / absolute error) than in `shouldShortCircuitDistributeExcessIdleShareProceeds` (multiplicative / relative error):

`PV_1 / l_1 >= (PV_0 / l_0) + tolerance` vs `PV_1 / l_1 >= (PV_0 / l_0) * (1 + tolerance)`.

contracts/src/libraries/LPMath.sol
jalextowle: I don't think so. I think short-circuiting should be stricter IMO.

contracts/src/libraries/LPMath.sol
jalextowle: Hmm, yeah that's fair. I'll switch it to multiplicative.

jrhea (approved):
reapproved for codesize

MrToph (approved):

PR #926, opened 2024-04-02
add `calculate_spot_rate_after_long` and general cleanup

dpaiton:
# Resolved Issues
Addresses one component required for https://github.com/delvtech/hyperdrive/issues/775

# Description
- Adds a helper function to calculate the spot rate after a long is opened
- Cleans up rust compiler warnings about unused imports
- Adds an optional `bond_amount` argument to `calculate_spot_price_after_long` to prevent duplicate calls to `open_long` and to better match the signature for `calculate_spot_price_after_short`.
- Removes deprecated `calculate_short_deposit` (used to  be `get_short_deposit`) that has not been recommended since `0.4.0`

I didn't fuzz test this against solidity because it uses the rust open long flow, which is tested against solidity. This follows the same pattern as `calculate_spot_price_after_long`.

# Review Checklists

be removed.

### Rust

- [ ] **Testing**

### DIFF
--- a/crates/hyperdrive-math/src/lib.rs
+++ b/crates/hyperdrive-math/src/lib.rs
@@ -6,14 +6,12 @@ mod yield_space;
 
 use ethers::types::{Address, I256, U256};
 use fixed_point::FixedPoint;
-use fixed_point_macros::{fixed, uint256};
+use fixed_point_macros::fixed;
 use hyperdrive_wrappers::wrappers::ihyperdrive::{Fees, PoolConfig, PoolInfo};
-pub use long::*;
 use rand::{
     distributions::{Distribution, Standard},
     Rng,
 };
-pub use short::*;
 pub use utils::*;
 pub use yield_space::YieldSpace;
 
--- a/crates/hyperdrive-math/src/lib.rs
+++ b/crates/hyperdrive-math/src/lib.rs
@@ -124,7 +124,7 @@ impl State {
         let annualized_time =
             self.position_duration() / FixedPoint::from(U256::from(60 * 60 * 24 * 365));
         let spot_price = self.calculate_spot_price();
-        (fixed!(1e18) - spot_price) / (spot_price * annualized_time)
+        calculate_fixed_rate_from_price(spot_price, annualized_time)
     }
 
     /// Converts a timestamp to the checkpoint timestamp that it corresponds to.
--- a/crates/hyperdrive-math/src/lib.rs
+++ b/crates/hyperdrive-math/src/lib.rs
@@ -119,12 +119,12 @@ impl State {
         YieldSpace::calculate_spot_price(self)
     }
 
-    /// Calculates the pool's spot rate.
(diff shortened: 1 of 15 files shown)


crates/hyperdrive-math/src/long/open.rs
```diff
@@ -44,8 +46,15 @@ impl State {
...
+    pub fn calculate_spot_price_after_long(
+        &self,
+        base_amount: FixedPoint,
+        bond_amount: Option<FixedPoint>,
+    ) -> FixedPoint {
+        let bond_amount = match bond_amount {
+            Some(bond_amount) => bond_amount,
+            None => self.calculate_open_long(base_amount),
+        };
         self.spot_price_after_long(base_amount, bond_amount)
```
ryangoree, line 58: Now that this function can take a `bond_amount`, we can probably consolidate it with the `spot_price_after_long` function below to avoid confusion about which one to use.

crates/hyperdrive-math/src/long/open.rs
```diff
@@ -61,6 +70,29 @@ impl State {
...
+    /// for `base_amount` is opened.
+    pub fn calculate_spot_rate_after_long(
+        &self,
+        base_amount: FixedPoint,
+        bond_amount: Option<FixedPoint>,
+    ) -> FixedPoint {
+        let annualized_time =
+            self.position_duration() / FixedPoint::from(U256::from(60 * 60 * 24 * 365));
+        let resulting_price = self.calculate_spot_price_after_long(base_amount, bond_amount);
+        (fixed!(1e18) - resulting_price) / (resulting_price * annualized_time)
```
ryangoree, line 94: It might be worth abstracting the conversion from spot price to rate since this is duplicated in `calculate_spot_rate` and will be duplicated in `calculate_spot_rate_after_short`.

ryangoree (approved):
Nice! 🚀

crates/hyperdrive-math/src/lib.rs
```diff
@@ -124,7 +124,7 @@ impl State {
         let annualized_time =
             self.position_duration() / FixedPoint::from(U256::from(60 * 60 * 24 * 365));
         let spot_price = self.calculate_spot_price();
-        (fixed!(1e18) - spot_price) / (spot_price * annualized_time)
+        calculate_fixed_rate_from_price(spot_price, annualized_time)
```
ryangoree, line 127: This is nice, but I was thinking something more along the lines of:

```rs
pub fn calculate_spot_rate(&self, spot_price: Option<FixedPoint>) -> FixedPoint {
      let annualized_time =
          self.position_duration() / FixedPoint::from(U256::from(60 * 60 * 24 * 365));
      let spot_price = match spot_price {
          Some(spot_price) => spot_price,
          None => self.calculate_spot_price(),
      };
      (fixed!(1e18) - spot_price) / (spot_price * annualized_time)
  }
```

This way we could easily convert with `self.calculate_spot_rate(arbitrary_spot_price)`

PR #930, opened 2024-04-03
HyperdriveRegistry deploy

mcclurejt:
# Resolved Issues

- fixes https://github.com/delvtech/hyperdrive/issues/922

# Description

Since anyone can deploy instances of Hyperdrive, it is difficult to determine and keep track of which instances are known and reputable.  To alleviate this, the `HyperdriveRegistry` contract is permissioned and includes an additional `data` field along with the instance address to give additional context. 

# Review Checklists

be removed.

### Rust

- [ ] **Testing**

### Rust

- [ ] **Testing**


jalextowle (approved):
LGTM. Nice dude!

crates/test-utils/src/chain/test_chain.rs
```diff
@@ -1016,19 +1023,35 @@ impl TestChain {
...
+        // and the registry to the admin address now that we're done
+        // minting tokens and updating the configuration.
         base.transfer_ownership(config.admin).send().await?;
         vault.transfer_ownership(config.admin).send().await?;
         lido.transfer_ownership(config.admin).send().await?;
         factory.update_governance(config.admin).send().await?;
+        hyperdrive_registry
+            .update_governance(config.admin)
+            .send()
+            .await?;
```
jalextowle, line 1047: One problem that we may run into is that the devnet `admin` defaults to `address(0)`, which would prevent the registry from being updated on `devnet`. I think we'll want to update `.github/workflows/devnet.yaml` so that it has an `ADMIN` environment variable similar to `.github/workflows/testnet.yaml`. We can use a Github variable or just put a comment with the private key in the workflow since this is used for developer purposes.

jalextowle: With this in mind (and looking at the review checklist), let's add some tests that verify that we can actually update the registry.

crates/test-utils/src/chain/test_chain.rs
```diff
@@ -1016,19 +1023,35 @@ impl TestChain {
...
+        // Add the 4626 Hyperdrive instance and the Lido Hyperdrive instance
+        // to the registry contract.
+        hyperdrive_registry
+            .set_hyperdrive_info(vault.address(), uint256!(1))
+            .send()
+            .await?;
+        hyperdrive_registry
+            .set_hyperdrive_info(lido.address(), uint256!(1))
+            .send()
+            .await?;
```
jalextowle, line 1035: Let's add some assertions in the test cases that verifies that the registry was properly updated.

jalextowle (commented):
Looks like I approved early. Let me know when this is addressed, and I'll take another look.

mcclurejt:
> LGTM. Nice dude!
> 
> One problem that we may run into is that the devnet `admin` defaults to `address(0)`, which would prevent the registry from being updated on `devnet`. I think we'll want to update `.github/workflows/devnet.yaml` so that it has an `ADMIN` environment variable similar to `.github/workflows/testnet.yaml`. We can use a Github variable or just put a comment with the private key in the workflow since this is used for developer purposes.

Awesome! Went for the "storing it in github variables" approach. Made the corresponding addition to the workflow as well.

jalextowle (approved):

PR #934, opened 2024-04-03
Add `calculate_add_liquidity` to rust sdk.

sentilesdal:
# Resolved Issues
https://github.com/orgs/delvtech/projects/13?pane=issue&itemId=58015840

# Description
Adds the rust function `calculate_add_liquidity`

# Review Checklists

be removed.

### Rust

- [ ] **Testing**

### DIFF
--- a/crates/hyperdrive-math/src/lp.rs
+++ b/crates/hyperdrive-math/src/lp.rs
@@ -1,12 +1,161 @@
-use std::cmp::{max, Ordering};
+use eyre::{eyre, Result};
+use std::cmp::Ordering;
 
 use ethers::types::{I256, U256};
 use fixed_point::FixedPoint;
 use fixed_point_macros::{fixed, int256};
 
-use crate::{get_effective_share_reserves, State, YieldSpace};
+use crate::{calculate_effective_share_reserves, State, YieldSpace};
 
 impl State {
+    // Calculates the lp_shares for a given contribution when adding liquidity.
+    pub fn calculate_add_liquidity(
+        &self,
+        current_block_timestamp: U256,
+        contribution: FixedPoint,
+        min_lp_share_price: FixedPoint,
+        min_apr: FixedPoint,
+        max_apr: FixedPoint,
+        as_base: bool,
+    ) -> Result<FixedPoint> {
+        // Ensure that the contribution is greater than or equal to the minimum
+        // transaction amount.
+        if contribution < self.minimum_transaction_amount() {
+            return Err(eyre!(
+                "MinimumTransactionAmount: Contribution is smaller than the minimum transaction."
+            ));
+        }
+
+        // Enforce the slippage guard.
+        let apr = self.calculate_spot_rate();
+        if apr < min_apr || apr > max_apr {
+            return Err(eyre!("InvalidApr: Apr is outside the slippage guard."));
+        }
+
+        // Get lp_total_supply for the lp_shares calculation.
+        let lp_total_supply = self.lp_total_supply();
+
+        // Get the starting_present_value.
+        let starting_present_value = self.calculate_present_value(current_block_timestamp);
+
+        // Get the ending_present_value.
+        let share_contribution = {
+            if as_base {
+                // Attempt a crude conversion from base to shares.
+                I256::try_from(contribution / self.vault_share_price()).unwrap()
+            } else {
+                I256::try_from(contribution).unwrap()
+            }
+        };
+        let new_state = self.get_state_after_liquidity_update(share_contribution);
+        let ending_present_value = new_state.calculate_present_value(current_block_timestamp);
+
+        // Ensure the present value didn't decrease after adding liquidity.
+        if ending_present_value < starting_present_value {
+            return Err(eyre!("DecreasedPresentValueWhenAddingLiquidity: Present value decreased after adding liquidity."));
+        }
+
+        // Calculate the lp_shares.
+        let lp_shares = (ending_present_value - starting_present_value)
+            .mul_div_down(lp_total_supply, starting_present_value);
+
+        // Ensure that enough lp_shares are minted so that they can be redeemed.
+        if lp_shares < self.minimum_transaction_amount() {
+            return Err(eyre!(
+                "MinimumTransactionAmount: Not enough lp shares minted."
+            ));
+        }
+
+        // Enforce the minimum LP share price slippage guard.
+        if contribution.div_down(lp_shares) < min_lp_share_price {
+            return Err(eyre!("OutputLimit: not enough lp shares minted."));
+        }
+
+        Ok(lp_shares)
+    }
+
+    // Gets the resulting state when updating liquidity.
+    pub fn get_state_after_liquidity_update(&self, share_reserves_delta: I256) -> State {
+        let share_reserves = self.share_reserves();
+        let share_adjustment = self.share_adjustment();
+        let bond_reserves = self.bond_reserves();
+        let minimum_share_reserves = self.minimum_share_reserves();
+
+        // Calculate new reserve and adjustment levels.
+        let (updated_share_reserves, updated_share_adjustment, updated_bond_reserves) = self
+            .calculate_update_liquidity(
+                share_reserves,
+                share_adjustment,
+                bond_reserves,
+                minimum_share_reserves,
+                share_reserves_delta,
+            )
+            .unwrap();
+
+        // Update and return the new state.
+        let mut new_info = self.info.clone();
+        new_info.share_reserves = U256::from(updated_share_reserves);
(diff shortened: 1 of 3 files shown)


crates/hyperdrive-math/src/lp.rs
```diff
@@ -4,9 +4,144 @@ use ethers::types::{I256, U256};
...
+        current_block_timestamp: U256,
+        contribution: FixedPoint,
+        min_apr: FixedPoint,
+        max_apr: FixedPoint,
+        as_base: bool,
+    ) -> Result<FixedPoint, &str> {
+        // Ensure that the contribution is greater than or equal to the minimum
+        // transaction amount.
+        if contribution < self.minimum_transaction_amount() {
+            return Err("MinimumTransactionAmount");
```
dpaiton, line 22: we're using Eyre for our error messages, not just Err. No idea what the difference is, but this is the established pattern:

```rs
if self.y() < dy {
    return Err(eyre!(
        "calculate_shares_in_given_bonds_out_up_safe: y = {} < {} = dy",
        self.y(),
        dy,
    ));
}
```

crates/hyperdrive-math/src/lp.rs
```diff
@@ -4,9 +4,144 @@ use ethers::types::{I256, U256};
...
+        }
+
+        // Enforce the slippage guard.
+        let apr = self.calculate_spot_rate();
+        if apr < min_apr || apr > max_apr {
+            return Err("InvalidApr");
```
dpaiton, line 28: I think something like "Pool APR is outside the provided constraints." would be more clear to the user.

dpaiton: Ahh is `InvalidApr` the solidity error? I don't think we have been matching their error codes exactly in any of the other rust implementations, although that is an interesting idea... Is that your intent here? I'm wondering if we could do both? Like include the SOL error code but also add context.

crates/hyperdrive-math/src/lp.rs
```diff
@@ -4,9 +4,144 @@ use ethers::types::{I256, U256};
...
+                // Attempt a crude conversion from base to shares.
+                I256::try_from(contribution / self.vault_share_price()).unwrap()
+            } else {
+                I256::try_from(contribution).unwrap()
+            }
+        };
+        let new_state = self.update_liquidity(share_contribution);
+        let ending_present_value = new_state.calculate_present_value(current_block_timestamp);
+
+        // Ensure the present value didn't decreased after adding liquidity.
```
dpaiton, line 49: typo -- "decrease"

crates/hyperdrive-math/src/lp.rs
```diff
@@ -4,9 +4,144 @@ use ethers::types::{I256, U256};
...
+        }
+
+        // Calculate the lp_shares.
+        let lp_shares = (ending_present_value - starting_present_value)
+            .mul_div_down(lp_total_supply, starting_present_value);
+
+        // Ensure that enough lp_shares are minted so that they can be redeemed.
+        if lp_shares < self.minimum_transaction_amount() {
+            return Err("MinimumTransactionAmount");
+        }
```
dpaiton, line 61: What about this check from the solidity?

```sol
        // NOTE: Round down to make the check more conservative.
        //
        // Enforce the minimum LP share price slippage guard.
        if (_contribution.divDown(lpShares) < _minLpSharePrice) {
            revert IHyperdrive.OutputLimit();
        }
```

crates/hyperdrive-math/src/lp.rs
```diff
@@ -4,9 +4,144 @@ use ethers::types::{I256, U256};
...
+
+        return Ok(lp_shares);
+    }
+
+    // Gets the resulting state when updating liquidity.
+    pub fn update_liquidity(&self, share_reserves_delta: I256) -> State {
```
dpaiton, line 67: I think a better name for this would be something like `get_state_after_liquidity_update`. To me `update_liquidity` implies that it would happen in-place.

crates/hyperdrive-math/src/lp.rs
```diff
@@ -4,9 +4,144 @@ use ethers::types::{I256, U256};
...
+    ) -> Result<(FixedPoint, I256, FixedPoint), &'static str> {
+        if share_reserves_delta == I256::zero() {
+            return Ok((share_reserves, share_adjustment, bond_reserves));
+        }
+
+        // Get the updated share reserves.
+        let new_share_reserves = if share_reserves_delta > I256::zero() {
+            share_reserves + FixedPoint::from(share_reserves_delta)
+        } else {
+            share_reserves - FixedPoint::from(share_reserves_delta)
```
dpaiton, line 113: I think we want a check in here that `share_reserves > share_reserves_delta`, otherwise we can get a FixedPoint negative number error. You could combine this with the check on line 117:

```rs
if share_reserves > FixedPoint::from(share_reserves_delta) + minimum_share_reserves {
  share_reserves - FixedPoint::from(share_reserves_delta)
} else {
  return Err("Update would result in share reserves below minimum.");
}
```

crates/hyperdrive-math/src/lp.rs
sentilesdal: Yeah that was the intent, makes matching errors easier when debugging.  Happy to do what you propose and have it be more clear like "InvalidApr: apr is outside the provided constraints."

crates/hyperdrive-math/src/lp.rs
sentilesdal: nice call, this cleans up the return signature of the function.

dpaiton (approved):

crates/hyperdrive-math/src/lp.rs
```diff
@@ -189,15 +334,182 @@ impl State {
...
+        let mut rng = thread_rng();
+        let chain = TestChain::new(2).await?;
+        let (alice, bob) = (chain.accounts()[0].clone(), chain.accounts()[1].clone());
+        let mut alice =
+            Agent::new(chain.client(alice).await?, chain.addresses().clone(), None).await?;
+        let mut bob = Agent::new(chain.client(bob).await?, chain.addresses(), None).await?;
+        let config = bob.get_config().clone();
+
+        // Test happy paths.  Limit this to 100 runs as it takes a long time.
+        for _ in 0..100 {
```
ryangoree, line 447: Looks like we can use `FUZZ_RUNS`, which is set to `100` for longer tests like this (vs `FAST_FUZZ_RUNS` which is `10_000`).

crates/hyperdrive-math/src/lp.rs
```diff
@@ -189,15 +334,182 @@ impl State {
...
+                    FixedPoint::from(U256::MAX),
+                    true,
+                )
+                .unwrap();
+
+            // Rust can't account for slippage.
+            assert!(lp_shares >= lp_shares_mock, "Should over estimate.");
+            // Answer should still be mostly the same.
+            assert!(
+                fixed!(1e18) - lp_shares_mock / lp_shares < fixed!(1e11),
```
ryangoree, line 500: ooo, I like this one-liner for checking the delta

ryangoree (approved):

crates/hyperdrive-math/src/lp.rs
```diff
@@ -1,12 +1,161 @@
...
+        }
+
+        // Enforce the slippage guard.
+        let apr = self.calculate_spot_rate();
+        if apr < min_apr || apr > max_apr {
+            return Err(eyre!("InvalidApr: Apr is outside the slippage guard."));
+        }
+
+        // Get lp_total_supply for the lp_shares calculation.
+        let lp_total_supply = self.lp_total_supply();
```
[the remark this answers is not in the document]
sentilesdal: No, I thought so too, but if you look in HyperdriveTarget0.sol, we return this formula as lpTotalSupply in PoolInfo.  Since `self.lp_total_supply()` is reading from pool info, we assume this formula is baked in.

crates/hyperdrive-math/src/lp.rs
```diff
@@ -1,12 +1,161 @@
...
+        // Enforce the minimum LP share price slippage guard.
+        if contribution.div_down(lp_shares) < min_lp_share_price {
+            return Err(eyre!("OutputLimit: Not enough lp shares minted."));
+        }
+
+        Ok(lp_shares)
+    }
+
+    // Gets the resulting state when updating liquidity.
+    pub fn get_state_after_liquidity_update(&self, share_reserves_delta: I256) -> State {
```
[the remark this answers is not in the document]
sentilesdal: That's what I had before ;).  I'm going to land as-is per Dylan's PR request.  I like your idea though, let's just get everyone in agreement first.

PR #938, opened 2024-04-04
fixes bug in lp net curve trade flow

dpaiton:
# Resolved Issues
https://github.com/delvtech/hyperdrive/issues/927

# Description
The `calculate_net_curve_trade` was intermittently failing in CI. The failure was due to a uncommonly entered branch that was missing... another branch... (there are a few nested branching paths).

I was able to consistently see the bug by upping the number of fuzz tests to 100k. I just ran 100k and was not able to see it again, so I'm _pretty sure_ it's fixed. [Here's](https://github.com/delvtech/hyperdrive/blob/4f701159a795a8131fe6e1284b371e8c97218b6b/contracts/src/libraries/LPMath.sol#L356) the relevant solidity code.

# Review Checklists

Please check each item **before approving** the pull request. While going
through the checklist, it is recommended to leave comments on items that are
referenced in the checklist to make sure that they are reviewed. If there are
multiple reviewers, copy the checklists into sections titled `## [Reviewer Name]`.
If the PR doesn't touch Solidity and/or Rust, the corresponding checklist can
be removed.

### Rust

- [ ] **Testing**
    - [ ] Are there new or updated unit or integration tests?
    - [ ] Do the tests cover the happy paths?
    - [ ] Do the tests cover the unhappy paths?
    - [ ] Are there an adequate number of fuzz tests to ensure that we are
          covering the full input space?
    - [ ] If matching Solidity behavior, are there differential fuzz tests that
          ensure that Rust matches Solidity?

PR #939, opened 2024-04-04
Bump tolerance for `fuzz_calculate_spot_price_after_short`

ryangoree:
# Description

The `fuzz_calculate_spot_price_after_short` test was intermittently failing with it's current tolerance of `fixed!(1e10)`. This PR bumps it to `fixed!(1e11)`.

### Rust

- [ ] **Testing**

### DIFF
--- a/crates/hyperdrive-math/src/short/open.rs
+++ b/crates/hyperdrive-math/src/short/open.rs
@@ -145,8 +145,7 @@ mod tests {
             } else {
                 expected_spot_price - actual_spot_price
             };
-            // TODO: Why can't this pass with a tolerance of 1e9?
-            let tolerance = fixed!(1e10);
+            let tolerance = fixed!(1e11);
 
             assert!(
                 delta < tolerance,


mcclurejt:
any idea as to why this started happening?

dpaiton (approved):

dpaiton:
> any idea as to why this started happening?

I think 1e11 (aka correct up to the 8th decimal) is pretty reasonable given that the spot price itself is an approximation of the realized price for a non-infinitesimal trade size.

That being said, AFAIK this was not failing before the latest flurry of activity in the rust repo. I'm going to approve to unblock other PRs from landing with this failing, but I think we should keep it in mind in case we start seeing more failures.

ryangoree:
> That being said, AFAIK this was not failing before the latest flurry of activity in the rust part of the repo.

Because the test added in the flurry #919

dpaiton:
I absorbed this PR into https://github.com/delvtech/hyperdrive/pull/938 so that we could have both fixes in one PR and CI would pass. Closing this issue.

Issue #858, opened 2024-03-19
`test-utils` overhaul


jalextowle:
A major area of technical debt in the Rust codebase is the `test-utils` crate. The abstractions that are currently being used are brittle and hard to work with. They don't lend themselves well to working with a remote chain, and they are tightly coupled to a deployment that uses `anvil`. The `Agent` struct and methods leaves things to be desired and is an area that could be improved. The major areas to improve are:

1. Re-work the `Chain` abstraction so that it works equally well for local and remote deployments. We may not even need something like the `Chain` trait, and we may be better off working with a set of orthogonal tools that we can use to build `ether-rs` clients with signer, nonce management, and error handling middleware that makes it trivial to get a provider, signer, and client set up for an arbitrary chain with arbitrary signer metadata. 
2. Re-work the `DevChain` abstraction. Instead of tightly coupling address lookups to a `DevChain` struct, it would be better if this was a utility that could be called to read from the artifacts server. That way we could mix and match test utils regardless of whether we are interacting with a local anvil chain or an `infra` deployment.
3. Re-work the `TestChain` abstraction. Instead of tightly coupling pool deployment to a `TestChain` struct, it would be better to have separate deployment functions that make it easy to deploy the factory, deployer coordinators, and pools. It would be nice to still have an integrated deployment flow (we'll still need this to build the docker images), but this should be more flexible. The benefit of doing this is that it would make it easier to test `hyperdrive-math` with many different pool configurations.
4. Revamp the `Agent` struct. This struct is straddling the line between a tool that could actually be used to write trading bots and something that is only useful for testing. Some important logic is held in this struct (like the logic surrounding getting a conservative price for `get_max_short`) that should be upstreamed to `hyperdrive-math`. Generally speaking, we should just clean this struct up and make sure that it is well-tested and fully-featured.

dpaiton:
I'd like to point out that most of this has been implemented in the Python `IHyperdrive` class. That might adjust your perceived priority on updating the test-utils if we can run equivalent tests in Python already. I'd be happy to meet to elaborate further (this should also include @slundqui).

This also brings up a meta discussion that we should have on differentiating the python & rust feature sets. While it makes sense for there to be parity in many areas (e.g. everything that we wrap via hyperdrivepy), there are other areas where we should take care to avoid duplicate work. That being said, I realize that testing in particular might require duplicates since we want most tests for rust fns to live in rust.

jalextowle:
[...]

PR #947, opened 2024-04-05
Refactor Test Utils

jalextowle:
# Resolved Issues

Fixes: #858.

# Description

While it may not be apparent when implementing the typical differential testing test, the current test utilities are quite cumbersome to work with. In retrospect, I think the abstractions came together too early, and it has been a pain to implement things like migration scripts and to consider doing more with the tools. The goal of this PR was to decouple the utilities of `TestChain` and `DevChain` to simplify the DevEx of interacting with the Rust repo. The guiding principle was to preserve as much "orthogonality" as possible.

While it may not look like it when reviewing the `hyperdrive-math` tests, for example, the tools are now much easier to work with. If you want to connect to a remote chain and deploy something, no problem. You don't have to use a `TestChain`, and you don't have to bundle signers into it. If you want to get the addresses from a `DevChain`, no problem. Just call the function. 

To keep ergonomics relatively constant (I actually think they have improved), I re-implemented `TestChain` using the new utilities and made some test accounts like `ALICE`, `BOB`, and `CELINE` first-class citizens. It's now easier to set up a test since account management isn't a concern anymore. This significantly reduces the number of imports required. Perhaps more importantly, the tools are more orthogonal, which sets us up for future improvements. 

There is more that could and should be improved with these tools, but I didn't want the PR to be too large (it's already fairly large as-is). I'm going to need this orthogonality to implement the testnet migration script.

# Review Checklists

be removed.

### Rust

- [ ] **Testing**


crates/test-utils/src/chain/deploy.rs
```diff
@@ -0,0 +1,1050 @@
...
+                governance_zombie: test_chain_config.erc4626_hyperdrive_governance_zombie_fee,
+            }
+        );
+
+        // Verify that the steth pool config is correct.
+        let hyperdrive = IHyperdrive::new(addresses.steth_hyperdrive, client.clone());
+        let config = hyperdrive.get_pool_config().call().await?;
+        assert_eq!(
+            config.base_token,
+            "0xEeeeeEeeeEeEeeEeEeEeeEEEeeeeEeeeeeeeEEeE".parse::<Address>()?
```
sentilesdal, line 885: add this to constants.rs?

crates/test-utils/src/chain/deploy.rs
```diff
@@ -0,0 +1,1050 @@
...
+                        HyperdriveFactoryEvents::decode_log(&log.into())
+                    {
+                        Some(log)
+                    } else {
+                        None
+                    }
+                })
+                .collect::<Vec<_>>();
+            logs[0].clone().hyperdrive
+        };
```
mcclurejt, line 783: What do you think about abstracting out the deployment/initialization of hyperdrive instances to a separate function? The processes seem relatively similar and mechanical for the current instances... would also be a great place to handle adding to the registry. 

Cool with punting on this if we expect dramatic changes just worth thinking about

mcclurejt (commented):
Not in love with getting rid of examples, I find them super helpful whenever I'm exploring a new sdk.

PR #950, opened 2024-04-05
Fixing `calc_max_short` for accuracy

slundqui:
# Resolved Issues
General updates towards https://github.com/delvtech/hyperdrive/issues/871

# Description
This PR does the following updates:
- Adds a test comparing calculated derivative of `long_amount_derivative` and `short_deposit_derivative` emperically.
    - Further work needs to be done here to decrease comparison epsilon
- Adds a test ensuring the absolute max short is a valid short size on chain.
- Implements the max short derivative calculation from https://github.com/delvtech/hyperdrive/pull/733.
- Fixes for the case where the initial max short guess made the pool insolvent.
- Fixes for convergence of `calculate_max_short`.
- ~Increasing the default number of iterations from 7 to 10 for `calculate_max_short`.~ Solved issue where we needed more iterations, so the default number of iterations is not changed.

# Review Checklists

be removed.

### Rust

- [ ] **Testing**

### DIFF
--- a/crates/hyperdrive-math/src/short/max.rs
+++ b/crates/hyperdrive-math/src/short/max.rs
@@ -81,12 +81,12 @@ impl State {
         let mut max_bond_amount =
             self.absolute_max_short(spot_price, checkpoint_exposure, maybe_max_iterations);
         let absolute_max_bond_amount = max_bond_amount;
-        let deposit =
+        let absolute_max_deposit =
             match self.calculate_open_short(max_bond_amount, spot_price, open_vault_share_price) {
                 Ok(d) => d,
                 Err(_) => return max_bond_amount,
             };
-        if deposit <= budget {
+        if absolute_max_deposit <= budget {
             return max_bond_amount;
         }
 
@@ -116,9 +116,15 @@ impl State {
             open_vault_share_price,
             maybe_conservative_price,
         );
-        let mut previous_max_bond_amount = max_bond_amount;
-        let mut step_size = fixed!(1e18);
-        for _ in 0..maybe_max_iterations.unwrap_or(10) {
+        let mut best_valid_max_bond_amount = max_bond_amount;
+
+        // To avoid the case where Newton's method overshoots and stays on
+        // the invalid side of the optimization equation (i.e., when deposit > budget),
+        // we artificially set the target budget to be less than the actual budget.
+        // We use the minimum transaction amount here to avoid the underflow
+        // where budget is less than this bias.
+        let target_budget = budget - self.minimum_transaction_amount();
+        for _ in 0..maybe_max_iterations.unwrap_or(7) {
             let deposit = match self.calculate_open_short(
                 max_bond_amount,
                 spot_price,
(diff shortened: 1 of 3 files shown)


crates/hyperdrive-math/src/short/max.rs
```diff
@@ -115,21 +116,41 @@ impl State {
...
             };
-            max_bond_amount += (budget - deposit)
-                / self.short_deposit_derivative(
-                    max_bond_amount,
-                    spot_price,
-                    open_vault_share_price,
-                );
+
+            // If we overshot, we undo a step and cut the
+            // step size by half
```
jalextowle, line 137: This is fine IMO, but it might be better to just proceed with Newton's method and properly handle crossing over. We'd just do something like this:

```
if budget >= deposit {
                max_bond_amount += (budget - deposit)
                / self.short_deposit_derivative(
                    max_bond_amount,
                    spot_price,
                    open_vault_share_price,
                );
} else {
                    max_bond_amount -= (deposit - budget)
                / self.short_deposit_derivative(
                    max_bond_amount,
                    spot_price,
                    open_vault_share_price,
                );
}
```

The derivative might need to be updated in this case, but you could just switch to `I256`. It's up to y'all, but Newton's method is going to have better convergence properties.

jalextowle: You can see what we do here: https://github.com/delvtech/hyperdrive/blob/ed2bc6d4387d6de00670869ccea8a548873bacd5/contracts/src/libraries/LPMath.sol#L862.

slundqui: The issue with this is that only one side is valid, so we really only want to approach from one side only (the side where budget >= deposit). Recovering back to a valid state when on the other side is more difficult.

jalextowle: We can just record the best value on the right side. It'll look similar to what we do in `LPMath`. Just because we only want to take values less than the max doesn't mean we can't leverage Newton's convergence. It should start flipping back and forth.

It's just a suggestion.

slundqui: Implemented the suggestion above, with an added bias to the budget to avoid the case where newton's method never crosses back over to the valid side.

dpaiton (approved):
Thanks for this! I think @jalextowle should give it a once-over to verify that the error tolerances (`max.rs:593`) are reasonable.

crates/hyperdrive-math/src/short/max.rs
```diff
@@ -115,33 +116,73 @@ impl State {
...
             };
-            max_bond_amount += (budget - deposit)
-                / self.short_deposit_derivative(
-                    max_bond_amount,
-                    spot_price,
-                    open_vault_share_price,
-                );
+
+            let derivative =
+                self.short_deposit_derivative(max_bond_amount, spot_price, open_vault_share_price);
```
jalextowle, line 144: Can we move this closer to where it is used? If we move this below `// Iteratively update max_bond_amount via newton's method`, it would be easier to read IMO.

crates/hyperdrive-math/src/short/max.rs
```diff
@@ -115,33 +116,73 @@ impl State {
...
+
+            // We update the best valid max bond amount if the deposit amount
+            // is valid and the current guess is better than the current estimate.
+            if (deposit < budget) && (best_valid_max_bond_amount < max_bond_amount) {
+                best_valid_max_bond_amount = max_bond_amount;
+            }
+
+            // Iteratively update max_bond_amount via newton's method
```
jalextowle, line 152: NIT: 
```suggestion
            // Iteratively update max_bond_amount via newton's method.
```

crates/hyperdrive-math/src/short/max.rs
```diff
@@ -115,33 +116,73 @@ impl State {
...
+                )
                 .unwrap()
         {
             panic!("max short exceeded budget");
         }
 
-        max_bond_amount
+        // Ensure that the max bond amount is within the absolute max bond amount.
+        if best_valid_max_bond_amount > absolute_max_bond_amount {
+            best_valid_max_bond_amount = absolute_max_bond_amount;
```
jalextowle, line 182: We should probably throw an error here, right? We've already checked the `absolute_max_bond_amount` and it leads to a deposit larger than the budget.

crates/hyperdrive-math/src/short/max.rs
```diff
@@ -115,33 +116,73 @@ impl State {
...
-                    spot_price,
-                    open_vault_share_price,
-                );
+
+            let derivative =
+                self.short_deposit_derivative(max_bond_amount, spot_price, open_vault_share_price);
+
+            // We update the best valid max bond amount if the deposit amount
+            // is valid and the current guess is better than the current estimate.
+            if (deposit < budget) && (best_valid_max_bond_amount < max_bond_amount) {
```
jalextowle, line 148: I don't think the parenthesis are needed FWIW. This may cause a clippy warning or error.

jalextowle (approved):
LGTM once the comments are addressed

Issue #955, opened 2024-04-09
improve error handling in `targeted_long` for too-big and too-small targets


dpaiton:
There are bounds on the `target_rate` argument that are not observed presently. The rate upper bound is obvious -- they should not be targeting a rate that is greater than the current fixed rate, since longs can only reduce the rate. The lower bound is trickier -- in addition to the base amount being bound by the output of `calculate_maximum_long`, we can assume that there is a point near zero where the fixed rate is too small.

As of now if you provide pathological arguments for `target_rate` it will throw errors that are not sufficiently informative; it would be better to check that the `target_rate` argument is within a reasonable bounds.

dpaiton:
https://github.com/delvtech/hyperdrive/pull/956 partially addresses this by adding a check if the `target_rate` is above the `current_rate` (which means we would want a targeted _short_).

dpaiton:
Duplicate issue with https://github.com/delvtech/hyperdrive/issues/861 ; closing

PR #956, opened 2024-04-09
update calc open short & target long

dpaiton:
# Resolved Issues
partially addresses https://github.com/delvtech/hyperdrive/issues/955

# Description

- The `calculate_open_short` was written at a different time from `calculate_open_long`; I updated it so that they follow the same process.
edit: this caused trouble for reasons I can't explain but probably have to do with share/base conversions. I reverted it, but left some new error stuff.

- I also added some better error checking in `calculate_targeted_long` so that it can safely return in what was previously considered an error state.

- I added a new random test variable, the target rate itself, in the `calculate_targeted_long` tests.

# Review Checklists

be removed.

### Rust

- [ ] **Testing**

### DIFF
--- a/crates/hyperdrive-math/src/long/targeted.rs
+++ b/crates/hyperdrive-math/src/long/targeted.rs
@@ -495,7 +495,8 @@ mod tests {
 
             // Get a targeted long amount.
             // TODO: explore tighter bounds on this.
-            let target_rate = initial_fixed_rate / rng.gen_range(fixed!(1.0001e18)..=fixed!(10e18));
+            let target_rate = bob.get_state().await?.calculate_spot_rate()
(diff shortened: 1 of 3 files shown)


crates/hyperdrive-math/src/long/targeted.rs
```diff
@@ -513,18 +545,19 @@ mod tests {
             // 3. IF Bob's budget is not consumed; then new rate is close to the target rate
 
             // Check that our resulting price is under the max
-            let spot_price_after_long = bob.get_state().await?.calculate_spot_price();
+            let current_state = bob.get_state().await?;
+            let max_spot_price_before_long = current_state.calculate_max_spot_price();
```
ryangoree, line 549: Either the name or the value is wrong here. This is the max spot price *after* opening the long.

ryangoree (approved):

PR #960, opened 2024-04-11
Improves code coverage on a few tests

jalextowle:
# Description

Increases code coverage on the `contracts/src/external/Hyperdrive.sol` and `contracts/src/factory/HyperdriveFactory.sol`.


test/utils/InstanceTest.sol
```diff
@@ -238,6 +201,7 @@ abstract contract InstanceTest is HyperdriveTest {
 
         // Ensure that refunds are handled properly.
         if (config.enableBaseDeposits && asBase && isBaseETH) {
+            // Ensure that alice's ETH balance only decreased by the contribution.
```
jrhea, line 204:
```suggestion
            
            // Ensure that alice's ETH balance only decreased by the contribution.
```

jrhea (approved):
lgtm. just one fun nit

test/utils/InstanceTest.sol
jalextowle: We discussed this offline and resolved this.

PR #963, opened 2024-04-12
contract deployment + verification tooling

mcclurejt:
# Description

Contracts deployed via create2 don't automatically get verified in foundry, and they can be pretty difficult to get right manually. This automatically handles the verification process for a hyperdrive contract and its targets.

Explored a couple routes on how to best do this, and I liked the path of parsing TOML files and looping through for multi-chain deployments. 

At present, the bulk of it is implemented but there's still some cleanup and artifact producing work to go. Wanted to get this up and see how y'all feel about it. 

# Review Checklists

be removed.

### Tooling

- [ ] Do you see any ways devex could be improved?
- [ ] Does this cover the expected use cases?
- [ ] Is its functionality and documentation clear enough that you would feel comfortable modifying it?


script/verify-hyperdrive.sh
```diff
@@ -0,0 +1,220 @@
...
+####            ./script/verify-hyperdrive.sh \
+####              <foundry_profile> \
+####              <rpc_url> \
+####              <chain_id> \
+####              <etherscan_api_key> \
+####              <pool_type> \
+####              <pool_address> \
+####              <gov_address> \
+####
+################################################################################
```
jrhea, line 48: this is nice, but it might be nicer in a readme.md

.prettierrc
```diff
@@ -1,3 +0,0 @@
-{
-  "bracketSpacing": true
-}
```
jalextowle, line 3: Why'd we update this? Looks like it caused a bunch of changes in `cSpell.json`.

contracts/test/MockRocketPool.sol
```diff
@@ -115,13 +116,13 @@ contract MockRocketPool is MultiRolesAuthority, ERC20Mintable {
 
     function getSharesByPooledEth(
```
jalextowle, line 117: Hmm, do we even need these functions? Looks like this was copied from `MockLido`.

tasks/registry/index.ts
```diff
@@ -0,0 +1,40 @@
...
+};
+
+task("registry:add", "adds the specified hyperdrive instance to the registry")
+  .addParam(
+    "address",
+    "address of the hyperdrive instance to add",
+    undefined,
+    types.string,
+  )
+  .addOptionalParam(
```
jalextowle, line 15: Might be better for this to not be optional. I think we want people to think about this whenever they run the script

Makefile
```diff
@@ -74,3 +74,8 @@ lint-rust:
 
 prettier:
 	npx prettier --write .
+
+### Deploy ###
+
+deploy:
+	forge script ./script/deploy/Deploy.s.sol --broadcast --slow --verify --resume --multi --isolate --legacy
```
jalextowle, line 81: Is this going to be added, or should we delete this?

cSpell.json
```diff
@@ -1,188 +1,188 @@
 {
-    "version": "0.2",
```
jalextowle, line 2: Why was this reformatted? FWIW, I prefer 4 spaces since it's more similar to our Solidity spacing, and it feels less cramped. Was there a particular reason that we're using 2 spaces for Typescript formatting?

deploy/factory.ts
```diff
@@ -0,0 +1,139 @@
+import { HardhatRuntimeEnvironment } from "hardhat/types";
+import { DeployFunction } from "hardhat-deploy/types";
+import dayjs from "dayjs";
+import duration from "dayjs/plugin/duration";
+import { parseEther } from "viem";
+dayjs.extend(duration);
+
+const ADMIN = "0x25a6592F1eaA22cD7BFc590A90d56FbB765b1Edc";
+const GOVERNANCE = "0x25a6592F1eaA22cD7BFc590A90d56FbB765b1Edc";
```
jalextowle, line 9: Should these be environment variables or otherwise passed in?

deploy/factory.ts
```diff
@@ -0,0 +1,139 @@
+import { HardhatRuntimeEnvironment } from "hardhat/types";
```
jalextowle, line 1: This is a meta comment for all of these scripts. I'm not expecting these to be as heavily commented as the Solidity or Rust SDK -- this code is much more self-explanatory. That said, I'd like to see some high-level comments describing the different steps of the deployment.

deploy/factory.ts
```diff
@@ -0,0 +1,139 @@
...
+  const lido =
+    hre.network.name == "sepolia"
+      ? (
+          await deploy("MockLido", {
+            from: deployer,
+            args: [parseEther("0.035"), ADMIN, true, parseEther("500")],
+            log: true,
+          })
+        ).address
+      : LIDO_ADDRESS;
```
jalextowle, line 124: Shouldn't we be deploying a mock vault for the ERC4626 instances on Sepolia too? I get that we don't need it to instantiate the coordinator, but we'll still need to deploy it eventually.

deploy/factory.ts
```diff
@@ -0,0 +1,139 @@
+import { HardhatRuntimeEnvironment } from "hardhat/types";
+import { DeployFunction } from "hardhat-deploy/types";
+import dayjs from "dayjs";
+import duration from "dayjs/plugin/duration";
+import { parseEther } from "viem";
+dayjs.extend(duration);
+
+const ADMIN = "0x25a6592F1eaA22cD7BFc590A90d56FbB765b1Edc";
+const GOVERNANCE = "0x25a6592F1eaA22cD7BFc590A90d56FbB765b1Edc";
+const LIDO_ADDRESS = "";
```
jalextowle, line 10: Why is this empty?

deploy/steth.ts
```diff
@@ -0,0 +1,139 @@
+import { HardhatRuntimeEnvironment } from "hardhat/types";
+import { DeployFunction } from "hardhat-deploy/types";
+import dayjs from "dayjs";
+import duration from "dayjs/plugin/duration";
+import { parseEther } from "viem";
+dayjs.extend(duration);
+
+const ADMIN = "0x25a6592F1eaA22cD7BFc590A90d56FbB765b1Edc";
+const GOVERNANCE = "0x25a6592F1eaA22cD7BFc590A90d56FbB765b1Edc";
+const LIDO_ADDRESS = "";
```
jalextowle, line 10: Same here.

deploy/steth.ts
```diff
@@ -0,0 +1,139 @@
+import { HardhatRuntimeEnvironment } from "hardhat/types";
```
jalextowle, line 1: As you mentioned offline, looks like this is copy-pasta. I'll review this once it's updated.

foundry.toml
```diff
@@ -1,50 +1,55 @@
 [profile.default]
```
jalextowle, line 1: Why was this reformatted?

hardhat.config.ts
```diff
@@ -0,0 +1,182 @@
...
+          governanceLP: "0.15",
+          governanceZombie: "0.03",
+        },
+      },
+      // coordinators: {
+      //   reth: "0x....",
+      //   lido: "0x...."
+      // },
+      instances: {
+        erc4626: [
```
jalextowle, line 139: Why don't we have this for stETH.

.prettierrc
mcclurejt: had moved config to `package.json` for prettier for reasons that idt are relevant anymore, will restore this

package.json
```diff
@@ -1,40 +1,77 @@
...
-        "lint-staged": "^13.0.0",
-        "patch-package": "^6.4.7",
-        "prettier": "^2.6.2",
-        "prettier-plugin-solidity": "^1.0.0-beta.19",
-        "solhint": "^4.5.2",
-        "solhint-plugin-prettier": "0.0.5"
-    },
-    "dependencies": {
-        "yarn": "^1.22.19"
+  "name": "hyperdrive",
```
jalextowle, line 2: Same here. Looks like we're reformatting a lot of files in this PR. As said earlier, I'm not a fan of two spaces, but it's not a hill I'm going to die on if there is a good reason for making the change.

script/deploy/FactoryDeployer.s.sol
```diff
@@ -0,0 +1,357 @@
...
+    address reth;
+    address steth;
+    address erc4626Coordinator;
+    address ezethCoordinator;
+    address lsethCoordinator;
+    address rethCoordinator;
+    address stethCoordinator;
+}
+
+contract Deployer is Script, FactoryDeploymentConfig {
```
jalextowle, line 80: Is this still going to be used, or should we remove this?

cSpell.json
mcclurejt: not intentional, likely from prettier issues mentioned above, will restore

Makefile
mcclurejt: it's gonna go

contracts/test/MockRocketPool.sol
mcclurejt: probably not, definitely copied from there

deployments/sepolia/.chainId
```diff
@@ -0,0 +1 @@
+11155111
```
jalextowle, line 1: If we gitignore deployments, I think this will be gitignored too

hardhat.config.ts
```diff
@@ -0,0 +1,199 @@
...
+import "dotenv/config";
+import "./tasks";
+import {
+  ERC4626InstanceDeployConfigInput,
+  FactoryDeployConfigInput,
+  StETHInstanceDeployConfigInput,
+} from "./tasks";
+import { RETHInstanceDeployConfigInput } from "./tasks/deploy/instances/reth";
+
+const TEST_FACTORY: FactoryDeployConfigInput = {
```
jalextowle, line 14: I might be missing something, but why are the constants in this file prefixed with `TEST_`. Do you anticipate this changing when we go to mainnet? IMO, the factory isn't a test contract, but `MockLido` and `MockERC4626` are.

hardhat.config.ts
```diff
@@ -0,0 +1,199 @@
...
+const { env } = process;
+const config: HardhatUserConfig = {
+  solidity: {
+    version: "0.8.20",
+    settings: {
+      optimizer: {
+        enabled: true,
+        runs: 10_000_000,
+      },
+      evmVersion: "paris",
```
jalextowle, line 146: We should probably switch to `cancun`. We should probably do this in foundry too FWIW

tasks/deploy/all.ts
```diff
@@ -0,0 +1,47 @@
+import { task, types } from "hardhat/config";
+import dayjs from "dayjs";
+import duration from "dayjs/plugin/duration";
+import { DeployFactoryParams } from "./factory";
```
jalextowle, line 4: It looks like a few of these imports aren't used in this file. I might be missing something, but it would be good to get something like eslint set up to lint the deployment script. I'm fine with doing that in a follow-up FWIW, but this made me think about it.

contracts/src/interfaces/IRenzo.sol
```diff
@@ -58,7 +58,7 @@ interface IRenzoOracle {
         uint256 _ezETHBeingBurned,
         uint256 _existingEzETHSupply,
         uint256 _currentValueInProtocol
-    ) external pure returns (uint256);
+    ) external view returns (uint256);
```
jalextowle, line 61: Why'd we change this? It should be pure unless I'm missing something: https://github.com/Renzo-Protocol/contracts-public/blob/0530cbd20590677da5fa6fd32a0a8b250a9eb028/contracts/Oracle/RenzoOracle.sol#L134

contracts/test/MockEzEthPool.sol
```diff
@@ -0,0 +1,247 @@
+// SPDX-License-Identifier: Apache-2.0
```
jalextowle, line 1: In the future, it would be better for this to be a standalone PR. I get why it isn't considering this hasn't been merged, but we could base the PR containing this on top of the deployment tools.

contracts/test/MockEzEthPool.sol
```diff
@@ -0,0 +1,247 @@
...
+        revert("lookupTokenValue: Not Implemented");
+    }
+
+    function calculateRedeemAmount(
+        uint256 _ezETHBeingBurned,
+        uint256, // _existingEzETHSupply,
+        uint256 //  _currentValueInProtocol
+    ) public view returns (uint256) {
+        return getPooledEthByShares(_ezETHBeingBurned);
+    }
```
jalextowle, line 246: It's a bit jank IMO to make the interface function `view` when the mainnet implementation is `pure`. In practice, both are going to make a `staticcall`, so we can just leave this as-is and revert the interface.

contracts/test/MockEzEthPool.sol
```diff
@@ -0,0 +1,247 @@
...
+        revert("lookupTokenValue: Not Implemented");
+    }
+
+    function calculateMintAmount(
+        uint256, //  _currentValueInProtocol,
+        uint256, // _newValueAdded,
+        uint256 // _existingEzETHSupply
+    ) external pure returns (uint256) {
+        revert("lookupTokenValue: Not Implemented");
+    }
```
jalextowle, line 238: This is used in `_convertToShares`. `_convertToShares` is used in `_withdraw`, so I think this will cause problems for us.

tasks/deploy/coordinators/ezeth.ts
```diff
@@ -0,0 +1,132 @@
...
+      },
+    ) => {
+      // Retrieve the HyperdriveFactory deployment artifact for the current network
+      let factory = await deployments.get("HyperdriveFactory");
+      let factoryAddress = factory.address as `0x${string}`;
+
+      // Set the admin address to the deployer address if one was not provided
+      if (!admin?.length) admin = (await getNamedAccounts())["deployer"];
+
+      // Deploy a mock Lido contract if no adress was provided
```
jalextowle, line 47: Nit: 
```suggestion
      // Deploy a mock Lido contract if no address was provided
```

tasks/README.md
```diff
@@ -0,0 +1,138 @@
...
+|:--------------------------------------------------------------------------------------------------------------------------------------|
+| The bare `deploy` task is unrelated to our deployment process and SHOULD NOT BE RUN. It will be disabled/overridden in the future     |
+
+```sh
+  deploy:all                 	deploys the HyperdriveFactory and all deployer coordinators
+  deploy:coordinators:all    	deploys all deployment coordinators
+  deploy:coordinators:erc4626	deploys the ERC4626 deployment coordinator
+  deploy:coordinators:reth   	deploys the RETH deployment coordinator
+  deploy:coordinators:steth  	deploys the STETH deployment coordinator
+  deploy:coordinators:ezeth   deploys the EzETH deployment coordinator
```
jalextowle, line 29: Nit:

```suggestion
    deploy:coordinators:ezeth   deploys the EzETH deployment coordinator
```

tasks/README.md
```diff
@@ -0,0 +1,138 @@
...
+  deploy:factory             	deploys the hyperdrive factory to the configured chain
+  deploy:forwarder           	deploys the ERC20ForwarderFactory to the configured chain
+  deploy:instances:all       	deploys the ERC4626 deployment coordinator
+  deploy:instances:erc4626   	deploys the ERC4626 deployment coordinator
```
jalextowle, line 33: Are these comments wrong? I was expected to see something like: `deploys an ERC4626 pool`

tasks/deploy/coordinators/ezeth.ts
```diff
@@ -0,0 +1,132 @@
...
+        ]);
+        await deployments.save("MockEzEthPool", {
+          ...mockEzEth,
+          args: [
+            parseEther("0.035"),
+            admin as `0x${string}`,
+            true,
+            parseEther("500"),
+          ],
+        });
```
jalextowle, line 64: It would be nice to be able to parameterize deployments of vaults. Can totally leave for later, but my guess is that we could just add another Zod schema and add it as an extension to the hardhat configuration.

tasks/deploy/coordinators/reth.ts
```diff
@@ -0,0 +1,136 @@
...
+        viem,
+        getNamedAccounts,
+        config: hardhatConfig,
+      },
+    ) => {
+      let artifacts = await deployments.all();
+      if (!overwrite && artifacts["HyperdriveFactory"]) {
+        console.log(`HyperdriveFactory already deployed`);
+        return;
+      }
```
jalextowle, line 51: Is this right? I'd expect us to check the `RETHHyperdriveDeployerCoordinator` artifact to make sure we don't overwrite the coordinator. FWIW, I didn't see this in the `ezETH` script, so we may want to add that there.

tasks/deploy/coordinators/shared.ts
```diff
@@ -0,0 +1,61 @@
...
+        ...core,
+        args: [],
+      } as any);
+      await run("deploy:verify", {
+        name: `${prefixValue}HyperdriveCoreDeployer`,
+      });
+
+      // Deploy the targets
+      let targets: `0x${string}`[] = [];
+      for (let i = 0; i < 5; i++) {
```
jalextowle, line 46: Just a heads up that this will get bumped to 6 in the k-inflation PRs. I was thinking about something today, and this reminds me. We should add a function to `HyperdriveDeployerCoordinator` that records how many targets need to be deployed for a full deployment. That would be helpful to be able to parameterize loops like this without magic numbers.

I can add this once we get the k-inflation PR in, but I didn't want to forget about it.

tasks/deploy/instances/steth.ts
```diff
@@ -0,0 +1,117 @@
...
+import {
+  DeployInstanceParams,
+  PoolConfig,
+  PoolDeployConfig,
+  zInstanceDeployConfig,
+} from "./schema";
+
+dayjs.extend(duration);
+
+// Set the base token to always be the ETH letant.
```
jalextowle, line 14: What does `letant` mean?

jalextowle (approved):
Really nice job on this! The flows make a lot of sense, and it seems like the tool is quite easy to use. 

I've given this a full review, but I will say that my eyes definitely started to glaze over, so I'm going to need to use these tools to feel more comfortable with this. I think it would make sense to switch to `agent0` (if that's the platform we choose to use for our protocol dev tools) for future iterations of the protocol so that our tools are a bit more consistent. I totally get the constraints that you were under for this project (`ethers-rs` and `foundry` are both jank and caused us problems + we don't have dev tools that we're married to for the future). Curious to hear your perspective on this after doing more deployments with Hardhat.

tasks/deploy/coordinators/shared.ts
mcclurejt: Thanks for the call out! Hoping the switch should be pretty easy

hardhat.config.ts
mcclurejt: We should absolutely go through these carefully before mainnet, IIRC I copied the bulk of these parameters from our prior testing deploys hence the name. 

I've set up the hardhat network to have its own `deployments.local.json` file, and was thinking we would have dedicated factory/pool configurations for it as well. No strong preferences either way.

tasks/deploy/instances/steth.ts
mcclurejt: I think a letant might be someone who gets overzealous with find-replace 🤪

tasks/deploy/coordinators/ezeth.ts
mcclurejt: Totally agree, it's pretty tough to track all the inline definitions. Having it in the config with everything else makes a lot of sense to me

tasks/deploy/coordinators/ezeth.ts
mcclurejt: 🙏

contracts/test/MockEzEthPool.sol
sentilesdal: Yeah great idea, I was trying to mock using Lido functions, I updated this and calculateMintAmount to use the actual RenzoOracle implementations.

contracts/src/interfaces/IRenzo.sol
sentilesdal: I updated to use the actual RenzoOracle implementation and changed this back.

contracts/test/MockEzEthPool.sol
sentilesdal: Yup good call, I'll pull the contracts out and land separately and then this PR can rebase to get the contracts.

PR #967, opened 2024-04-12
code cleanup

dpaiton:
# Resolved Issues
no issues

# Description

- removes unused imports
- uses lib calc rate function instead of duplicated fn in targeted_long
- moves reserves_given_rate into lib for use by targeted short
- updates docstrings, var names, & fn names to be more clear

# Review Checklists

be removed.

### Rust

- [ ] **Testing**

### DIFF
--- a/crates/hyperdrive-math/src/lib.rs
+++ b/crates/hyperdrive-math/src/lib.rs
@@ -146,6 +146,59 @@ impl State {
         }
     }
 
+    /// Calculates the pool reserve levels to achieve a target interest rate.
+    /// This calculation does not take Hyperdrive's solvency constraints or exposure
+    /// into account and shouldn't be used directly.
+    ///
+    /// The price for a given fixed-rate is given by $p = 1 / (r \cdot t + 1)$, where
+    /// $r$ is the fixed-rate and $t$ is the annualized position duration. The
+    /// price for a given pool reserves is given by $p = \frac{\mu z}{y}^t_{s}$,
+    /// where $\mu$ is the initial share price and $t_{s}$ is the time stretch
+    /// constant. By setting these equal we can solve for the pool reserve levels
+    /// as a function of a target rate.
+    ///
+    /// For some target rate, $r_t$, the pool share reserves, $z_t$, must be:
+    ///
+    /// $$
+    /// z_t = \frac{1}{\mu} \left(
+    ///   \frac{k}{\frac{c}{\mu} + \left(
+    ///     (r_t \cdot t + 1)^{\frac{1}{t_{s}}}
+    ///   \right)^{1 - t_{s}}}
+    /// \right)^{\tfrac{1}{1 - t_{s}}}
+    /// $$
+    ///
+    /// and the pool bond reserves, $y_t$, must be:
+    ///
+    /// $$
+    /// y_t = \left(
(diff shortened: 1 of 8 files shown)


crates/hyperdrive-math/src/lib.rs
```diff
@@ -144,6 +144,59 @@ impl State {
...
+    /// as a function of a target rate.
+    ///
+    /// For some target rate, $r_t$, the pool share reserves, $z_t$, must be:
+    ///
+    /// $$
+    /// z_t = \frac{1}{\mu} \left(
+    ///   \frac{k}{\frac{c}{\mu} + \left(
+    ///     (r_t \cdot t + 1)^{\frac{1}{t_{s}}}
+    ///   \right)^{1 - t_{s}}}
+    /// \right)^{\tfrac{1}{1 - t_{s}}}
```
ryangoree, line 165:
```suggestion
    /// \right)^{\frac{1}{1 - t_{s}}}
```

crates/hyperdrive-math/src/long/targeted.rs
```diff
@@ -122,6 +125,8 @@ impl State {
                     "We overshot the zero-crossing during Newton's method.",
                 ));
             }
+            // The optimization loss can be the difference without abs or squaring
+            // because of the above check.
```
ryangoree, line 129: This seems like comment overkill since the line it's explaining immediately follows the referenced check. It took me longer to make sense of what the comment was trying to clarify than just reading the code without it.

ryangoree (approved):

crates/hyperdrive-math/src/lib.rs
```diff
@@ -144,6 +144,59 @@ impl State {
...
+    /// $$
+    /// y_t = \left(
+    ///   \frac{k}{ \frac{c}{\mu} +  \left(
+    ///     \left( r_t \cdot t + 1 \right)^{\frac{1}{t_{s}}}
+    ///   \right)^{1 - t_{s}}}
+    /// \right)^{1 - t_{s}} \left( r_t t + 1 \right)^{\frac{1}{t_{s}}}
+    /// $$
+    fn reserves_given_rate_ignoring_exposure<F: Into<FixedPoint>>(
+        &self,
+        target_rate: F,
```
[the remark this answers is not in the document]
dpaiton: I agree; seems like the tradeoff is whether or not the _caller_ should have to write `into` or if the fn should do it. The problem is if one arg is UINT and another arg is FixedPoint then we need to support `F1`, `F2`, etc. Otherwise all have to be the same type (`F`). I think this is part of the overhead you're talking about?

I would rather keep all fns `FixedPoint` and force the caller to include `into`.

This is a pervasive change to have it be uniform across the repo. If you agree then I'd rec we do it in a followup PR.

crates/hyperdrive-math/src/lib.rs
dpaiton: fwiw this is strictly a style preference; I generally like to use `tfrac` everywhere. I make exceptions for small fractions (usually `c/µ`) because it makes the fraction bar vertical instead of slanted.

ryangoree: Ah! I didn't realize tfrac was valid. I thought it threw an error in overleaf.

PR #986, opened 2024-04-17
coverage for steth and 4626

jrhea:
# Description

Add tests for these coredeployers

- [ ] **Testing**


contracts/src/deployers/steth/StETHHyperdriveCoreDeployer.sol
```diff
@@ -31,14 +31,13 @@ contract StETHHyperdriveCoreDeployer is IHyperdriveCoreDeployer {
...
     ) external returns (address) {
-        return (
-            address(
-                // NOTE: We hash the sender with the salt to prevent the
-                // front-running of deployments.
-                new StETHHyperdrive{
-                    salt: keccak256(abi.encode(msg.sender, _salt))
-                }(_config, target0, target1, target2, target3, target4)
-            )
+        address deployAddress = address(
```
jalextowle, line 34: Nit: I think `hyperdrive `would be a more intuitive name.

```suggestion
        address hyperdrive = address(
```

test/integrations/deployers/StethDeployerCoordinator.t.sol
```diff
@@ -0,0 +1,103 @@
...
+import { HyperdriveFactory } from "contracts/src/factory/HyperdriveFactory.sol";
+import { AssetId } from "contracts/src/libraries/AssetId.sol";
+import { FixedPointMath, ONE } from "contracts/src/libraries/FixedPointMath.sol";
+import { ERC20Mintable } from "contracts/test/ERC20Mintable.sol";
+import { MockLido } from "contracts/test/MockLido.sol";
+import { HyperdriveTest } from "test/utils/HyperdriveTest.sol";
+import { DeployerCoordinatorTest, MockHyperdriveDeployerCoordinator } from "test/integrations/deployers/DeployerCoordinator.t.sol";
+import { Lib } from "test/utils/Lib.sol";
+
+contract StethDeployerCoordinatorTest is DeployerCoordinatorTest {
```
jalextowle, line 23: Shouldn't we have a corresponding test for `RETH`?

test/integrations/deployers/ERC4626DeployerCoordinator.t.sol
```diff
@@ -0,0 +1,206 @@
...
+        assertEq(
+            lpShares,
+            contribution.divDown(
+                hyperdrive.getPoolConfig().initialVaultSharePrice
+            ) - 2 * hyperdrive.getPoolConfig().minimumShareReserves
+        );
+        assertEq(lpShares, hyperdrive.balanceOf(AssetId._LP_ASSET_ID, alice));
+    }
+
+    function test_initialize_success_asShares() external {
```
jalextowle, line 159: It would really be better if we did this in the contract we're inheriting from so that we get this test coverage across all of the deployment providers. I get that you probably don't want to spend the time doing that, but we might as well add a TODO so that we can keep track of the technical debt.

jalextowle (approved):

test/integrations/deployers/ERC4626DeployerCoordinator.t.sol
jrhea: the code doesn't work for steth otherwise I would have

jrhea: actually I was still working on getting this test to work with steth 😅

test/integrations/deployers/StethDeployerCoordinator.t.sol
jrhea: sure. I tried ezeth and it did nothing for additional coverage so I deleted it

contracts/src/deployers/steth/StETHHyperdriveCoreDeployer.sol
jrhea: this just needs to be reverted

Issue #973, opened 2024-04-16
`hyperdrive` metadata functions


jalextowle:
Add a `name` and `version` function to each instance, each deployer coordinator, and the hyperdrive factory. We should also add an interface for these metadata functions.

slundqui:
Having the block that the contract was deployed would be great

PR #989, opened 2024-04-18
Metadata Functions

jalextowle:
# Resolved Issues

Fixes: #973.

# Description

Adds `name` and `version` functions to each of the major contract types. 

# Review Checklists

be removed.

### Solidity

- [x] **Tokens**
              token spend?
- [ ] **Reentrancy**
- [ ] **Gas Optimizations**
          memory variables?
          issues?
- [ ] **Visibility**
- [ ] **Math**
          catch underflows?
- [ ] **Testing**

### Solidity

- [ ] **Tokens**
              token spend?
- [ ] **Reentrancy**
- [ ] **Gas Optimizations**
          memory variables?
          issues?
- [ ] **Visibility**
- [ ] **Math**
          catch underflows?
- [ ] **Testing**

### DIFF
--- a/contracts/src/deployers/erc4626/ERC4626HyperdriveDeployerCoordinator.sol
+++ b/contracts/src/deployers/erc4626/ERC4626HyperdriveDeployerCoordinator.sol
@@ -22,9 +22,6 @@ contract ERC4626HyperdriveDeployerCoordinator is HyperdriveDeployerCoordinator {
     string public constant override name =
         "ERC4626HyperdriveDeployerCoordinator";
 
-    /// @notice The deployer coordinator's version.
-    string public constant override version = "v1.0.0";
-
     /// @notice Instantiates the deployer coordinator.
     /// @param _factory The factory that this deployer will be registered with.
     /// @param _coreDeployer The core deployer.
--- a/contracts/src/deployers/erc4626/ERC4626HyperdriveDeployerCoordinator.sol
+++ b/contracts/src/deployers/erc4626/ERC4626HyperdriveDeployerCoordinator.sol
@@ -18,6 +18,13 @@ import { HyperdriveDeployerCoordinator } from "../HyperdriveDeployerCoordinator.
 contract ERC4626HyperdriveDeployerCoordinator is HyperdriveDeployerCoordinator {
     using SafeERC20 for ERC20;
 
+    /// @notice The deployer coordinator's name.
+    string public constant override name =
+        "ERC4626HyperdriveDeployerCoordinator";
+
+    /// @notice The deployer coordinator's version.
+    string public constant override version = "v1.0.0";
+
     /// @notice Instantiates the deployer coordinator.
     /// @param _factory The factory that this deployer will be registered with.
     /// @param _coreDeployer The core deployer.
--- a/contracts/src/factory/HyperdriveRegistry.sol
+++ b/contracts/src/factory/HyperdriveRegistry.sol
@@ -8,12 +8,24 @@ contract HyperdriveRegistry is
     IHyperdriveRegistry,
     IHyperdriveGovernedRegistry
 {
+    /// @notice The registry's name.
+    string public name;
+
+    /// @notice The registry's version.
+    string public constant version = "v1.0.0";
+
+    /// @notice The registry's governance address.
     address public governance;
 
+    /// @notice A mapping from hyperdrive instances to info associated with
+    ///         those instances.
     mapping(address hyperdrive => uint256 data) internal _hyperdriveInfo;
 
-    constructor() {
+    /// @notice Instantiates the hyperdrive registry.
+    /// @param _name The registry's name.
+    constructor(string memory _name) {
         governance = msg.sender;
+        name = _name;
     }
 
     modifier onlyGovernance() {
--- a/contracts/src/factory/HyperdriveRegistry.sol
+++ b/contracts/src/factory/HyperdriveRegistry.sol
@@ -3,6 +3,7 @@ pragma solidity 0.8.20;
 
 import { IHyperdriveGovernedRegistry } from "../interfaces/IHyperdriveGovernedRegistry.sol";
 import { IHyperdriveRegistry } from "../interfaces/IHyperdriveRegistry.sol";
+import { VERSION } from "../libraries/Constants.sol";
 
 contract HyperdriveRegistry is
     IHyperdriveRegistry,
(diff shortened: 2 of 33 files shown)


test/utils/Lib.sol
```diff
@@ -218,4 +218,9 @@ library Lib {
         return
             keccak256(abi.encodePacked(b1)) != keccak256(abi.encodePacked(b2));
     }
+
+    function eq(string memory b1, string memory b2) public pure returns (bool) {
+        return
+            keccak256(abi.encodePacked(b1)) == keccak256(abi.encodePacked(b2));
```
mcclurejt, line 224:
```suggestion
        if(bytes(b1).length != bytes(b2).length) {
            return false;
        } else {
            return keccak256(b1) == keccak256(b2);
        }
```

Kinda nitty but [adding the length check comes with some gas savings at little to no downside](https://fravoll.github.io/solidity-patterns/string_equality_comparison.html)

mcclurejt (approved):
threw in a suggestion but otherwise LGTM

contracts/src/factory/HyperdriveRegistry.sol
```diff
@@ -8,12 +8,24 @@ contract HyperdriveRegistry is
...
+    string public name;
+
+    /// @notice The registry's version.
+    string public constant version = "v1.0.0";
+
+    /// @notice The registry's governance address.
     address public governance;
 
+    /// @notice A mapping from hyperdrive instances to info associated with
+    ///         those instances.
```
jrhea, line 21: nice

contracts/src/instances/ezeth/EzETHTarget0.sol
```diff
@@ -19,6 +19,12 @@ import { EzETHBase } from "./EzETHBase.sol";
 contract EzETHTarget0 is HyperdriveTarget0, EzETHBase {
     using SafeERC20 for ERC20;
 
+    /// @dev The instance's name.
+    string internal constant NAME = "EzETHHyperdrive";
+
+    /// @dev The instance's version.
+    string internal constant VERSION = "v1.0.0";
```
jrhea, line 26: these version strings are going to be tedious update depending on what we consider a version rev. does this rev when the internal hyperdrive code revs?

jrhea (commented):
In general, questions are:

1) do we need deployer names? h/t @jalextowle 
2) is there an easier way to maintain version strings and test updates? Future integrations will need to strictly add these tests so that coverage doesn't drop - plus these tests are actually useful in catching stuff.
3) How are versions going to be rev'd? if the core code is updated do all versions need to be updated? if so, is there an easy way to ensure this is done and that it is easy to check?

contracts/src/factory/HyperdriveRegistry.sol
```diff
@@ -8,12 +8,24 @@ contract HyperdriveRegistry is
     IHyperdriveRegistry,
     IHyperdriveGovernedRegistry
 {
+    /// @notice The registry's name.
+    string public name;
+
+    /// @notice The registry's version.
+    string public constant version = "v1.0.0";
```
jrhea, line 15: does this need to use the constant?

contracts/src/instances/erc4626/ERC4626Target0.sol
```diff
@@ -14,9 +14,18 @@ import { ERC4626Base } from "./ERC4626Base.sol";
 ///                    only, and is not intended to, and does not, have any
 ///                    particular legal or regulatory significance.
 contract ERC4626Target0 is HyperdriveTarget0, ERC4626Base {
+    /// @dev The instance's name.
+    string internal constant NAME = "ERC4626Hyperdrive";
```
jrhea, line 18: it's capital here and lowercase in other places

contracts/src/deployers/erc4626/ERC4626HyperdriveDeployerCoordinator.sol
```diff
@@ -18,6 +18,10 @@ import { HyperdriveDeployerCoordinator } from "../HyperdriveDeployerCoordinator.
 contract ERC4626HyperdriveDeployerCoordinator is HyperdriveDeployerCoordinator {
     using SafeERC20 for ERC20;
 
+    /// @notice The deployer coordinator's name.
+    string public constant override name =
```
jrhea, line 22: should this be in caps?

contracts/test/MockHyperdriveDeployer.sol
```diff
@@ -8,6 +8,10 @@ import { IHyperdriveTargetDeployer } from "contracts/src/interfaces/IHyperdriveT
 import { MockHyperdrive } from "./MockHyperdrive.sol";
 
 contract MockHyperdriveDeployer is IHyperdriveDeployerCoordinator {
+    string public constant name = "MockHyperdriveDeployer";
+
+    string public constant version = "v1.0.0";
```
jrhea, line 13: wrong version. should use constant, right?

test/units/HyperdriveRegistry.t.sol
```diff
@@ -1,17 +1,28 @@
...
-        registry = new HyperdriveRegistry();
+        registry = new HyperdriveRegistry(NAME);
+    }
+
+    function test_name() public view {
+        assert(registry.name().eq(NAME));
+    }
+
+    function test_version() public view {
+        assert(registry.version().eq("v1.0.0"));
```
jrhea, line 25: use the constant

test/utils/Lib.sol
```diff
@@ -211,11 +211,19 @@ library Lib {
...
     function neq(bytes memory b1, bytes memory b2) public pure returns (bool) {
         return
+            b1.length != b2.length ||
             keccak256(abi.encodePacked(b1)) != keccak256(abi.encodePacked(b2));
     }
+
+    function eq(string memory b1, string memory b2) public pure returns (bool) {
+        return
+            bytes(b1).length == bytes(b2).length &&
+            keccak256(abi.encodePacked(b1)) == keccak256(abi.encodePacked(b2));
```
jrhea, line 227: nice

contracts/src/deployers/erc4626/ERC4626HyperdriveDeployerCoordinator.sol
jalextowle: Since it's public, we can leave it as lower-case

contracts/test/MockHyperdriveDeployer.sol
jalextowle: Good catch!

jrhea (approved):

jrhea (approved):

Issue #969, opened 2024-04-16
update `calc_open_short` to include price backdating


dpaiton:
`calc_open_short` was recently updated to include fees, but still does not account for the price discount that is applied if the current share price is different from the checkpoint share price. This means the contract & rust outputs will be different at any time other than the checkpoint mint time.

dpaiton:
This has to be fixed before we can make progress on https://github.com/delvtech/hyperdrive/issues/865

PR #990, opened 2024-04-18
Rust cleanup; add tests; simplify fn signatures

dpaiton:
# Resolved Issues
working towards resolving https://github.com/delvtech/hyperdrive/issues/969

# Description
- whenever `calculate_open_short` was called we would first compute `state.calculate_spot_price` and then pass it in as an argument, I removed the argument and put the calculation inside the function. If one wants to calculate open short for a different spot price, they should adjust the state and then call `state.calculate_open_short(bond_amount)`. This is the same pattern we use for e.g. `calc_spot_price_after_*`.
- I renamed a bunch of variables for consistency. Namely, adding "`maybe`" in front of `Option` parameters and changing `short_amount` to `bond_amount`.
- I added public `wallet` and `hyperdrive` attributes to the agent, which will be helpful for the calc_open_short test I'm writing in a follow-up PR.
- I renamed some tests so that they say `fuzz_*` if they are doing fuzzing (this makes it more consistent with the rest of the codebase).
- I modified the max short test to use a random `open_vault_share_price` to increase fuzz coverage.
- I added a test that had a TODO to test the unhappy path of opening a short that is  greater than the max.
- I moved the `short_principal_derivative` code out of `max.rs` and into `open.rs`, which  will be helpful for my later targeted short PR.
- I added tests for `short_principal` and `short_principal_derivative`
- I renamed `short_principal`[`_derivative`] to `calculate_short_principal`[`_derivative`] to be consistent with the rest of the codebase.

# Review Checklists

be removed.

### Rust

- [ ] **Testing**

### DIFF
--- a/contracts/src/internal/HyperdriveShort.sol
+++ b/contracts/src/internal/HyperdriveShort.sol
@@ -471,8 +471,7 @@ abstract contract HyperdriveShort is IHyperdriveEvents, HyperdriveLP {
             Errors.throwInsufficientLiquidityError();
         }
 
-        // Calculate the fees charged to the user (curveFee) and the portion
-        // of those fees that are paid to governance (governanceCurveFee).
+        // Calculate the current spot price.
         uint256 curveFee;
         uint256 governanceCurveFee;
         uint256 spotPrice = HyperdriveMath.calculateSpotPrice(
--- a/crates/hyperdrive-math/src/short/max.rs
+++ b/crates/hyperdrive-math/src/short/max.rs
@@ -82,7 +82,7 @@ impl State {
             self.absolute_max_short(spot_price, checkpoint_exposure, maybe_max_iterations);
         let absolute_max_bond_amount = max_bond_amount;
         let absolute_max_deposit =
-            match self.calculate_open_short(max_bond_amount, spot_price, open_vault_share_price) {
+            match self.calculate_open_short(max_bond_amount, open_vault_share_price) {
                 Ok(d) => d,
                 Err(_) => return max_bond_amount,
             };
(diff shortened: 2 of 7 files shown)


crates/hyperdrive-math/src/short/max.rs
```diff
@@ -511,10 +500,10 @@ impl State {
     /// doesn't support negative values.
     fn solvency_after_short_derivative(
         &self,
-        short_amount: FixedPoint,
+        bond_amount: FixedPoint,
         spot_price: FixedPoint,
     ) -> Option<FixedPoint> {
-        let lhs = self.short_principal_derivative(short_amount);
+        let lhs = self.calculateshort_principal_derivative(bond_amount);
```
slundqui, line 506: typo

crates/hyperdrive-math/src/short/open.rs
```diff
@@ -120,19 +120,138 @@ impl State {
...
+    /// bonds that are shorted $x$.
+    ///
+    /// The derivative is calculated as:
+    ///
+    /// $$
+    /// P'(x) = \tfrac{1}{c} \cdot (y + x)^{-t_s} \cdot \left(
+    ///             \tfrac{\mu}{c} \cdot (k - (y + x)^{1 - t_s})
+    ///         \right)^{\tfrac{t_s}{1 - t_s}}
+    /// $$
+    pub fn calculateshort_principal_derivative(&self, bond_amount: FixedPoint) -> FixedPoint {
```
slundqui, line 137: typo

slundqui (approved):
LGTM, but may want an additional review from @ryangoree wrt the removed parameter in `calculate_open_short`.

crates/hyperdrive-math/src/short/max.rs
dpaiton: oops thanks!

contracts/src/internal/HyperdriveShort.sol
```diff
@@ -471,8 +471,7 @@ abstract contract HyperdriveShort is IHyperdriveEvents, HyperdriveLP {
             Errors.throwInsufficientLiquidityError();
         }
 
-        // Calculate the fees charged to the user (curveFee) and the portion
-        // of those fees that are paid to governance (governanceCurveFee).
+        // Calculate the current spot price.
```
jalextowle, line 474: Good catch dude. This LGTM

jalextowle (approved):

ryangoree (approved):

PR #992, opened 2024-04-18
coverage increase for deployers

jrhea:
# Description
add return variables to increase coverage

### DIFF
--- a/contracts/src/deployers/steth/StETHHyperdriveCoreDeployer.sol
+++ b/contracts/src/deployers/steth/StETHHyperdriveCoreDeployer.sol
@@ -31,13 +31,14 @@ contract StETHHyperdriveCoreDeployer is IHyperdriveCoreDeployer {
         address target4,
         bytes32 _salt
     ) external returns (address) {
-        address hyperdrive = address(
-            // NOTE: We hash the sender with the salt to prevent the
-            // front-running of deployments.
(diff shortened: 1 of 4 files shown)


contracts/src/deployers/steth/StETHHyperdriveCoreDeployer.sol
```diff
@@ -31,13 +31,14 @@ contract StETHHyperdriveCoreDeployer is IHyperdriveCoreDeployer {
         address target4,
         bytes32 _salt
     ) external returns (address) {
-        address hyperdrive = address(
-            // NOTE: We hash the sender with the salt to prevent the
-            // front-running of deployments.
-            new StETHHyperdrive{
-                salt: keccak256(abi.encode(msg.sender, _salt))
-            }(_config, target0, target1, target2, target3, target4)
+        return (
```
jalextowle, line 34: Why aren’t we using a return variable here? Why is this one different?

jalextowle (approved):
LGTM assuming the different between StETH and the other instances is intentional. I’m still curious about the difference

contracts/src/deployers/steth/StETHHyperdriveCoreDeployer.sol
jrhea: the tests didn't like it. the coverage went down

PR #993, opened 2024-04-18
misc coverage increase

jrhea:
# Description

This PR covers an invalid deployer coordinator case that is untested and a InsufficientLiquidity revert in YS Math


jalextowle (approved):
Damn, nice tests dude

Issue #978, opened 2024-04-16
Move the Rust SDK to its own repo


jalextowle:

dpaiton:
Done: https://github.com/delvtech/hyperdrive-rs

PR #1002, opened 2024-04-25
remove all rust-related things

dpaiton:
# Resolved Issues
https://github.com/delvtech/hyperdrive/issues/978

# Description
This PR removes all rust-related stuff in favor of 
https://github.com/delvtech/hyperdrive-rs


Dockerfile
```diff
@@ -1,11 +1,5 @@
 FROM ghcr.io/foundry-rs/foundry:master as builder
 
-# Install the Rust toolchain.
```
jalextowle, line 3: Hmm, this is actually going to have some consequences for how we do testnet deployments. It’s fine to remove, but we will need a new solution. I’ll work with @mcclurejt to get this figured out before the next release, but I just wanted to make a note of it.

jalextowle (approved):
LGTM

ryangoree (approved):

Issue #1004, opened 2024-04-25
Fix `absolute_max_long` in rust sdk


slundqui:
`absolute_max_long` is failing due to `target_share_reserves < effective_share_reserves` with `arithmetic operation overflow` here:
https://github.com/delvtech/hyperdrive/blob/1576e4c4368dd9c21b2df60274f223271e793d46/crates/hyperdrive-math/src/long/max.rs#L214

This is the pool state that caused this issue:
```
pool_config = {
   base_token='0x610178dA211FEF7D417bC0e6FeD39F05609AD788',
   vault_shares_token='0xB7f8BC63BbcaD18155201308C8f3540b07f84F5e',
   linker_factory='0x5FbDB2315678afecb367f032d93F642f64180aa3',
   linker_code_hash=b'\x91\xa6<\xce\xc3\xbbXQ\xa0\x88\xc4G\xd5\xd3@`4AP\xbe+q\xac`\xc1\x97\x82\x7f\xd1HIL',
   initial_vault_share_price=FixedPoint("1.0"), minimum_share_reserves=FixedPoint("0.206282109669978528"),
   minimum_transaction_amount=FixedPoint("2.4934863032025728"),
   position_duration=30801600,
   checkpoint_duration=82800,
   time_stretch=FixedPoint("0.168128366020933774"),
   governance='0xf39Fd6e51aad88F6F4ce6aB8827279cffFb92266',
   fee_collector='0xf39Fd6e51aad88F6F4ce6aB8827279cffFb92266',
   sweep_collector='0xf39Fd6e51aad88F6F4ce6aB8827279cffFb92266',
   fees={
      curve=FixedPoint("0.063774932363567072"),
      flat=FixedPoint("0.180852177854770176"),
      governance_lp=FixedPoint("0.19281944111115744"),
      governance_zombie=FixedPoint("0.052803595874591256")
   }
}

pool_info = {
   share_reserves=FixedPoint("26176.625462115783671808"),
   share_adjustment=FixedPoint("0.0"),
   zombie_base_proceeds=FixedPoint("0.0"),
   zombie_share_reserves=FixedPoint("0.0"),
   bond_reserves=FixedPoint("73185.723713861986104902"),
   lp_total_supply=FixedPoint("26176.41918000611369328"),
   vault_share_price=FixedPoint("1.000001457064193671"),
   longs_outstanding=FixedPoint("0.0"),
   long_average_maturity_time=FixedPoint("0.0"),
[...]

PR #1005, opened 2024-04-25
Various bug fixes in rust sdk

slundqui:
# Resolved Issues
This solves various rust crashes resulting from python fuzz testing.
https://github.com/delvtech/hyperdrive/issues/1004

# Description
Fixes the following issues:
- Max long guesses is below the minimum transaction amount, so `calc_open_long` throws a minimum transaction amount error. We clamp the guess to be minimum transaction amount. and do a final check at the end to ensure the max amount is greater than the minimum transaction amount.
- We catch a case in `absolute_max_long` where the `target_share_reserves < effective_share_reserves`. In this case, we throw an error in `absolute_max_long`, and catch it and return 0 in `calc_max_long`.

# Review Checklists

be removed.

### Rust

- [ ] **Testing**

### DIFF
--- a/crates/hyperdrive-math/src/long/max.rs
+++ b/crates/hyperdrive-math/src/long/max.rs
@@ -1,4 +1,5 @@
 use ethers::types::I256;
+use eyre::{eyre, Result};
 use fixed_point::FixedPoint;
 use fixed_point_macros::{fixed, int256};
 
@@ -50,7 +51,11 @@ impl State {
 
         // Calculate the maximum long that brings the spot price to 1. If the pool is
         // solvent after opening this long, then we're done.
-        let (absolute_max_base_amount, absolute_max_bond_amount) = self.absolute_max_long();
+        let (absolute_max_base_amount, absolute_max_bond_amount) = match self.absolute_max_long() {
+            Ok(result) => result,
+            Err(_) => return fixed!(0),
+        };
+
         if self
             .solvency_after_long(
                 absolute_max_base_amount,
@@ -166,7 +171,7 @@ impl State {
     /// Calculates the largest long that can be opened without buying bonds at a
     /// negative interest rate. This calculation does not take Hyperdrive's
     /// solvency constraints into account and shouldn't be used directly.
-    fn absolute_max_long(&self) -> (FixedPoint, FixedPoint) {
+    fn absolute_max_long(&self) -> Result<(FixedPoint, FixedPoint)> {
         // We are targeting the pool's max spot price of:
         //
         // p_max = (1 - flatFee) / (1 + curveFee * (1 / p_0 - 1) * (1 - flatFee))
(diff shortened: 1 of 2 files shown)


crates/hyperdrive-math/src/long/max.rs
```diff
@@ -81,6 +81,12 @@ impl State {
         // we converge to the solution.
         let mut max_base_amount =
             self.max_long_guess(absolute_max_base_amount, checkpoint_exposure);
+
+        // possible_max_base_amount might be less than minimum transaction amount.
+        // we clamp here if so
+        if max_base_amount < self.minimum_transaction_amount() {
```
jalextowle, line 87: We might want to return 0 if we get here. If the max long is less than the minimum transaction amount, then the long can’t be opened. Ofc, we can do this at the end of the function instead of short-circuiting Newton’s. Nice catch!

crates/hyperdrive-math/src/long/max.rs
```diff
@@ -118,7 +124,15 @@ impl State {
             if maybe_derivative.is_none() {
                 break;
             }
-            let possible_max_base_amount = max_base_amount + solvency / maybe_derivative.unwrap();
+            let mut possible_max_base_amount =
+                max_base_amount + solvency / maybe_derivative.unwrap();
+
+            // possible_max_base_amount might be less than minimum transaction amount.
+            // we clamp here if so
+            if possible_max_base_amount < self.minimum_transaction_amount() {
```
jalextowle, line 132: Same here

crates/hyperdrive-math/src/short/max.rs
```diff
@@ -318,7 +318,12 @@ impl State {
         let mut maybe_solvency =
             self.solvency_after_short(max_bond_amount, spot_price, checkpoint_exposure);
         if maybe_solvency.is_none() {
-            panic!("Initial guess in `max_short` is insolvent.");
+            // If initial guess is insolvent, use the absolute max short
+            // as the initial guess
+            max_bond_amount = absolute_max_bond_amount;
+            // We expect by definition for the pool to be solvent with the absolute max bond amount.
+            maybe_solvency =
+                self.solvency_after_short(max_bond_amount, spot_price, checkpoint_exposure);
```
jalextowle, line 326: Didn’t we already check that this isn’t insolvent on line 294? I might be missing something, but I don’t expect this to be solvent

slundqui: Ah, my mistake, I didn't realize this was in the `absolute_max_short` function. Is it safe to say here then that if `absolute_max_short_guess` returns a value that is insolvent, that the pool doesn't allow for a short at all? Or is there a better initial guess we can use here

crates/hyperdrive-math/src/long/max.rs
slundqui: Makes sense. I'll keep this check here just in case `max_long_guess` underestimates too much, and allow newton's method to iterate using this as an initial guess.

crates/hyperdrive-math/src/long/max.rs
slundqui: Again, I'll keep this check here due to a possibility of the `possible_max_base_amount` overshooting. This clamp is mostly for avoiding a `minimum_transaction_error` in `calculate_open_long`. I'll add a final check and return 0 if this value is too small at the end.

crates/hyperdrive-math/src/short/max.rs
jalextowle: We might be able to use a better initial guess. We could always use 0, or we could allow “cross-overs” by allowing invalid solutions so that we can proceed with iterations. The latter sounds fine to me since Newton’s should get to a solution that is solvent after a few iterations.

This check was based on my (erroneous) thinking at implementation-time that this algorithm would never cross over the solution

slundqui:
Closing in favor of https://github.com/delvtech/hyperdrive-rs/pull/5

Issue #1000, opened 2024-04-23
Fix testnet bug


jalextowle:
- [x] Fix k inflation in the code
- [x] Fix the tests
- [x] Write up k inflation
- [x] Analyze the circuit breaker feature to choose an appropriate circuit breaker for different time stretches.
- [x] Deploy the fix to testnet
- [x] Pause the existing pools
- [x] @CPSTL @jrhea To communicate fix to Discord/Telegram

PR #1012, opened 2024-05-02
Price Discovery Fix Pt. 1

jalextowle:
# Resolved Issues

Addresses a component of #1000.

# Description

https://hackmd.io/yMNxotdqRKSuNrac9w8uFw?view

# Review Checklists

be removed.

## @jrhea 

- [ ] **Tokens**
              token spend?
- [ ] **Reentrancy**
- [ ] **Gas Optimizations**
          memory variables?
          issues?
- [ ] **Visibility**
- [ ] **Math**
          catch underflows?
- [ ] **Testing**

## @mcclurejt 

- [ ] **Tokens**
              token spend?
- [ ] **Reentrancy**
- [ ] **Gas Optimizations**
          memory variables?
          issues?
- [ ] **Visibility**
- [ ] **Math**
          catch underflows?
- [ ] **Testing**

## @cashd 

- [ ] **Tokens**
              token spend?
- [ ] **Reentrancy**
- [ ] **Gas Optimizations**
          memory variables?
          issues?
- [ ] **Visibility**
- [ ] **Math**
          catch underflows?
- [ ] **Testing**


contracts/src/deployers/HyperdriveDeployerCoordinator.sol
```diff
@@ -44,6 +44,8 @@ abstract contract HyperdriveDeployerCoordinator is
         address target3;
         /// @dev The address of the HyperdriveTarget4 contract.
         address target4;
+        /// @dev The address of the HyperdriveTarget5 contract.
+        address target5;
```
mcclurejt, line 48: 😿

contracts/src/libraries/LPMath.sol
```diff
@@ -28,6 +28,103 @@ library LPMath {
...
+
+    /// @dev Calculates the initial reserves. We solve for the initial reserves
+    ///      by solving the following equations simultaneously:
+    ///
+    ///      (1) c * z = c * z_e + p_target * y
+    ///
+    ///      (2) p_target = ((mu * z_e) / y) ** t_s
+    ///
+    ///      where p_target is the target spot price implied by the target spot
+    ///      rate.
```
mcclurejt, line 66: It looks to me like the formula for (2) is different than [what is stated in the hackmd](https://hackmd.io/yMNxotdqRKSuNrac9w8uFw?view#Zeta-and-Bond-Reserves-for-Initialization).

Is this due to a simplification?

contracts/src/libraries/LPMath.sol
```diff
@@ -28,6 +28,103 @@ library LPMath {
...
+        shareReserves = _shareAmount;
+
+        // NOTE: Round down to underestimate the initial bond reserves.
+        //
+        // Calculate the initial bond reserves. This is given by:
+        //
+        // y = (mu * c * z) / (c * p_target ** (1 / t_s) + mu * p_target)
+        bondReserves = _initialVaultSharePrice.mulDivDown(
+            _vaultSharePrice.mulDown(shareReserves),
+            _vaultSharePrice.mulUp(targetPrice.pow(ONE.divDown(_timeStretch))) +
```
mcclurejt, line 114: I think that rounding up for the `1/t_s` would result in a larger denominator, which would further underestimate the bond reserves.

```suggestion
            _vaultSharePrice.mulUp(targetPrice.pow(ONE.divUp(_timeStretch))) +
```

test/units/hyperdrive/UpdateLiquidityTest.t.sol
```diff
@@ -4,6 +4,7 @@ pragma solidity 0.8.20;
 import { IERC20 } from "contracts/src/interfaces/IERC20.sol";
 import { IHyperdrive } from "contracts/src/interfaces/IHyperdrive.sol";
 import { FixedPointMath } from "contracts/src/libraries/FixedPointMath.sol";
+import { HyperdriveMath } from "contracts/src/libraries/HyperdriveMath.sol";
```
mcclurejt, line 7:
```suggestion
```

test/units/libraries/HyperdriveMath.t.sol
```diff
@@ -4,7 +4,8 @@ pragma solidity 0.8.20;
 import { IHyperdrive } from "contracts/src/interfaces/IHyperdrive.sol";
 import { FixedPointMath, ONE } from "contracts/src/libraries/FixedPointMath.sol";
 import { HyperdriveMath } from "contracts/src/libraries/HyperdriveMath.sol";
-import { YieldSpaceMath } from "contracts/src/libraries/HyperdriveMath.sol";
+import { LPMath } from "contracts/src/libraries/LPMath.sol";
+import { YieldSpaceMath } from "contracts/src/libraries/YieldSpaceMath.sol";
```
mcclurejt, line 8:
```suggestion
```

test/units/libraries/LPMath.t.sol
```diff
@@ -14,6 +14,159 @@ contract LPMathTest is HyperdriveTest {
...
+        // NOTE: Coverage only works if I initialize the fixture in the test function
+        MockLPMath lpMath = new MockLPMath();
+
+        // Test .1% APR
+        uint256 shareReserves = 500_000_000e18;
+        uint256 initialVaultSharePrice = 0.5e18;
+        uint256 vaultSharePrice = 2.5e18;
+        uint256 apr = 0.001e18;
+        uint256 positionDuration = 365 days;
+        uint256 timeStretch = ONE.divDown(1109.3438508425959e18);
```
mcclurejt, line 29: I'm not in love with the magic number timestretch values used in this test.

Would be nice to have a clarifying comment around why these numbers in particular were chosen.

Ideally we'd have some formulaic generation for it, but I realize that may be infeasible or unreasonably time consuming.

mcclurejt (commented):
Really awesome job.

Left some minor nits and questions.

contracts/src/libraries/LPMath.sol
jalextowle: The effective share reserves, $z_e$, is defined as $z_e = z - \zeta$, where $z$ denotes the shares reserves and $\zeta$ denotes the share adjustment. This is typically how I write the price, and it's just a matter of personal preference IMO (I didn't write the HackMD).

contracts/src/libraries/LPMath.sol
jalextowle: In this case, it's actually the opposite. `targetPrice` is less than one (we know this because the target rate is a `uint256` and can thus never be less than 0). With this in mind, exponentiating `targetPrice` by a smaller exponent will result in a larger value. 

Think of this example. We have that $\frac{1}{2} = \sqrt{\frac{1}{4}}$ since $\frac{1}{2} \cdot \frac{1}{2} = \frac{1}{4}$. We also have that $\frac{1}{2} = \frac{1}{4}^{\frac{1}{2}}  > \frac{1}{4}^{1} = \frac{1}{4}$.

contracts/src/external/HyperdriveTarget0.sol
```diff
@@ -323,11 +323,15 @@ abstract contract HyperdriveTarget0 is
...
-            : 0;
+        uint256 presentValue;
+        if (vaultSharePrice > 0) {
+            (presentValue, ) = LPMath.calculatePresentValueSafe(
+                _getPresentValueParams(vaultSharePrice)
+            );
+            presentValue = presentValue.mulDown(vaultSharePrice);
+        } else {
+            presentValue = 0;
+        }
```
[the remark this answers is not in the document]
jalextowle: Good call! We don't even need the `uint256 presentValue = 0` since an uninitialized value is zero. This was a brain fart.

test/units/libraries/LPMath.t.sol
jalextowle: Agreed. This test was moved from `HyperdriveMath`, but I can update this.

contracts/src/libraries/LPMath.sol
mcclurejt: Really appreciate the explanation, makes total sense.

jrhea (approved):

PR #1019, opened 2024-05-09
static type checking for deployment config + full verifications

mcclurejt:
# Resolved Issues

- Deployment configuration files are now statically type-checked instead of being verified at runtime
- Deployment configuration types are derived from the contract abis, so breaking contract changes are reflected immediately.
- Verification works for all contract types.

# Description

# Review Checklists

be removed.


tasks/README.md
```diff
@@ -20,20 +20,20 @@ The complete list of tasks can be seen by running `npx hardhat --help` in your t
...
-  deploy:registry       deploys the hyperdrive factory to the configured chain
-  etherscan-verify      submit contract source code to etherscan
+  deploy:hyperdrive     deploys the HyperdriveFactory, all deployer coordinators, and all hyperdrive instances
+  deploy:verify         attempts to verify all deployed contracts for the specified network
 ```
 
-## Configuration
+The `deploy:hyperdrive` task should be run for all deployments. It handles resuming existing deploys
+and only deploys configurations that are not already present on the specified chain. It must be
+provided a `--network` flag.
```
jalextowle, line 29: I think it would be good to add an examples section that shows the exact command used to deploy and verify. That'll help people when they are getting started so they don't need to figure out the cli args

tasks/deploy/verify.ts
```diff
@@ -0,0 +1,195 @@
...
+
+            // verify the factory
+            console.log(`verifying ${f.name}...`);
+            await run("verify:verify", {
+                address: hyperdriveDeploy.deployments.byName(f.name).address,
+                constructorArguments,
+            });
+        }
+
+        // loop through all instances
```
jalextowle, line 54:
```suggestion
        // loop through all deployer coordinators
```

tasks/deploy/verify.ts
```diff
@@ -0,0 +1,195 @@
...
+                constructorArguments: c.extraConstructorArgs
+                    ? await evaluateValueOrHREFn(
+                          c.extraConstructorArgs,
+                          hre,
+                          {},
+                      )
+                    : [],
+            });
+
+            // verify the targets
```
jalextowle, line 76:
```suggestion
            // verify the target deployers
```

tasks/deploy/config/sepolia/dai-14day.ts
```diff
@@ -1,121 +1,119 @@
...
-    timestretchAPR: "0.05",
+    fixedAPR: parseEther("0.10"),
+    timestretchAPR: parseEther("0.10"),
+    targetCount: 4,
     options: {
-        // destination: "0xsomeone", defaults to deployer
+        extraData: "0x",
         asBase: true,
-        // extraData: "0x",
+        destination: "0xd94a3A0BfC798b98a700a785D5C610E8a2d5DBD8",
```
jalextowle, line 27: It's fine to leave this as is for now, but it would be good for this to be the `msg.sender` in the future

tasks/deploy/config/sepolia/dai-14day.ts
```diff
@@ -1,121 +1,119 @@
...
+        hre.hyperdriveDeploy.deployments.byName("ERC4626_COORDINATOR").address,
+    deploymentId: toBytes32("DAI_14_DAY"),
+    salt: toBytes32("0x69420"),
+    extraData: "0x",
     contribution: CONTRIBUTION,
-    fixedAPR: "0.05",
-    timestretchAPR: "0.05",
+    fixedAPR: parseEther("0.10"),
+    timestretchAPR: parseEther("0.10"),
+    targetCount: 4,
```
jalextowle, line 23: We have a function that we can query on the deployer coordinator to get the number of targets. Using that would make this a bit more generic.

tasks/deploy/config/sepolia/dai-14day.ts
```diff
@@ -1,121 +1,119 @@
...
+                hre.hyperdriveDeploy.deployments.byName("SDAI").address,
+            circuitBreakerDelta: parseEther("0.6"),
+            minimumShareReserves: parseEther("10"),
+            minimumTransactionAmount: parseEther("0.001"),
+            positionDuration: parseDuration("14 days"),
+            checkpointDuration: parseDuration("1 day"),
+            timeStretch: 0n,
+            governance: "0xc187a246Ee5A4Fe4395a8f6C0f9F2AA3A5a06e9b",
+            feeCollector: "0xc187a246Ee5A4Fe4395a8f6C0f9F2AA3A5a06e9b",
+            sweepCollector: "0xc187a246Ee5A4Fe4395a8f6C0f9F2AA3A5a06e9b",
```
jalextowle, line 106: Would be good to read these from the factory. The deployment will fail if the details are different from the factory details.

tasks/deploy/config/sepolia/dai-14day.ts
jalextowle: The comments in this file can be repeated for the other scripts.

tasks/deploy/config/sepolia/erc4626-coordinator.ts
```diff
@@ -1,40 +1,11 @@
...
-                let tx = await factory.write.addDeployerCoordinator([
-                    coordinator.address,
-                ]);
-                await pc.waitForTransactionReceipt({ hash: tx });
-            }
-        },
+        extraConstructorArgs: [],
+        factoryAddress: async (hre) =>
+            hre.hyperdriveDeploy.deployments.byName("FACTORY").address,
     };
```
jalextowle, line 11: Nice, this is much cleaner!

tasks/deploy/config/sepolia/ezeth-14day.ts
```diff
@@ -1,59 +1,71 @@
...
+            hre.hyperdriveDeploy.deployments.byName("EZETH").address,
+        );
+        let pc = await hre.viem.getPublicClient();
+        // mint the contribution
+        let tx = await vaultSharesToken.write.mint([CONTRIBUTION]);
+        // approve the coordinator
+        tx = await vaultSharesToken.write.approve([
+            hre.hyperdriveDeploy.deployments.byName("EZETH_COORDINATOR")
+                .address,
+            CONTRIBUTION + parseEther("10"),
```
jalextowle, line 41: Why are we approving with this amount?

tasks/deploy/lib/environment-extensions.ts
```diff
@@ -28,75 +28,72 @@ export type HyperdriveDeployRuntimeOptions = {
...
-        approveWithFactory?: boolean;
-    };
-
-export const HyperdriveDeployCoordinatorTask = (
-    task: ConfigurableTaskDefinition,
-) =>
-    HyperdriveDeployBaseTask(task).addOptionalParam(
-        "approveWithFactory",
-        "attempt approving the HyperdriveDeployerCoordinator with its factory (will fail if deployer address is not governance)",
+    task.addOptionalParam(
```
jalextowle, line 48: When do we not want to save? This is a genuine question, I'm just curious

jalextowle (approved):
LGTM except for the comments

tasks/deploy/config/sepolia/dai-14day.ts
mcclurejt: Totally agree with you.

Unfortunately doing this is be pretty non-trivial since we get the `PoolDeployConfig` type directly from the abi, so we'd have to start dissecting it a good bit.

Nice middle ground might be auto-filling them when we do config generation from the env file... will keep it in mind

tasks/deploy/lib/environment-extensions.ts
mcclurejt: Original idea was for local or hardhat network deploys so the `deployments.local.json` file wouldn't have to be deleted between subsequent runs.... however the deployment process is so dependent on checking previous deployment entries that this strategy wouldn't work. Will remove

tasks/deploy/config/sepolia/ezeth-14day.ts
mcclurejt: For some reason (unsure why) we run into issues with insufficient approval amount on this contract. Rounding error maybe? IIRC you and I both encountered it when we ran the deploy script over vc last week.

Will reduce it to `0.01` ether since that should be sufficient

tasks/deploy/config/sepolia/dai-14day.ts
mcclurejt: Agreed, in the same vein as [this](https://github.com/delvtech/hyperdrive/pull/1019#discussion_r1604144477) but easier, can tackle at a later point in time

PR #1024, opened 2024-05-13
Updated the event schema

jalextowle:
# Description

Updated the event schema so that it is more feasible to recreate the state using events.

# Review Checklists

be removed.

## [[Reviewer Name]]

- [ ] **Tokens**
              token spend?
- [ ] **Reentrancy**
- [ ] **Gas Optimizations**
          memory variables?
          issues?
- [ ] **Visibility**
- [ ] **Math**
          catch underflows?
- [ ] **Testing**

### DIFF
--- a/contracts/src/internal/HyperdriveLP.sol
+++ b/contracts/src/internal/HyperdriveLP.sol
@@ -126,16 +126,11 @@ abstract contract HyperdriveLP is
         );
 
         // Emit an Initialize event.
-        uint256 baseContribution = _convertToBaseFromOption(
-            _contribution,
-            vaultSharePrice,
-            _options
-        );
         emit Initialize(
             _options.destination,
             lpShares,
-            baseContribution, // base contribution
-            shareContribution, // vault shares contribution
+            _contribution,
+            vaultSharePrice,
             _options.asBase,
             _apr
         );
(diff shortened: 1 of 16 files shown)


contracts/src/internal/HyperdriveLP.sol
```diff
@@ -383,12 +373,8 @@ abstract contract HyperdriveLP is
             msg.sender, // provider
             _options.destination, // destination
             _lpShares,
-            _convertToBaseFromOption(proceeds, vaultSharePrice, _options), // base proceeds
-            _convertToVaultSharesFromOption(
```
jrhea, line 387: Can you get rid of these helper methods now?

jrhea (commented):
This seems like a good simplification and should reduce gas costs some. A few comments:

1) did the coverage go down bc the loc was reduced?
2) is the frontend team aware that they will have to look at the asBase option to determine the units?
3) can we get rid of the convertToBase/SharesFromOption helper methods?

contracts/src/internal/HyperdriveLP.sol
jalextowle: Yep, I did that

test/units/hyperdrive/FeeTest.t.sol
```diff
@@ -1,7 +1,11 @@
 // SPDX-License-Identifier: Apache-2.0
 pragma solidity 0.8.20;
 
+// FIXME
+import { console2 as console } from "forge-std/console2.sol";
```
jrhea, line 5: fix me

jrhea (approved):
looks good!

PR #1028, opened 2024-05-15
update gov fee in utils to match base

dpaiton:
# Description

The gov fee for shorts in `HyperdriveUtils.sol` did not match the behavior in `HyperdriveBase.sol`:

`HyperdriveUtils.sol`:
```sol
    /// @dev Gets the governance fee paid by the trader when they open a short.
    /// @param _bondAmount The bond amount.
    /// @param _spotPrice The spot price.
    /// @param _curveFee The curve fee parameter.
    /// @param _governanceLPFee The governance fee parameter.
    /// @return The governance fee.
    function calculateShortGovernanceFee(
        uint256 _bondAmount,
        uint256 _spotPrice,
        uint256 _curveFee,
        uint256 _governanceLPFee
    ) internal pure returns (uint256) {
        return
            calculateShortCurveFee(_bondAmount, _spotPrice, _curveFee).mulUp(
                _governanceLPFee
            );
    }
```

It is mulDown in `HyperdriveBase.sol::_calculateFeesGivenBonds`:
```sol
        // NOTE: Round down to underestimate the governance curve fee.
        //
        // Calculate the curve portion of the governance fee:
        //
        // governanceCurveFee = curve_fee * phi_gov
        //                    = shares * phi_gov
        governanceCurveFee = curveFee.mulDown(_governanceLPFee);
```

# Review Checklists

be removed.

## [[Reviewer Name]]

- [ ] **Tokens**
              token spend?
- [ ] **Reentrancy**
- [ ] **Gas Optimizations**
          memory variables?
          issues?
- [ ] **Visibility**
- [ ] **Math**
          catch underflows?
- [ ] **Testing**


test/utils/HyperdriveUtils.sol
```diff
@@ -755,8 +755,8 @@ library HyperdriveUtils {
 
         // Finish computing the derivative.
         derivative += _params.governanceLPFee.mulDown(_params.curveFee).mulDown(
-            ONE - _spotPrice
-        );
+                ONE - _spotPrice
```
jalextowle, line 758: You might need to run prettier

jalextowle (approved):

PR #1034, opened 2024-05-21
testnet/devnet image using hardhat deploy

mcclurejt:
# Description
- Uses hardhat-based hyperdrive deploy for devnet/testnet docker image
- Fixes some bugs with deploy diff checking and running verify for non-live networks
- Removes some vestigial rust code references


.dockerignore
```diff
@@ -0,0 +1,11 @@
...
+cache/
+deployments/
+forge-cache/
+node_modules/
+out/
+target/
+*.env
+deployments.json
+deployments.*.json
+
```
jalextowle, line 11: Nit: Remove the newline

```suggestion
```

Dockerfile
```diff
@@ -1,92 +1,107 @@
...
+# Copy only package.json and yarn.lock to nearly guarantee using a cached layer for node_modules.
+RUN apk add --no-cache npm && \
+  npm install -g yarn
+WORKDIR /app
+COPY ./package.json ./package.json
+COPY ./yarn.lock ./yarn.lock
+RUN yarn install --immutable 
 
+FROM ghcr.io/foundry-rs/foundry@sha256:4606590c8f3cef6a8cba4bdf30226cedcdbd9f1b891e2bde17b7cf66c363b2b3 as builder
+# Install node related depencies. The `npm` apk package includes node 18 as a dependency.
```
jalextowle, line 14: Nit: I'd move this comment about the `FROM` and just use this to describe what the layer is doing.

Dockerfile
```diff
@@ -1,92 +1,107 @@
-FROM ghcr.io/foundry-rs/foundry:master as builder
+# NOTE: The `latest` version of foundry does not include arm64 as a build platform,
```
jalextowle, line 1: Can we re-evaluate the style for this file? Try to break the code up so that each comment is preceded by a newline and followed by several lines of code. I make an exception for `NOTE` and `HACK` comments FWIW.

jalextowle (approved):
Nice job dude! LGTM once the style is updated. I'm assuming this has been tested, but if not, it would be good to do some smoke testing to make sure that it will #justwork.

Dockerfile
mcclurejt: Totally! It is somewhat inconsistent with the rest of our files.

Some context on why I changed it... empty lines within a stage in a multi-stage build make it super difficult to quickly grok where one stage begins and another ends. 

I took a peek at some of docker's documentation and it looks like they feel similarly, albeit most of what I saw was trivial with little commentary.

Personally not against whitespace within a stage, so long as we have something more visually significant to break up stages....

Dockerfiles only have single-line comments, so maybe a header for each stage like: 

```dockerfile
######################################
# Stage Name
# 
# what the stage does....
######################################
```

PR #1036, opened 2024-05-22
mainnet forking

mcclurejt:
# Description

Handle deployment of and interaction with a fork of mainnet.

See [this PR](https://github.com/delvtech/hyperdrive-infra/pull/117) to get a sense of how it's being used.

- Split up hh configs to avoid needing to define excessive and unnecessary environment variables
- Scripts to mint all tokens
- Dockerfile for fork deployment and script running


.github/workflows/devnet_docker.yml
```diff
@@ -53,12 +53,13 @@ jobs:
...
             FACTORY_MIN_CURVE_FEE=0
             FACTORY_MIN_FLAT_FEE=0
             FACTORY_MIN_GOVERNANCE_LP_FEE=0
             FACTORY_MIN_GOVERNANCE_ZOMBIE_FEE=0
-            FACTORY_MAX_CURVE_FEE=1000000000000000000
-            FACTORY_MAX_FLAT_FEE=1000000000000000000
-            FACTORY_MAX_GOVERNANCE_LP_FEE=1000000000000000000
-            FACTORY_MAX_GOVERNANCE_ZOMBIE_FEE=1000000000000000000
-            FACTORY_MIN_POSITION_DURATION=86400
+            FACTORY_MAX_CURVE_FEE=1
```
jalextowle, line 61: Why are these all 1 now?

fork.Dockerfile
```diff
@@ -0,0 +1,28 @@
+# NOTE: The `latest` version of foundry does not include arm64 as a build platform,
```
jalextowle, line 1: What's the idea behind this dockerfile? Is the idea just that it ships with compiled contracts and the deployment tools? If so, we should consider using a different name than `fork.Dockerfile` since the contents of this file don't really relate to using a forked anvil node. This said, I could be missing something.

scripts/deploy-fork.sh
```diff
@@ -0,0 +1,41 @@
...
+  dai14Day: .DAI_14_DAY.address,
+  dai30Day: .DAI_30_DAY.address,
+  steth14Day: .STETH_14_DAY.address,
+  steth30Day: .STETH_30_DAY.address,
+  reth14Day: .RETH_14_DAY.address,
+  reth30Day: .RETH_30_DAY.address,
+  factory: .FACTORY.address,
+  hyperdriveRegistry: .MAINNET_FORK_REGISTRY.address,
+  }" >./artifacts/addresses.json
+cp ./deployments.local.json ./artifacts/
```
jalextowle, line 41: These comments are excellent. Thank you 🙏

.github/workflows/devnet_docker.yml
jalextowle: NVM, I see that it's because they are parsed differently. This indirection is a bit subtle, so I think some inline comments like `# 1e18` would go a long way towards making this more readable.

tasks/fork/mint-reth.ts
```diff
@@ -0,0 +1,51 @@
...
+)
+    .addParam("address", "address to send RETH", undefined, types.string)
+    .addParam("amount", "amount (in ether) to mint", undefined, types.string)
+    .setAction(
+        async (
+            { address, amount }: Required<MintRETHParams>,
+            { viem, artifacts },
+        ) => {
+            let transferData = encodeFunctionData({
+                abi: (await artifacts.readArtifact("ERC20Mintable")).abi,
```
jalextowle, line 31: This is fine IMO, but it might be clearer to use `IERC20` since we're not using any of the special functionality introduced by `ERC20Mintable`.

tasks/fork/mint-dai.ts
```diff
@@ -0,0 +1,49 @@
...
+)
+    .addParam("address", "address to send DAI", undefined, types.string)
+    .addParam("amount", "amount (in ether) to mint", undefined, types.string)
+    .setAction(
+        async (
+            { address, amount }: Required<MintDAIParams>,
+            { viem, artifacts },
+        ) => {
+            let transferData = encodeFunctionData({
+                abi: (await artifacts.readArtifact("ERC20Mintable")).abi,
```
jalextowle, line 29: Same here

tasks/fork/maintain-rate.ts
```diff
@@ -0,0 +1,89 @@
...
+                    });
+                    increase =
+                        (currentBalance * BigInt(interval)) /
+                        (365n * 24n * rateFraction);
+                    await tc.setBalance({
+                        address: MAINNET_RETH_ADDRESS,
+                        value: currentBalance + rateFraction,
+                    });
+                    console.log(`reth balance increased by ${increase}`);
+                    break;
```
jalextowle, line 78: Some comments would make this a bit more parseable.

tasks/fork/maintain-rate.ts
```diff
@@ -0,0 +1,89 @@
...
+
```
jalextowle, line 79: Nit: Dangling newline?

tasks/fork/create-checkpoints.ts
```diff
@@ -0,0 +1,109 @@
...
+                      poolAddresses.map(
+                          async (a) =>
+                              (
+                                  await (
+                                      await viem.getContractAt(
+                                          "HyperdriveTarget0",
+                                          a,
+                                      )
+                                  ).read.getPoolConfig()
+                              ).checkpointDuration,
```
jalextowle, line 56: Why are we reading this from `Target0`? We can just use `IHyperdrive` for this. Maybe this is one of the getters that you were talking about the other day.

tasks/fork/create-checkpoints.ts
```diff
@@ -0,0 +1,109 @@
...
+        "fork:create-checkpoints",
+        "Long-running process that creates checkpoints for all hyperdrive instances when necessary",
+    ),
+)
+    .addOptionalParam(
+        "pollInterval",
+        "Amount of time in minutes between polls for checkpoint state",
+        240,
+        types.int,
+    )
```
jalextowle, line 31: I sort of like how the other checkpoint bot calculates the poll interval so that it always targets a particular time in the checkpoint (like halfway through the checkpoint) to checkpoint. It only needs to check once a checkpoint if we use this method. The method used here definitely works, but it will result in the checkpoint bots minting more of the checkpoints for lower volume pools. The downside of this is that it costs money to mint a checkpoint, and it would be preferable to wait until some amount of the checkpoint has passed before re-minting a checkpoint. 

I say all of this to add some context for why the other checkpoint bot works the way that it does. These concerns aren't relevant for the mainnet fork environment, but we'd most likely want to make these changes if we decided to use this bot script for mainnet.

tasks/deploy/config/mainnet-fork/steth-30day.ts
```diff
@@ -0,0 +1,74 @@
...
+            amount: formatEther(CONTRIBUTION),
+            address: (await hre.getNamedAccounts())["deployer"],
+        });
+        let vaultSharesToken = await hre.viem.getContractAt(
+            "MockLido",
+            MAINNET_STETH_ADDRESS,
+        );
+        let pc = await hre.viem.getPublicClient();
+        // approve the coordinator
+        // TODO: Investigate why significant additional allowance is needed for deposit (10 ether is too little)
```
jalextowle, line 41: I think that what's going on here is that steth approvals are in units of tokens rather than shares: https://github.com/lidofinance/lido-dao/blob/master/contracts/0.4.24/StETH.sol#L212, but we are trying to deposit `500.001e18 shares`. Let's say that the share price is `1.2e18 steth/share`. Then we'd need an approval amount of  `600.0012 steth`.

tasks/deploy/config/mainnet-fork/steth-14day.ts
```diff
@@ -0,0 +1,74 @@
...
+        extraData: "0x",
+    },
+    prepare: async (hre) => {
+        let vaultSharesToken = await hre.viem.getContractAt(
+            "MockLido",
+            MAINNET_STETH_ADDRESS,
+        );
+        let pc = await hre.viem.getPublicClient();
+        // approve the coordinator
+        // TODO: Investigate why significant additional allowance is needed for deposit (10 ether is too little)
```
jalextowle, line 36: Same here.

tasks/deploy/config/mainnet-fork/steth-30day.ts
jalextowle: We should be able to calculate this close to perfectly using `getPooledEthByShares`.

tasks/deploy/config/mainnet-fork/reth-30day.ts
```diff
@@ -0,0 +1,71 @@
...
+        let vaultSharesToken = await hre.viem.getContractAt(
+            "MockRocketPool",
+            MAINNET_RETH_ADDRESS,
+        );
+        let pc = await hre.viem.getPublicClient();
+        // TODO: Investigate why significant additional allowance is needed for deposit (10 ether is too little)
+        let tx = await vaultSharesToken.write.approve([
+            hre.hyperdriveDeploy.deployments.byName("RETH_COORDINATOR").address,
+            CONTRIBUTION + parseEther("100"),
+        ]);
```
jalextowle, line 43: This is more confusing to me since RETH isn't a rebasing token. Is this just legacy from the steth scripts?

tasks/deploy/config/mainnet-fork/reth-14day.ts
```diff
@@ -0,0 +1,71 @@
...
+        await hre.run("fork:mint-reth", {
+            address: (await hre.getNamedAccounts())["deployer"],
+            amount: formatEther(CONTRIBUTION),
+        });
+        let vaultSharesToken = await hre.viem.getContractAt(
+            "MockRocketPool",
+            MAINNET_RETH_ADDRESS,
+        );
+        let pc = await hre.viem.getPublicClient();
+        // TODO: Investigate why significant additional allowance is needed for deposit (10 ether is too little)
```
jalextowle, line 39: Same here

tasks/deploy/config/mainnet-fork/factory.ts
```diff
@@ -0,0 +1,60 @@
...
+        );
+    },
+    constructorArguments: async (hre) => [
+        {
+            governance: process.env.ADMIN! as `0x${string}`,
+            deployerCoordinatorManager: process.env.ADMIN! as `0x${string}`,
+            hyperdriveGovernance: "0xc187a246Ee5A4Fe4395a8f6C0f9F2AA3A5a06e9b",
+            defaultPausers: [
+                (await hre.getNamedAccounts())["deployer"] as Address,
+            ],
```
jalextowle, line 21: Nice. Can we make sure this gets updated in the sepolia scripts too if it hasn't already? We should also add governance here just in case.

jalextowle (commented):
It's looking great! I left a few comments, but just let me know when it's ready for a final review.

.github/workflows/devnet_docker.yml
mcclurejt: added, good call!

fork.Dockerfile
mcclurejt: combined with original

tasks/fork/create-checkpoints.ts
mcclurejt: Really appreciate the context here, thank you so much!

Will leave as is since this script is ideally temporary until agent0 checkpointing gets used.

tasks/deploy/config/mainnet-fork/steth-30day.ts
mcclurejt: fixed

jalextowle (approved):

PR #1039, opened 2024-05-30
Made the registry more fully-featured

jalextowle:
# Description

Updates the schema for the registry so that it is more helpful for programmatic users. This PR also updates some inconsistencies in the `HyperdriveFactory` and interfaces throughout the codebase.

# Review Checklists

be removed.

## [[Reviewer Name]]

- [ ] **Tokens**
              token spend?
- [ ] **Reentrancy**
- [ ] **Gas Optimizations**
          memory variables?
          issues?
- [ ] **Visibility**
- [ ] **Math**
          catch underflows?
- [ ] **Testing**

### DIFF
--- a/contracts/src/factory/HyperdriveFactory.sol
+++ b/contracts/src/factory/HyperdriveFactory.sol
@@ -179,11 +179,11 @@ contract HyperdriveFactory is IHyperdriveFactory {
     ///         by governance.
     mapping(address => bool) public isDeployerCoordinator;
 
-    /// @notice A mapping from deployed Hyperdrive instances to the deployer
-    ///         coordinator that deployed them. This is useful for verifying
-    ///         the bytecode that was used to deploy the instance.
+    /// @dev A mapping from deployed Hyperdrive instances to the deployer
+    ///      coordinator that deployed them. This is useful for verifying the
+    ///      bytecode that was used to deploy the instance.
     mapping(address instance => address deployCoordinator)
-        public instancesToDeployerCoordinators;
+        public _instancesToDeployerCoordinators;
 
     /// @dev Array of all instances deployed by this factory.
     address[] internal _instances;
@@ -760,7 +760,7 @@ contract HyperdriveFactory is IHyperdriveFactory {
 
         // Add this instance to the registry and emit an event with the
         // deployment configuration.
-        instancesToDeployerCoordinators[
+        _instancesToDeployerCoordinators[
             address(hyperdrive)
         ] = _deployerCoordinator;
         _config.governance = hyperdriveGovernance;
@@ -883,33 +883,35 @@ contract HyperdriveFactory is IHyperdriveFactory {
     }
 
     /// @notice Gets the instance at the specified index.
-    /// @param index The index of the instance to get.
+    /// @param _index The index of the instance to get.
     /// @return The instance at the specified index.
-    function getInstanceAtIndex(uint256 index) external view returns (address) {
-        return _instances[index];
+    function getInstanceAtIndex(
+        uint256 _index
+    ) external view returns (address) {
+        return _instances[_index];
     }
 
     /// @notice Returns the _instances array according to specified indices.
-    /// @param startIndex The starting index of the instances to get (inclusive).
-    /// @param endIndex The ending index of the instances to get (exclusive).
+    /// @param _startIndex The starting index of the instances to get (inclusive).
+    /// @param _endIndex The ending index of the instances to get (exclusive).
     /// @return range The resulting custom portion of the _instances array.
     function getInstancesInRange(
-        uint256 startIndex,
-        uint256 endIndex
+        uint256 _startIndex,
+        uint256 _endIndex
     ) external view returns (address[] memory range) {
         // If the indexes are malformed, revert.
-        if (startIndex >= endIndex) {
+        if (_startIndex >= _endIndex) {
             revert IHyperdriveFactory.InvalidIndexes();
         }
-        if (endIndex > _instances.length) {
+        if (_endIndex > _instances.length) {
             revert IHyperdriveFactory.EndIndexTooLarge();
         }
 
         // Return the range of instances.
-        range = new address[](endIndex - startIndex);
-        for (uint256 i = startIndex; i < endIndex; i++) {
+        range = new address[](_endIndex - _startIndex);
+        for (uint256 i = _startIndex; i < _endIndex; i++) {
             unchecked {
-                range[i - startIndex] = _instances[i];
+                range[i - _startIndex] = _instances[i];
             }
         }
     }
@@ -922,38 +924,51 @@ contract HyperdriveFactory is IHyperdriveFactory {
     }
 
     /// @notice Gets the deployer coordinator at the specified index.
-    /// @param index The index of the deployer coordinator to get.
+    /// @param _index The index of the deployer coordinator to get.
     /// @return The deployer coordinator at the specified index.
     function getDeployerCoordinatorAtIndex(
-        uint256 index
+        uint256 _index
     ) external view returns (address) {
-        return _deployerCoordinators[index];
+        return _deployerCoordinators[_index];
     }
 
     /// @notice Returns the deployer coordinators with an index between the
-    ///         starting and ending indexes (inclusive).
-    /// @param startIndex The starting index (inclusive).
-    /// @param endIndex The ending index (inclusive).
+    ///         starting and ending indexes.
+    /// @param _startIndex The starting index (inclusive).
+    /// @param _endIndex The ending index (exclusive).
     /// @return range The deployer coordinators within the specified range.
     function getDeployerCoordinatorsInRange(
-        uint256 startIndex,
-        uint256 endIndex
+        uint256 _startIndex,
+        uint256 _endIndex
     ) external view returns (address[] memory range) {
         // If the indexes are malformed, revert.
-        if (startIndex > endIndex) {
+        if (_startIndex >= _endIndex) {
             revert IHyperdriveFactory.InvalidIndexes();
         }
-        if (endIndex >= _deployerCoordinators.length) {
+        if (_endIndex > _deployerCoordinators.length) {
             revert IHyperdriveFactory.EndIndexTooLarge();
         }
 
         // Return the range of instances.
-        range = new address[](endIndex - startIndex + 1);
-        for (uint256 i = startIndex; i <= endIndex; i++) {
+        range = new address[](_endIndex - _startIndex);
+        for (uint256 i = _startIndex; i < _endIndex; i++) {
             unchecked {
-                range[i - startIndex] = _deployerCoordinators[i];
+                range[i - _startIndex] = _deployerCoordinators[i];
             }
         }
+    }
+
+    /// @notice Gets the deployer coordinators that deployed a list of instances.
+    /// @param __instances The instances.
+    /// @return coordinators The deployer coordinators.
+    function getDeployerCoordinatorByInstances(
+        address[] calldata __instances
+    ) external view returns (address[] memory coordinators) {
+        coordinators = new address[](_instances.length);
+        for (uint256 i = 0; i < __instances.length; i++) {
+            coordinators[i] = _instancesToDeployerCoordinators[__instances[i]];
+        }
+        return coordinators;
     }
 
     /// @dev Overrides the config values to the default values set by
--- a/contracts/src/factory/HyperdriveFactory.sol
+++ b/contracts/src/factory/HyperdriveFactory.sol
@@ -890,24 +890,24 @@ contract HyperdriveFactory is IHyperdriveFactory {
     }
 
     /// @notice Returns the _instances array according to specified indices.
-    /// @param startIndex The starting index of the instances to get.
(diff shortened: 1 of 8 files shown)


contracts/src/interfaces/IHyperdriveRegistry.sol
```diff
@@ -2,10 +2,58 @@
...
+        /// @dev The factory that deployed this instance.
+        address factory;
+        /// @dev The base token of the instance.
+        IERC20 baseToken;
+        /// @dev The vault shares token of the instance.
+        IERC20 vaultSharesToken;
+        /// @dev The name of the instance.
+        string name;
+        /// @dev The version of the instance.
+        string version;
```
mcclurejt, line 19: Why are we including this when it can be obtained from the instance itself?

Don't get me wrong, I'm all for wrapper functions or things that make obtaining instance information more convenient... I'm just not sure why it's beneficial to store these.

contracts/src/factory/HyperdriveRegistry.sol
```diff
@@ -46,17 +62,184 @@ contract HyperdriveRegistry is
 
     /// @inheritdoc IHyperdriveGovernedRegistry
     function setHyperdriveInfo(
```
mcclurejt, line 64: Very much personal preference but I kinda like the explicitness of separate add/update/remove as opposed to combining them

contracts/src/factory/HyperdriveRegistry.sol
```diff
@@ -46,17 +62,184 @@ contract HyperdriveRegistry is
...
+        }
+    }
+
+    /// @inheritdoc IHyperdriveRegistry
+    function getNumberOfHyperdriveInstances() external view returns (uint256) {
+        return _hyperdriveInstances.length;
+    }
+
+    /// @inheritdoc IHyperdriveRegistry
+    function getHyperdriveInstancesInRange(
```
mcclurejt, line 123: Love this

contracts/src/interfaces/IHyperdriveRegistry.sol
```diff
@@ -2,10 +2,58 @@
...
+    }
+
+    /// @notice Gets the number of Hyperdrive instances that have been registered.
+    /// @return The number of registered instances.
+    function getNumberOfHyperdriveInstances() external view returns (uint256);
```
mcclurejt, line 24: *pedantic:* feel like `count()` would work equally well here.

mcclurejt (commented):
tried to keep comments high-level since it's a wip

contracts/src/interfaces/IHyperdriveRegistry.sol
jalextowle: This is the struct that is returned in the getter. It's not actually stored. The struct used for storage is `HyperdriveInfoInternal`.

contracts/src/factory/HyperdriveRegistry.sol
```diff
@@ -18,9 +28,14 @@ contract HyperdriveRegistry is
     /// @notice The registry's governance address.
     address public governance;
 
-    /// @notice A mapping from hyperdrive instances to info associated with
-    ///         those instances.
-    mapping(address hyperdrive => uint256 data) internal _hyperdriveInfo;
+    /// @dev A list of all of the Hyperdrive pools that have been added to the
+    ///      Hyperdrive registry and have non-zero data.
+    address[] internal _hyperdriveInstances;
```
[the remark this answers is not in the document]
jalextowle: I'm keeping this consistent with the `HyperdriveFactory`. We've received this comment from auditors, and I didn't think it was worth changing.

contracts/src/interfaces/IHyperdriveRegistry.sol
jalextowle: I'm keeping this consistent with the `HyperdriveFactory`.

contracts/src/interfaces/IHyperdriveRegistry.sol
```diff
@@ -2,10 +2,58 @@
...
 
 interface IHyperdriveRegistry {
-    /// @notice Allows anyone to get the info for a hyperdrive instance.
-    /// @param _hyperdriveInstance The hyperdrive instance address.
-    /// @return The uint256 value set by governance.
+    /// @dev The info related to each Hyperdrive instance.
+    struct HyperdriveInfo {
+        /// @dev Data about the instance. Different registries can utilize
+        ///      different schemas for these values.
+        uint256 data;
```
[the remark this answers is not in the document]
jalextowle: From my perspective, `uint256` and `bytes32` offer the same room for creativity. They are both just 32 bytes. That said, I'm fine with switching this to `bytes32` since it would indicate that it's more of a "free form" field. 

I thought about changing it to `bytes` too. I like the idea, but I think we'll need a new function like `getHyperdriveDataSchema` that returns the ABI of the data field.

contracts/src/factory/HyperdriveRegistry.sol
```diff
@@ -46,17 +62,184 @@ contract HyperdriveRegistry is
...
+        address _instance,
+        uint256 _data,
+        address _factory
+    ) internal {
+        // Verify that the Hyperdrive instance was actually deployed by the
+        // factory. If the updated factory address is zero, we skip this check.
+        if (
+            _factory != address(0) &&
+            !IHyperdriveFactory(_factory).isInstance(_instance)
+        ) {
```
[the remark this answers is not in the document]
jalextowle: Given the current logic, `_addHyperdriveInfo` will only be called when the entry hasn't been added. If `_data[i] != 0 && data == 0` is the condition for this to be called. Similarly, `_updateHyperdriveInfo` will only be called if `_data[i] != 0 && data != 0`. 

With this in mind, I think the check would be superfluous.

contracts/src/factory/HyperdriveRegistry.sol
jalextowle: I hear you, but this will be more efficient for the use-cases that I'm imagining. Let's say that we need to make a bulk update to the registry (which is something that I anticipate happening in the future). We may need to add new entries, update others, and delete yet others. In this case, being able to call `setHyperdriveInfo` with the whole batch would be cheaper since you'd only need to call one function in one transaction. A related point is that the events are more consolidated. The only event you need to accumulate over to have an up-to-date state of the repo is `HyperdriveInfoUpdated`.

contracts/src/factory/HyperdriveRegistry.sol
```diff
@@ -29,6 +44,7 @@ contract HyperdriveRegistry is
         name = _name;
     }
 
+    /// @dev Ensures that the modified function is only called by governance.
     modifier onlyGovernance() {
```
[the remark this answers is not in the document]
jalextowle: Yeah that's fair. I had a similar thought. I'll change it.

contracts/src/interfaces/IHyperdriveRegistry.sol
slundqui: Very excited for the name field in the registry getter, my current method of generating a name is 3 contract calls per pool, so combining that to one call is great

contracts/src/factory/HyperdriveRegistry.sol
```diff
@@ -46,17 +62,184 @@ contract HyperdriveRegistry is
...
+        instances = new address[](_endIndex - _startIndex);
+        for (uint256 i = _startIndex; i < _endIndex; i++) {
+            instances[i] = _hyperdriveInstances[i];
+        }
+
+        return instances;
+    }
+
+    /// @inheritdoc IHyperdriveRegistry
+    function getHyperdriveInstancesAtIndex(
```
DannyDelott, line 145: typo: `getHyperdriveInstanceAtIndex` (singular)

ryangoree (approved):
Looks great! The API is really nice with flexible getters and streamlined setters. 🔥

contracts/src/factory/HyperdriveRegistry.sol
```diff
@@ -1,62 +1,392 @@
...
     /// @notice The registry's name.
     string public name;
 
     /// @notice The registry's version.
     string public constant version = VERSION;
 
-    /// @notice The registry's governance address.
-    address public governance;
+    /// @notice The registry's admin address.
+    address public admin;
```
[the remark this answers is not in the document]
jalextowle: Yep, exactlg

contracts/src/factory/HyperdriveFactory.sol
```diff
@@ -922,40 +924,53 @@ contract HyperdriveFactory is IHyperdriveFactory {
...
-        if (endIndex >= _deployerCoordinators.length) {
+        if (_endIndex > _deployerCoordinators.length) {
             revert IHyperdriveFactory.EndIndexTooLarge();
         }
 
         // Return the range of instances.
-        range = new address[](endIndex - startIndex + 1);
-        for (uint256 i = startIndex; i <= endIndex; i++) {
+        range = new address[](_endIndex - _startIndex);
+        for (uint256 i = _startIndex; i < _endIndex; i++) {
```
jrhea, line 954: up to you if it is worth it, but elsewhere we pre-increment an unchecked counter

contracts/src/factory/HyperdriveRegistry.sol
```diff
@@ -1,62 +1,392 @@
...
-    function getHyperdriveInfo(
-        address _hyperdriveInstance
-    ) external view override returns (uint256) {
-        return _hyperdriveInfo[_hyperdriveInstance];
+    function getInstancesInRange(
+        uint256 _startIndex,
+        uint256 _endIndex
+    ) external view returns (address[] memory instances) {
+        // If the indexes are malformed, revert.
+        if (_startIndex > _endIndex) {
```
jrhea, line 224: what if _startIndex == _endIndex? I think we should fail in this case bc it won't return anything, right?

contracts/src/interfaces/IHyperdriveFactory.sol
```diff
@@ -470,38 +483,61 @@ interface IHyperdriveFactory {
...
     /// @return range The deployer coordinators within the specified range.
     function getDeployerCoordinatorsInRange(
-        uint256 startIndex,
-        uint256 endIndex
+        uint256 _startIndex,
+        uint256 _endIndex
     ) external view returns (address[] memory range);
+
+    /// @notice Returns a flag indicating whether or not a deployer coordinator
+    ///         was is registered in this factory.
```
jrhea, line 529:
```suggestion
    ///         is registered in this factory.
```

contracts/src/interfaces/IHyperdriveFactory.sol
```diff
@@ -470,38 +483,61 @@ interface IHyperdriveFactory {
...
+    /// @param _deployerCoordinator The deployer coordinator to check.
+    /// @return The flag indicating whether or not a deployer coordinator
+    ///         was is registered in this factory.
```
jrhea, line 532:
```suggestion
    ///         is registered in this factory.
```

contracts/src/interfaces/IHyperdriveGovernedRegistry.sol
```diff
@@ -4,26 +4,82 @@ pragma solidity 0.8.20;
...
-        uint256 _data
+    /// @notice Gets the admin address of this registry.
+    /// @return The admin address of this registry.
+    function admin() external view returns (address);
+
+    /// @notice Allows admin to transfer the admin role.
+    /// @param _admin The new admin address.
+    function updateAdmin(address _admin) external;
+
+    /// @notice Allows governance to set arbitrary info for Hyperdrive factories.
```
jrhea, line 68:
```suggestion
    /// @notice Allows admin to set arbitrary info for Hyperdrive factories.
```

contracts/src/interfaces/IHyperdriveGovernedRegistry.sol
```diff
@@ -4,26 +4,82 @@ pragma solidity 0.8.20;
...
+    /// @param __factories The Hyperdrive factories to update.
+    /// @param _data The data associated with the factories.
+    function setFactoryInfo(
+        address[] memory __factories,
+        uint128[] memory _data
+    ) external;
+
+    /// @notice Allows governance to set arbitrary info for Hyperdrive instances.
```
jrhea, line 76:
```suggestion
    /// @notice Allows admin to set arbitrary info for Hyperdrive instances.
```

jrhea (commented):
Just a few nits/questions

jrhea (approved):

mcclurejt (approved):

contracts/src/factory/HyperdriveFactory.sol
```diff
@@ -883,33 +883,35 @@ contract HyperdriveFactory is IHyperdriveFactory {
...
+    function getInstanceAtIndex(
+        uint256 _index
+    ) external view returns (address) {
+        return _instances[_index];
     }
 
     /// @notice Returns the _instances array according to specified indices.
-    /// @param startIndex The starting index of the instances to get.
-    /// @param endIndex The ending index of the instances to get.
+    /// @param _startIndex The starting index of the instances to get (inclusive).
```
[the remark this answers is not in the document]
jalextowle: Yeah, that's right.

PR #1049, opened 2024-06-06
Add  `totalAssets(uint256 timestamp)` to MockERC4626

sentilesdal:
# Resolved Issues

# Description

This helps with testing.  When solidity is called using anvil in rust tests, the block is ticked by 1 second.  Because there is a call to `_getAccruedInterest()` in `totalAssets()`, the only way to get the exact value is to pass the current time plus one second to `totalAssets()`.

# Review Checklists

be removed.

## [[Reviewer Name]]

- [ ] **Tokens**
              token spend?
- [ ] **Reentrancy**
- [ ] **Gas Optimizations**
          memory variables?
          issues?
- [ ] **Visibility**
- [ ] **Math**
          catch underflows?
- [ ] **Testing**

### DIFF
--- a/contracts/test/MockERC4626.sol
+++ b/contracts/test/MockERC4626.sol
@@ -150,7 +150,13 @@ contract MockERC4626 is ERC4626, MultiRolesAuthority {
     }
 
     function totalAssets() public view override returns (uint256) {
-        return asset.balanceOf(address(this)) + _getAccruedInterest();
+        return
+            asset.balanceOf(address(this)) +
+            _getAccruedInterest(block.timestamp);
+    }
+
+    function totalAssets(uint256 timestamp) public view returns (uint256) {
+        return asset.balanceOf(address(this)) + _getAccruedInterest(timestamp);
     }
 
     /// Mock ///
@@ -178,14 +184,16 @@ contract MockERC4626 is ERC4626, MultiRolesAuthority {
     }
 
     function _accrue() internal {
-        uint256 interest = _getAccruedInterest();
+        uint256 interest = _getAccruedInterest(block.timestamp);
         if (interest > 0) {
             ERC20Mintable(address(asset)).mint(interest);
         }
         _lastUpdated = block.timestamp;
     }
 
-    function _getAccruedInterest() internal view returns (uint256) {
+    function _getAccruedInterest(
+        uint256 timestamp
+    ) internal view returns (uint256) {
         // If the rate is zero, no interest has accrued.
         if (_rate == 0) {
             return 0;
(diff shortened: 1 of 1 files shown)


contracts/test/MockERC4626.sol
```diff
@@ -207,4 +211,27 @@ contract MockERC4626 is ERC4626, MultiRolesAuthority {
...
+            return 0;
+        }
+
+        // base_balance = base_balance * (1 + r * t)
+        uint256 timeElapsed = (timestamp - _lastUpdated).divDown(
+            365 days
+        );
+        uint256 accrued = asset.balanceOf(address(this)).mulDown(
+            _rate.mulDown(timeElapsed)
+        );
```
ryangoree, line 234: Timestamp based accounting creates [challenges for the frontend](https://github.com/delvtech/hyperdrive-frontend/pull/1104) which requires [targeted patches](https://github.com/delvtech/hyperdrive-frontend/blob/main/packages/hyperdrive-js-core/src/hyperdrive/steth/ReadStEthHyperdrive.ts#L127-L134). Can we try a configured update interval?

sentilesdal: This doesn't add any functionality that doesn't already exist, it just allows us to see into the future so that rust tests can be more exact

sentilesdal: Or rather I should say this doesn't change any functionality, just adds an overloaded version of what's already there for rust tests.

contracts/test/MockERC4626.sol
```diff
@@ -207,4 +211,27 @@ contract MockERC4626 is ERC4626, MultiRolesAuthority {
         );
         return accrued;
     }
+
```
jalextowle, line 214: Why don’t we just update the other function to have a timestamp parameter?

contracts/test/MockERC4626.sol
```diff
@@ -207,4 +211,27 @@ contract MockERC4626 is ERC4626, MultiRolesAuthority {
...
+    ) internal view returns (uint256) {
+        // If the rate is zero, no interest has accrued.
+        if (_rate == 0) {
+            return 0;
+        }
+
+        // If the block timestamp is less than last updated, the accrual
+        // calculation will underflow. This can occur when using anvil's state
+        // snapshots.
+        if (block.timestamp < _lastUpdated) {
```
jalextowle, line 226: This should be timestamp, right?

sentilesdal: oh yeah it should

contracts/test/MockERC4626.sol
sentilesdal: Yeah I guess we could huh, I was trying to keep this PR footprint small.  Ok let me do that instead.

sentilesdal: done.

sentilesdal:
@jalextowle updated with your suggestion to just have one `_getAccruedInterest(uint256 timestamp)`

jalextowle (approved):

PR #1050, opened 2024-06-06
Checkpoint Incentivization

jalextowle:
# Description

This PR implements a system that can be used to incentivize checkpoint rewards. This system has been intentionally designed to be flexible enough to adjust to new demands and to be able to be ignored entirely. 

One interesting thing that came up while testing is that rewarding the minting of past checkpoints is potentially dangerous. This is dangerous because a minter could go back and mint all of the past checkpoints to drain the subrewarder's source address.

# Review Checklists

be removed.

## [[Reviewer Name]]

- [ ] **Tokens**
              token spend?
- [ ] **Reentrancy**
- [ ] **Gas Optimizations**
          memory variables?
          issues?
- [ ] **Visibility**
- [ ] **Math**
          catch underflows?
- [ ] **Testing**


contracts/src/interfaces/IHyperdriveRegistry.sol
```diff
@@ -53,13 +71,36 @@ interface IHyperdriveRegistry {
         uint256 _endIndex
     ) external view returns (address[] memory);
 
+    /// @notice Gets the hyperdrive factory info for a factory.
```
jrhea, line 74:
```suggestion
    /// @notice Gets the Hyperdrive factory info for a factory.
```

contracts/src/interfaces/IHyperdriveRegistry.sol
```diff
@@ -53,13 +71,36 @@ interface IHyperdriveRegistry {
...
+    /// @param _factory The factory.
+    /// @return The hyperdrive factory info.
```
jrhea, line 76:
```suggestion
    /// @return The Hyperdrive factory info.
```

contracts/src/internal/HyperdriveCheckpoint.sol
```diff
@@ -315,6 +325,23 @@ abstract contract HyperdriveCheckpoint is
...
         );
 
+        // Claim the checkpoint reward on behalf of the sender.
+        //
+        // NOTE: We do this in a low-level call and ignore the status to ensure
+        // that the checkpoint will be minted regardless of whether or not the
+        // call succeeds.
+        if (_checkpointRewarder != address(0)) {
+            bool isTrader = _isTrader; // avoid stack-too-deep
+            (bool _success, ) = _checkpointRewarder.call(
```
jrhea, line 335: weren't you going to wrap this in a try-catch?

contracts/src/rewarder/HyperdriveCheckpointRewarder.sol
```diff
@@ -0,0 +1,94 @@
...
+pragma solidity 0.8.20;
+
+import { IERC20 } from "../interfaces/IERC20.sol";
+import { IHyperdriveCheckpointRewarder } from "../interfaces/IHyperdriveCheckpointRewarder.sol";
+import { IHyperdriveCheckpointSubrewarder } from "../interfaces/IHyperdriveCheckpointSubrewarder.sol";
+import { VERSION } from "../libraries/Constants.sol";
+
+/// @author DELV
+/// @notice A checkpoint rewarder that is controlled by an admin and delegates
+///         it's reward functionality to a subrewarder.
```
jrhea, line 11:
```suggestion
///         its reward functionality to a subrewarder.
```

contracts/src/rewarder/HyperdriveCheckpointSubrewarder.sol
```diff
@@ -0,0 +1,156 @@
...
+    }
+
+    /// @notice Allows the admin to update the reward token.
+    /// @param _rewardToken The new reward token.
+    function updateRewardToken(IERC20 _rewardToken) external onlyAdmin {
+        rewardToken = _rewardToken;
+        emit RewardTokenUpdated(_rewardToken);
+    }
+
+    /// @notice Allows the admin to update the source.
```
jrhea, line 101: Suggestion to make it more clear:

    /// @notice Allows the admin to update the source address containing the rewards

test/units/rewarder/HyperdriveCheckpointRewarder.t.sol
```diff
@@ -0,0 +1,170 @@
+// SPDX-License-Identifier: Apache-2.0
```
jrhea, line 1: Maybe name the file `MockHyperdriveCheckpointSubrewarder.t.sol` so it matched the contract name

contracts/src/interfaces/IHyperdriveCheckpointSubrewarder.sol
```diff
@@ -0,0 +1,52 @@
+// SPDX-License-Identifier: Apache-2.0
+pragma solidity 0.8.20;
+
+import { IERC20 } from "./IERC20.sol";
+import { IHyperdriveRegistry } from "../interfaces/IHyperdriveRegistry.sol";
+
+// FIXME: Add the remaining functions to this interface.
```
mcclurejt, line 7: Reminder to fix

contracts/src/rewarder/HyperdriveCheckpointRewarder.sol
```diff
@@ -0,0 +1,94 @@
...
+    /// @param _subrewarder The subrewarder that processes checkpoint rewards.
+    function updateSubrewarder(
+        IHyperdriveCheckpointSubrewarder _subrewarder
+    ) external onlyAdmin {
+        subrewarder = _subrewarder;
+        emit SubrewarderUpdated(_subrewarder);
+    }
+
+    /// @notice Claims a checkpoint reward.
+    /// @param _claimant The address that is claiming the checkpoint reward.
```
mcclurejt, line 65: Why not use `@inheritdoc` here?

mcclurejt (commented):
Overall LGTM, would like to see the appropriate changes made to the deployment scripting as well :)

contracts/src/internal/HyperdriveCheckpoint.sol
```diff
@@ -315,6 +325,23 @@ abstract contract HyperdriveCheckpoint is
...
+                abi.encodeCall(
+                    IHyperdriveCheckpointRewarder.claimCheckpointReward,
+                    (msg.sender, checkpointTime, isTrader)
+                )
+            );
```
Saw-mon-and-Natalie, line 340: `returndata` bombing is possible here to cause OOG in `hyperdrive`:
- https://github.com/spearbit-audits/delv-week-review/issues/2

contracts/src/internal/HyperdriveCheckpoint.sol
jalextowle: `try-catch` doesn't catch all errors. This issue goes into some details: https://github.com/ethereum/solidity/issues/12725. The issue mentioned in the original post doesn't pertain to this, but there are other issues that can also result in reverts. Doing a low-level call is (unfortunately) safer.

test/units/rewarder/HyperdriveCheckpointRewarder.t.sol
jalextowle: The file is named after the test contract (the second contract defined here). This is also the pattern used in the reentrancy tests FWIW

contracts/src/interfaces/IHyperdriveRegistry.sol
```diff
@@ -53,13 +71,36 @@ interface IHyperdriveRegistry {
...
+    /// @param _factory The factory.
+    /// @return The Hyperdrive factory info with associated metadata.
+    function getFactoryInfoWithMetadata(
+        address _factory
+    ) external view returns (FactoryInfoWithMetadata memory);
+
+    /// @notice Gets the Hyperdrive factory info with associated metadata for a
+    ///         list of factories.
+    /// @param __factories The list of factories.
+    /// @return The hyperdrive factory info with associated metadata.
```
jrhea, line 99: This should indicate that a list is returned or at least use the word infos

contracts/src/interfaces/IHyperdriveRegistry.sol
```diff
@@ -79,10 +120,32 @@ interface IHyperdriveRegistry {
...
+    /// @param _instance The instance.
+    /// @return The instance info with associated metadata.
+    function getInstanceInfoWithMetadata(
+        address _instance
+    ) external view returns (InstanceInfoWithMetadata memory);
+
+    /// @notice Gets the instance info with associated metadata for a list of
+    ///         instances.
+    /// @param __instances The list of instances.
+    /// @return The instance info with associated metadata.
```
jrhea, line 147: should indicate that a list is returned

contracts/src/internal/HyperdriveCheckpoint.sol
```diff
@@ -315,6 +327,27 @@ abstract contract HyperdriveCheckpoint is
...
 
+        // Claim the checkpoint reward on behalf of the sender.
+        //
+        // NOTE: We do this in a low-level call and ignore the status to ensure
+        // that the checkpoint will be minted regardless of whether or not the
+        // call succeeds. Furthermore, we use the `ExcessivelySafeCall` library
+        // to prevent returndata bombing.
+        if (_checkpointRewarder != address(0)) {
+            bool isTrader = _isTrader; // avoid stack-too-deep
+            (bool _success, ) = _checkpointRewarder.excessivelySafeCall(
```
jrhea, line 338: can we just get rid of this assignment?

jrhea (approved):
lgtm. just a couple of nits. one thing to note, i had to eyeball the coverage bc coveralls diffs were timing out on me.

PR #1057, opened 2024-06-13
Improved the debugging tools

jalextowle:
# Description

This PR adds a new mode to the debugging tools for simulating a transaction that has already been mined. This addresses a limitation of the tool by allowing users to simulate a transaction at a specified block height and transaction height within the block. This will make the tool more useful on mainnet and testnet environments where we can't assume that each transaction will be executed in separate blocks. 

In addition to adding the new debugging mode, this PR also spruces up the tool and makes it more usable by relying more heavily on debugging-specific environment variables.

# Review Checklists

be removed.

## [[Reviewer Name]]

- [ ] **Tokens**
              token spend?
- [ ] **Reentrancy**
- [ ] **Gas Optimizations**
          memory variables?
          issues?
- [ ] **Visibility**
- [ ] **Math**
          catch underflows?
- [ ] **Testing**


test/debug/Debug.t.sol
```diff
@@ -0,0 +1,204 @@
...
+    /// @param _data The transaction calldata.
+    /// @return The name of the function that was called.
+    /// @return A flag indicating whether or not the target name was
+    ///         successfully found.
+    function getTargetName(
+        bytes memory _data
+    ) internal pure returns (string memory, bool) {
+        // Attempt to match the selector to one of the functions in the
+        // Hyperdrive interface. If the selector doesn't match any of the
+
```
mcclurejt, line 123:
```suggestion
```

foundry.toml
```diff
@@ -21,11 +21,8 @@ remappings = [
 ]
 # gas limit - max u64
 gas_limit = "18446744073709551615"
-# Allows the scripts to write the addresses to a file.
```
mcclurejt, line 24: good catch :)

mcclurejt (approved):
Awesome job! looking forward to using these

PR #1058, opened 2024-06-17
Use the previous weighted spot price instead of the current weighted spot price for `addLiquidity`

jalextowle:
# Resolved Issues

Fixes: https://github.com/spearbit-audits/delv-week-review/issues/13

# Description

This PR adds some tests to make sure that the weighted spot price behaves as expected with instantaneous trades and fixes an oversight where the current weighted spot price is used for the `addLiquidity` check instead of the previous weighted spot price.

The previous weighted spot price will never be `0` in `addLiquidity` because the current checkpoint is always minted with [`_applyCheckpoint`](https://github.com/delvtech/hyperdrive/blob/e5525e96fea5bc61e0eec7fe694049aa6722aa1c/contracts/src/internal/HyperdriveLP.sol#L196). In `_applyCheckpoint`, we always [update the previous weighted spot price](https://github.com/delvtech/hyperdrive/blob/e5525e96fea5bc61e0eec7fe694049aa6722aa1c/contracts/src/internal/HyperdriveCheckpoint.sol#L157). This uses the current spot price for the update, so unless the current spot price is `0`, the weighted spot price will be non-zero. Some tests were added to clarify this.

# Review Checklists

be removed.

## [[Reviewer Name]]

- [ ] **Tokens**
              token spend?
- [ ] **Reentrancy**
- [ ] **Gas Optimizations**
          memory variables?
          issues?
- [ ] **Visibility**
- [ ] **Math**
          catch underflows?
- [ ] **Testing**

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -62,3 +62,6 @@ deployments/
 
 # etherscan debugging
 etherscan_requests/
+
+# todos
+TODO.md
--- a/test/integrations/hyperdrive/CircuitBreakerTest.t.sol
+++ b/test/integrations/hyperdrive/CircuitBreakerTest.t.sol
@@ -261,6 +261,100 @@ contract CircuitBreakerTest is HyperdriveTest {
         assertFalse(actualSpotPrice == weightedSpotPriceBefore);
     }
 
+    function test_weighted_average_spot_price_instantaneous_long() external {
+        // Ensure a feasible fixed rate.
+        uint256 fixedRate = 0.05e18;
+
+        // Ensure a feasible time stretch fixed rate.
+        uint256 timeStretchFixedRate = fixedRate;
+
+        // Deploy the pool and initialize the market
+        IHyperdrive.PoolConfig memory config = testConfig(
+            timeStretchFixedRate,
+            POSITION_DURATION
+        );
+        config.circuitBreakerDelta = 1e18;
+        deploy(alice, config);
+        uint256 contribution = 10_000_000e18;
+        initialize(alice, fixedRate, contribution);
+
+        // Get the current weighted spot price.
+        uint256 weightedSpotPriceBefore = hyperdrive
+            .getCheckpoint(HyperdriveUtils.latestCheckpoint(hyperdrive))
+            .weightedSpotPrice;
+
+        // Open a large long in the same block.
+        openLong(alice, hyperdrive.calculateMaxLong());
+        uint256 spotPriceAfterLong = hyperdrive.calculateSpotPrice();
+
+        // Ensure that the weighted spot price is equal to the previous weighted
+        // spot price.
+        uint256 weightedSpotPriceAfter = hyperdrive
+            .getCheckpoint(HyperdriveUtils.latestCheckpoint(hyperdrive))
+            .weightedSpotPrice;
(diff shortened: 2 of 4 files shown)


mcclurejt (approved):
🤘

.gitignore
```diff
@@ -62,3 +62,6 @@ deployments/
 
 # etherscan debugging
 etherscan_requests/
+
+# todos
+TODO.md
```
jrhea, line 67: what is this?

test/integrations/hyperdrive/CircuitBreakerTest.t.sol
```diff
@@ -103,6 +103,57 @@ contract CircuitBreakerTest is HyperdriveTest {
                 })
             );
         }
+
+        // This test ensures that the circuit breaker won't be effected by
```
jrhea, line 107:
```suggestion
        // This test ensures that the circuit breaker won't be affected by
```

test/integrations/hyperdrive/CircuitBreakerTest.t.sol
```diff
@@ -103,6 +103,57 @@ contract CircuitBreakerTest is HyperdriveTest {
...
+        // changes to the spot rate in the current checkpoint.
+        {
+            // Ensure a feasible fixed rate.
+            uint256 fixedRate = 0.05e18;
+
+            // Ensure a feasible time stretch fixed rate.
+            uint256 timeStretchFixedRate = fixedRate;
+
+            // Deploy the pool and initialize the market
```
jrhea, line 116:
```suggestion
            // Deploy the pool and initialize the market.
```

test/integrations/hyperdrive/CircuitBreakerTest.t.sol
```diff
@@ -213,6 +264,100 @@ contract CircuitBreakerTest is HyperdriveTest {
...
     }
 
+    function test_weighted_average_spot_price_instantaneous_long() external {
+        // Ensure a feasible fixed rate.
+        uint256 fixedRate = 0.05e18;
+
+        // Ensure a feasible time stretch fixed rate.
+        uint256 timeStretchFixedRate = fixedRate;
+
+        // Deploy the pool and initialize the market
```
jrhea, line 274:
```suggestion
        // Deploy the pool and initialize the market.
```

test/integrations/hyperdrive/CircuitBreakerTest.t.sol
```diff
@@ -213,6 +264,100 @@ contract CircuitBreakerTest is HyperdriveTest {
...
+    }
+
+    function test_weighted_average_spot_price_instantaneous_short() external {
+        // Ensure a feasible fixed rate.
+        uint256 fixedRate = 0.05e18;
+
+        // Ensure a feasible time stretch fixed rate.
+        uint256 timeStretchFixedRate = fixedRate;
+
+        // Deploy the pool and initialize the market
```
jrhea, line 321:
```suggestion
        // Deploy the pool and initialize the market.
```

jrhea (approved):
just some nits

.gitignore
jalextowle: I like to keep a `TODO.md` file in the repo that tracks the stuff I need to do.

PR #1061, opened 2024-06-17
update rounding behavior in `calculateSolvencyAfterShort`

dpaiton:
# Resolved Issues
working towards https://github.com/delvtech/hyperdrive-rs/issues/29

# Description
The solvency after short check used `divDown` to convert from base to shares, but the open short function uses `divUp`. Therefore it was possible to have the solvency check say "all good" and then open short reverts. This fixes that problem.

I also added some intermediate checks so that the function fails gracefully in more situations.

# Review Checklists

be removed.

## [[Reviewer Name]]

- [ ] **Tokens**
              token spend?
- [ ] **Reentrancy**
- [ ] **Gas Optimizations**
          memory variables?
          issues?
- [ ] **Visibility**
- [ ] **Math**
          catch underflows?
- [ ] **Testing**

### DIFF
--- a/test/utils/HyperdriveUtils.sol
+++ b/test/utils/HyperdriveUtils.sol
@@ -1233,7 +1233,6 @@ library HyperdriveUtils {
         }
 
         // Calculate the pool's solvency after opening the short.
-
         uint256 tradeFees = (calculateShortCurveFee(
             _shortAmount,
             _spotPrice,
@@ -1255,7 +1254,6 @@ library HyperdriveUtils {
         }
 
         uint256 shareReserves = _params.shareReserves - tradeDelta;
-
         uint256 exposure = (_params.longExposure -
             uint256(_checkpointExposure.max(0))).divDown(
                 _params.vaultSharePrice
--- a/test/utils/HyperdriveUtils.sol
+++ b/test/utils/HyperdriveUtils.sol
@@ -1247,13 +1247,11 @@ library HyperdriveUtils {
         if (shareAmount < tradeFees) {
             return (0, false);
         }
-
-        uint256 tradeDelta = shareAmount - tradeFees;
-        if (_params.shareReserves < tradeDelta) {
+        uint256 shareReservesDelta = shareAmount - tradeFees;
+        if (_params.shareReserves < shareReservesDelta) {
             return (0, false);
         }
-
-        uint256 shareReserves = _params.shareReserves - tradeDelta;
+        uint256 shareReserves = _params.shareReserves - shareReservesDelta;
         uint256 exposure = (_params.longExposure -
             uint256(_checkpointExposure.max(0))).divDown(
                 _params.vaultSharePrice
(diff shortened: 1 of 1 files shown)


test/utils/HyperdriveUtils.sol
```diff
@@ -1233,19 +1233,27 @@ library HyperdriveUtils {
...
+            calculateShortGovernanceFee(
+                _shortAmount,
+                _spotPrice,
+                _params.curveFee,
+                _params.governanceLPFee
+            )).divUp(_params.vaultSharePrice);
+        if (shareAmount < tradeFees) {
+            return (0, false);
+        }
+
```
jalextowle, line 1250: Nit: Remove new line.

test/utils/HyperdriveUtils.sol
```diff
@@ -1233,19 +1233,27 @@ library HyperdriveUtils {
...
+        uint256 tradeDelta = shareAmount - tradeFees;
+        if (_params.shareReserves < tradeDelta) {
+            return (0, false);
+        }
+
```
jalextowle, line 1255: Nit: Remove new line

test/utils/HyperdriveUtils.sol
```diff
@@ -1233,19 +1233,27 @@ library HyperdriveUtils {
...
+                _shortAmount,
+                _spotPrice,
+                _params.curveFee,
+                _params.governanceLPFee
+            )).divUp(_params.vaultSharePrice);
+        if (shareAmount < tradeFees) {
+            return (0, false);
+        }
+
+        uint256 tradeDelta = shareAmount - tradeFees;
```
jalextowle, line 1251: Nit: Let’s use “shareReservesDelta” to be consistent.

test/utils/HyperdriveUtils.sol
```diff
@@ -1233,19 +1233,27 @@ library HyperdriveUtils {
...
-                    _spotPrice,
-                    _params.curveFee
-                ) -
-                    calculateShortGovernanceFee(
-                        _shortAmount,
-                        _spotPrice,
-                        _params.curveFee,
-                        _params.governanceLPFee
-                    )).divDown(_params.vaultSharePrice));
+        uint256 tradeFees = (calculateShortCurveFee(
```
jalextowle, line 1236: Let’s use “curveFee” to be consistent.

jalextowle (approved):
LGTM once the nits are addressed.

test/utils/HyperdriveUtils.sol
dpaiton: it is both the curveFee and the governanceCurveFee, which are treaded differently [elsewhere](https://github.com/delvtech/hyperdrive/blob/02efffd127f197629641c91d5baaae61b7247968/contracts/src/internal/HyperdriveBase.sol#L777-L789). With that in mind do you st ill want to use "curveFee"?

dpaiton: updating to `totalCurveFee` per outside discusison.

PR #1063, opened 2024-06-18
Updated Hyperdrive metadata

jalextowle:
# Description

This PR adds a `kind` function to each contract that provides a human-readable way to identify a given contract. In addition to this change, the registry was updated to support `kind`, and the `name` and `version` fields were added to any straggler contracts.

# Review Checklists

be removed.

## [[Reviewer Name]]

- [ ] **Tokens**
              token spend?
- [ ] **Reentrancy**
- [ ] **Gas Optimizations**
          memory variables?
          issues?
- [ ] **Visibility**
- [ ] **Math**
          catch underflows?
- [ ] **Testing**

### DIFF
--- a/contracts/src/deployers/HyperdriveDeployerCoordinator.sol
+++ b/contracts/src/deployers/HyperdriveDeployerCoordinator.sol
@@ -62,7 +62,7 @@ abstract contract HyperdriveDeployerCoordinator is
     /// @notice The contract used to deploy new instances of HyperdriveTarget3.
     address public immutable target3Deployer;
 
-    /// @notice The coordinator's name.
+    /// @notice The deployer coordinator's name.
     string public override name;
 
     /// @notice A mapping from deployment ID to deployment.
--- a/contracts/src/deployers/HyperdriveDeployerCoordinator.sol
+++ b/contracts/src/deployers/HyperdriveDeployerCoordinator.sol
@@ -109,12 +109,14 @@ abstract contract HyperdriveDeployerCoordinator is
     /// @notice Deploys a Hyperdrive instance with the given parameters.
     /// @dev This can only be deployed by the associated factory.
     /// @param _deploymentId The ID of the deployment.
+    /// @param __name The name of the Hyperdrive pool.
     /// @param _deployConfig The deploy configuration of the Hyperdrive pool.
     /// @param _extraData The extra data that contains the pool and sweep targets.
     /// @param _salt The create2 salt used to deploy Hyperdrive.
     /// @return The address of the newly deployed Hyperdrive instance.
     function deployHyperdrive(
         bytes32 _deploymentId,
+        string memory __name,
         IHyperdrive.PoolDeployConfig memory _deployConfig,
         bytes memory _extraData,
         bytes32 _salt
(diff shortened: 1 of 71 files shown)


contracts/src/libraries/Constants.sol
```diff
@@ -5,7 +5,55 @@ pragma solidity 0.8.20;
 address constant ETH = 0xEeeeeEeeeEeEeeEeEeEeeEEEeeeeEeeeeeeeEEeE;
 
 /// @dev The version of the contracts.
-string constant VERSION = "v1.0.11";
+string constant VERSION = "v1.0.12";
```
jalextowle, line 8: I went ahead and updated this so that we can cut a new release once this PR and the audit changes land.

contracts/src/factory/HyperdriveFactory.sol
```diff
@@ -730,6 +733,7 @@ contract HyperdriveFactory is IHyperdriveFactory {
     /// @param _deploymentId The deployment ID to use when deploying the pool.
     /// @param _deployerCoordinator The deployer coordinator to use in this
     ///        deployment.
+    /// @param __name The name of the Hyperdrive pool.
```
mcclurejt, line 736:
```suggestion
    /// @param __name The name of the Hyperdrive factory.
```

mcclurejt: for some reason delete button isn't working for this comment, please disregard :)

contracts/src/instances/lseth/LsETHTarget0.sol
```diff
@@ -25,10 +23,10 @@ contract LsETHTarget0 is HyperdriveTarget0, LsETHBase {
 
     /// Getters ///
 
-    /// @notice Returns the instance's name.
-    /// @return The instance's name.
-    function name() external pure override returns (string memory) {
-        _revert(abi.encode(NAME));
+    /// @notice Gets the instance's kind.
```
mcclurejt, line 26:
```suggestion
    /// @notice Returns the instance's kind.
```

mcclurejt (approved):

contracts/src/deployers/HyperdriveDeployerCoordinator.sol
```diff
@@ -62,23 +62,29 @@ abstract contract HyperdriveDeployerCoordinator is
     /// @notice The contract used to deploy new instances of HyperdriveTarget3.
     address public immutable target3Deployer;
 
+    /// @notice The coordinator's name.
```
jrhea, line 65:
```suggestion
    /// @notice The deployer coordinator's name.
```

contracts/src/instances/lseth/LsETHTarget0.sol
jrhea: ya some of the getters say "returns" and some say "gets"

jrhea (approved):
lgtm just a couple of comments and a dumb question

contracts/src/external/HyperdriveTarget0.sol
```diff
@@ -238,7 +238,13 @@ abstract contract HyperdriveTarget0 is
 
     /// @notice Gets the instance's name.
     /// @return The instance's name.
-    function name() external pure virtual returns (string memory);
+    function name() external view returns (string memory) {
```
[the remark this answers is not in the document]
jalextowle: It can't because the name is stored at constructor time. Unfortunately, strings can't be immutable, so it has to be `view`.

PR #1065, opened 2024-06-20
Implemented a circuit breaker APR

jalextowle:
# Resolved Issues

Fixes: https://github.com/spearbit-audits/delv-week-review/issues/15

# Description

This PR adds a knob that can be used to protect against k-inflation attacks that can occur with pools that have been configured with high time stretches.

# Review Checklists

be removed.

## [[Reviewer Name]]

- [x] **Tokens**
              token spend?
- [x] **Reentrancy**
- [x] **Gas Optimizations**
          memory variables?
          issues?
- [x] **Visibility**
- [x] **Math**
          catch underflows?
- [x] **Testing**


contracts/src/internal/HyperdriveLP.sol
```diff
@@ -200,7 +205,7 @@ abstract contract HyperdriveLP is
             true
         );
 
-        // Ensure that the spot APR is close enough to the previous weighted
+        // Ensure that the spot APR is less than the circuit close enough to the previous weighted
```
jrhea, line 208: The words in this comment aren't making sense in my head

jrhea: I consider explicitly enumerating these cases in the comment.  This is one of the most complicated conditionals in the code so it warrants an explanation

jrhea (dismissed):
lgtm. just a nit about a comment

Issue #980, opened 2024-04-16
Risk Playbook


jalextowle:
We should create a risk/operational playbook that goes through scenarios and how to deal with them.

jrhea:
I will work on organizing the spreadsheet by combining dupes and making the list more succinct.  Should have this done by Friday

jrhea:
https://docs.google.com/spreadsheets/d/1OD53RA_H5Z83twGQWxgwoJvSCRe5E-SCjB_BjqWiTrU/edit#gid=1467860037

jrhea:
When we deploy a new pool with a new version, Agent zero should be able to run the checkpoint bot. In other words, we need to be able to test it on mainnet fork.

jrhea:
When we evaluate a new market we need a standard suite of tests/analysis we do to configure it.

jrhea:
Here is the official Operations Playbook:

https://www.notion.so/delv-tech/Operations-Playbook-bc21c7663d0a4cedbc2d863bb46d90e7

dpaiton:
Closing since this is marked as Done.

PR #1066, opened 2024-06-20
add pauser script

jrhea:
# Description

Part of #980. This PR defines a batch pauser script that reads from the registry and pauses all active pools.

### DIFF
--- a/scripts/pause.sh
+++ b/scripts/pause.sh
@@ -3,7 +3,7 @@
 # This script will pause all instances that have data = 1.
 # NOTE: User must have MAINNET_RPC_URL, PAUSER_ADDRESS, and PAUSER_KEY set in .env file.
 #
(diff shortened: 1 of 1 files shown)


scripts/pause.sh
```diff
@@ -0,0 +1,44 @@
...
+#
+# Example usage: bash scripts/pause.sh
+
+source .env
+
+# Set the network
+RPC_URL=$MAINNET_RPC_URL
+
+# Hyperdrive Registry contract address
+HYPERDRIVE_REGISTRY_ADDRESS="0x03f6554299acf544ac646305800f57db544b837a"
```
mcclurejt, line 14: this probably shouldn't be hardcoded, would leave this as an input environment variable

mcclurejt (approved):
couple suggestions, nothing critical, do as you see fit

PR #1067, opened 2024-06-20
Implemented a price discovery circuit breaker on `addLiquidity` and `initialize`

jalextowle:
# Resolved Issues

# Description

This PR adds a simple check to `addLiquidity()` and `initialize()` to ensure that when liquidity is added, a spot rate of 0% is still achievable.

# Review Checklists

be removed.

## [[Reviewer Name]]

- [ ] **Tokens**
              token spend?
- [ ] **Reentrancy**
- [ ] **Gas Optimizations**
          memory variables?
          issues?
- [ ] **Visibility**
- [ ] **Math**
          catch underflows?
- [ ] **Testing**


contracts/src/internal/HyperdriveLP.sol
```diff
@@ -200,6 +201,70 @@ abstract contract HyperdriveLP is
...
+                checkpointExposure + maxBondProceeds.toInt256()
+            );
+
+            // If the pool isn't solvent after opening the max long, then we
+            // prevent the liquidity from being added since it will cause issues
+            // with price discovery.
+            if (
+                shareReserves.mulDown(vaultSharePrice) <
+                longExposure + _minimumShareReserves.mulUp(vaultSharePrice)
+            ) {
```
jrhea, line 263: can we use isSolvent()?

jrhea: i guess you would have to update isSolvent to not read from state directly

contracts/src/libraries/LPMath.sol
```diff
@@ -127,6 +127,113 @@ library LPMath {
...
+            ONE - _timeStretch,
+            _vaultSharePrice,
+            _initialVaultSharePrice
+        );
+
+        // If one fails they should both be zero.
+        if (maxSharePayment == 0 || maxBondProceeds == 0) {
+            maxSharePayment = 0;
+            maxBondProceeds = 0;
+        }
```
jrhea, line 210: We could do something more efficient in the case of adding liquidity at 0% APR, but i do think they should be able to add liquidity at 0% if they want to.

contracts/src/internal/HyperdriveLP.sol
```diff
@@ -103,6 +104,25 @@ abstract contract HyperdriveLP is
             revert IHyperdrive.InvalidEffectiveShareReserves();
         }
 
+        // Check to see whether or not the initial liquidity will result in
+        // invalid price discovery. If the spot price can't be brought to one,
+        // we fail to avoid dangerous pool states.
```
jalextowle, line 109: Nit: “we revert” is clearer IMO since there are two meanings of “we fail to avoid”

contracts/src/internal/HyperdriveLP.sol
```diff
@@ -292,6 +312,26 @@ abstract contract HyperdriveLP is
             revert IHyperdrive.DistributeExcessIdleFailed();
         }
 
+        // Check to see whether or not adding this liquidity will result in
+        // worsened price discovery. If the spot price can't be brought to one,
+        // we fail to avoid dangerous pool states.
```
jalextowle, line 317: Same here. “we revert” seems a bit clearer IMO

contracts/src/libraries/LPMath.sol
jalextowle: This makes me a bit nervous. Is the only time these calculations will fail when the APR is zero?

jrhea: I suggest playing with it some. let me ask a different question: why would we not allow people to add liquidity if we are solvent?

jalextowle: Two things:

1. we’re throwing away the result of the failure flag which seems unsafe to me
2. We’re assuming that if one is zero, they should both be zero. I don’t know why that’s a good assumption (especially considering (1))

jrhea: 1) my assumption here is the these two failure calcs are supposed to recreate the state that we would be in if we performed a max long. if one fails, then we are already in the state where a long can't be opened. that, in combination with the solvency constraint should still be valid. if we change it to trigger the circuit breaker on any of these conditions there are a bunch of perfectly reasonable and solvent states (if we believe our solvency equation) where people just can't add liquidity. it's not a good user experience bc if you revert and look at the logs you will see a bunch of cases where valid states don't allow you to add liquidity.

2) I understand; however, it seemed weird to test solvency by only applying one or the other. there are plenty of cases where one passes and the other fails. since only applying one doesn't capture the spirit of, "what's the state solvency if a max long is opened?" it seemed weird to only apply one. that being said, I could be convinced to change it to where we do this - from my pov it makes the solvency calc a little more conservative which is totally fine.

tl;Dr, I'm open to not doing (2). this is how I had it originally - I just wanted to explain the reasoning. regarding (1), I suggest we talk this out

mcclurejt: it seems like throwing out the failure flag is fine in this case, since checking if either is 0 covers the cases where it failed and where maxbuy is 0.... 

are there any cases where one would return 0 (without failing) and the other would return something nonzero?

test/integrations/hyperdrive/PriceDiscovery.t.sol
```diff
@@ -0,0 +1,569 @@
...
+        } else {
+            // This condition doesn't make sense. It should never happen.
+            assertTrue(false);
+        }
+    }
+
+    // This test fuzzes over the partial range of inputs, but
+    // verifies that the low spot rate is approximately 0 and
+    // that the high spot rate >= initial rate.
+    function test_priceDiscovery_parial_range_fuzz(
```
jrhea, line 342:
```suggestion
    function test_priceDiscovery_partial_range_fuzz(
```

contracts/src/libraries/LPMath.sol
jalextowle: I don’t think the case where it failed and the max buy is zero is the issue. The issue is where it failed, the max buy is nonzero, and this implementation treats it like the max buy is zero. In this case, we’re not checking the solvency at the true max buy. That said, I haven’t looked through the max buy implementation to see if there are cases where it would actually fail in this way, but that’s what I would be concerned about.

test/integrations/hyperdrive/PriceDiscovery.t.sol
```diff
@@ -0,0 +1,569 @@
...
+        } else {
+            openShort(alice, hyperdrive.calculateMaxShort());
+        }
+
+        // The checkpoint passes.
+        advanceTime(CHECKPOINT_DURATION, 0);
+
+        // Alice opens a max long or a max short.
+        if (longFirst) {
+            openShort(alice, hyperdrive.calculateMaxShort());
```
jalextowle, line 130: IMO, it would be good to test with amounts other than the max long and short. I’m not saying this isn’t sufficient, but without looking at this closer (which I will do tomorrow on the flights), I still have a few cases that I’m unsure about.

contracts/src/libraries/LPMath.sol
```diff
@@ -127,6 +127,124 @@ library LPMath {
...
+        uint256 shareReserves = _shareReserves + maxSharePayment;
+        uint256 longExposure = calculateLongExposure(
+            _longExposure,
+            _checkpointExposure,
+            _checkpointExposure + maxBondProceeds.toInt256()
+        );
+
+        // If the pool isn't solvent after opening the max long, then we
+        // prevent the liquidity from being added since it will cause issues
+        // with price discovery. We know that when: cz <= e_l + c * z_min
```
Saw-mon-and-Natalie, line 235: should be `c * z > e_l + c * z_min`

contracts/src/libraries/LPMath.sol
```diff
@@ -127,6 +127,124 @@ library LPMath {
...
+        // is within some epilson of being true, the pool is still solvent,
+        // will have trouble price discovering back to 0%. We add an additional
+        // c * z_min to the right hand side to account for this epsilon.
+        if (
+            shareReserves.mulDown(_vaultSharePrice) <=
+            longExposure + 2 * _minimumShareReserves.mulUp(_vaultSharePrice)
+        ) {
+            return false;
+        }
```
Saw-mon-and-Natalie, line 244: I think beside this check we should also check that the updated point has a spot price close to one.

contracts/src/libraries/LPMath.sol
```diff
@@ -127,6 +127,124 @@ library LPMath {
...
+
+        // If one fails they should both be zero.
+        if (maxSharePayment == 0 || maxBondProceeds == 0) {
+            maxSharePayment = 0;
+            maxBondProceeds = 0;
+        }
+
+        // Calculate the pool's solvency after opening the max long. This
+        // doesn't account for fees, which is fine since this will be more
+        // conservative.
```
Saw-mon-and-Natalie, line 225: not accounting for fees is only conservative for `longExposure` side. To be more conservative one should apply fees to the `shareReserves`.

contracts/src/libraries/LPMath.sol
```diff
@@ -127,6 +127,124 @@ library LPMath {
...
+            _bondReserves,
+            ONE - _timeStretch,
+            _vaultSharePrice,
+            _initialVaultSharePrice
+        );
+
+        // If one fails they should both be zero.
+        if (maxSharePayment == 0 || maxBondProceeds == 0) {
+            maxSharePayment = 0;
+            maxBondProceeds = 0;
```
Saw-mon-and-Natalie, line 220: I think calculations should follow the logic from `_calculateOpenLong` where bond reserve delta is given derived from the given share reserve delta.

contracts/src/libraries/LPMath.sol
jalextowle: I reviewed this, and what you did makes sense to me @jrhea. I added a comment or two to explain the reasoning, but this part of the implementation was basically left unchanged.

contracts/src/internal/HyperdriveLP.sol
```diff
@@ -287,21 +329,47 @@ abstract contract HyperdriveLP is
...
+            _marketState.shareReserves,
+            _marketState.shareAdjustment,
+            _marketState.bondReserves,
+            vaultSharePrice_,
+            _marketState.longExposure,
+            _nonNettedLongs(latestCheckpoint_ + _positionDuration)
+        );
+        if (
+            !success ||
+            solvencyAfterMaxLongAfter < solvencyAfterMaxLongBefore.min(0)
```
jrhea, line 359: why do we need the min(0)?

jalextowle: Good question. I can add a comment to clarify. This is the logic that I added to handle the case where adding liquidity improves price discovery. If solvencyAfterMaxLongBefore is less than 0, we make sure that solvencyAfterMaxLongAfter is larger than solvencyAfterMaxLongBefore. Otherwise, we just make sure it’s greater than zero

contracts/src/internal/HyperdriveBase.sol
```diff
@@ -682,6 +677,152 @@ abstract contract HyperdriveBase is IHyperdriveEvents, HyperdriveStorage {
...
+        uint256 longExposure = LPMath.calculateLongExposure(
+            _longExposure,
+            _checkpointExposure,
+            _checkpointExposure + maxBondProceeds.toInt256()
+        );
+        uint256 vaultSharePrice = _vaultSharePrice;
+        return (
+            shareReserves.mulDown(vaultSharePrice).toInt256() -
+                longExposure.toInt256() -
+                _minimumShareReserves.mulUp(vaultSharePrice).toInt256(),
```
jrhea, line 765: this was 2*z_min before, what did you do to make this work without the padding?

contracts/src/internal/HyperdriveBase.sol
```diff
@@ -682,6 +677,152 @@ abstract contract HyperdriveBase is IHyperdriveEvents, HyperdriveStorage {
...
+            _bondReserves,
+            _initialVaultSharePrice,
+            _timeStretch
+        );
+        (maxSharePayment, maxBondProceeds, ) = _calculateOpenLongFees(
+            maxSharePayment,
+            maxBondProceeds,
+            _vaultSharePrice,
+            spotPrice
+        );
```
jrhea, line 752: i'm just curious, couldn't we have used _calculateOpenLong now that we are accounting for fees?

contracts/src/internal/HyperdriveBase.sol
```diff
@@ -682,6 +677,152 @@ abstract contract HyperdriveBase is IHyperdriveEvents, HyperdriveStorage {
...
+    ///      the YieldSpace curve.
+    /// @param _shareReserves The pool's share reserves.
+    /// @param _shareAdjustment The pool's share adjustment.
+    /// @param _bondReserves The pool's bond reserves.
+    /// @param _vaultSharePrice The vault share price.
+    /// @param _longExposure The pool's long exposure.
+    /// @param _checkpointExposure The pool's checkpoint exposure.
+    /// @return The solvency after opening the max long.
+    /// @return A flag indicating whether or not the calculation succeeded.
+    function _calculateSolvencyAfterMaxLongSafe(
```
jrhea, line 691: just curious why you moved this to HyperdriveBase from LPMath

jalextowle: It ends up being cheaper because we ran into stack too deep issues. Within hyperdrive base, I can access the immutables without having to create new stack variables for them. I actually moved it into the library to try to fix the code size issue, but it made it worse and forced me to create a memory struct which would increase gas costs.

contracts/src/internal/HyperdriveBase.sol
jalextowle: We can’t because _calculateOpenLong includes several invariant checks that we don’t want to fail for this calculation (the negative interest checks and the checks within YieldSpaceMath). I could make it a safe function that returns the bytes for an error, which would allow us to reuse the code. I actually kind of like that idea, and we can chat more about it tomorrow

contracts/src/internal/HyperdriveBase.sol
jalextowle: Let’s talk about that tomorrow. I suspect it has to do with the improvements we made to the invariant check, but I was actually somewhat confused about why we needed it before since all of the tests passed without it (unless I’m missing something)

contracts/src/internal/HyperdriveBase.sol
jrhea: okay ya. that makes sense

contracts/src/internal/HyperdriveLP.sol
jrhea: okay okay i follow it now. it was late and i was reading it wrong. ya this fails if the solvency gets worse (when before and after are negative) or if just after is negative.

contracts/src/internal/HyperdriveBase.sol
jrhea: it's possible the inclusion of the fees into the calc fixed it?

contracts/src/internal/HyperdriveLP.sol
```diff
@@ -103,6 +104,25 @@ abstract contract HyperdriveLP is
             revert IHyperdrive.InvalidEffectiveShareReserves();
         }
 
+        // Check to see whether or not the initial liquidity will result in
+        // invalid price discovery. If the spot price can't be brought to one,
+        // we revert to avoid dangerous pool states.
+        uint256 latestCheckpoint = _latestCheckpoint();
```
mcclurejt, line 110: Why call this here as opposed to where it was before?

contracts/src/libraries/YieldSpaceMath.sol
```diff
@@ -63,7 +97,7 @@ library YieldSpaceMath {
 
         // If k < ze, we have no choice but to revert.
```
mcclurejt, line 98: might want to change this since it's a "safe" call

test/integrations/hyperdrive/PriceDiscovery.t.sol
```diff
@@ -0,0 +1,673 @@
...
+
+    function test_solvency_at_0_apr(
+        uint256 fixedAPR,
+        uint256 initialContribution,
+        uint256 addLiquidityContribution
+    ) external {
+        uint256 minimumShareReserves = 10e18;
+
+        // Normalize the fuzzing parameters to a reasonable range.
+        fixedAPR = fixedAPR.normalizeToRange(0.01e18, 2e18);
```
mcclurejt, line 23: I'm wondering whether it'd be helpful to use the same fuzzing parameters across all tests... like all fuzz tests that modify fixedAPR use the same lower and upper bound? Initial thinking is that it'd increase confidence in pools with those ranges at the cost of decreased specificity.

mcclurejt (approved):
a couple comments but nothing blocking... LGTM 👍

test/integrations/hyperdrive/PriceDiscovery.t.sol
jalextowle: I like having different ranges for different tests. A lot of the tests (all of the `test_priceDiscovery` tests) are doing the same things with different parameters. By having both, we get to fuzz across the full range but we also concentrate.

contracts/src/internal/HyperdriveBase.sol
```diff
@@ -682,6 +677,152 @@ abstract contract HyperdriveBase is IHyperdriveEvents, HyperdriveStorage {
...
+                _spotPrice,
+                _vaultSharePrice
+            );
+
+        // Calculate the impact of the curve fee on the bond reserves. The curve
+        // fee benefits the LPs by causing less bonds to be deducted from the
+        // bond reserves.
+        _bondReservesDelta -= curveFee;
+
+        // NOTE: Round down to underestimate the governance fee.
```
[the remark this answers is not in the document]
jalextowle: We underestimate governance fees because we want to err on the side of giving LPs more. It's a < 10 wei difference, so it's not that big of a deal.

test/integrations/hyperdrive/PriceDiscovery.t.sol
```diff
@@ -1,12 +1,16 @@
 // SPDX-License-Identifier: Apache-2.0
 pragma solidity 0.8.20;
 
+// FIXME
+import { console2 as console } from "forge-std/console2.sol";
```
jrhea: delete

jrhea (approved):
lgtm

contracts/src/libraries/LPMath.sol
jalextowle: We acknowledge this, but we've decided not to include this check because (1) fees will result in the spot price not being equal to 1 and (2) the method that we are using is mathematically equivalent when there are no fees.

Issue #976, opened 2024-04-16
Deploy the mainnet contracts


jalextowle:
- [x] Create a new DELV deployer address
- [x] `HyperdriveFactory`
- [x] `HyperdriveRegistry`
- [x] All of the deployer coordinators that have been implemented
- [x] All of the sub-deployers
- [x] Use the configuration from the snapshot

jalextowle:
This is blocked because we need to hear back from @CPSTL and legal to know when we can deploy the contracts.

PR #1072, opened 2024-06-26
Implemented the mainnet deployment scripts

jalextowle:
# Resolved Issues

Addresses: #976, https://github.com/spearbit-audits/delv-week-review/issues/2

# Description

This PR adds the mainnet deployment scripts for the ElementDAO deployment and addresses the last remaining issues from Saw Mon. 

# Review Checklists

be removed.

## [[Reviewer Name]]

- [ ] **Tokens**
              token spend?
- [ ] **Reentrancy**
- [ ] **Gas Optimizations**
          memory variables?
          issues?
- [ ] **Visibility**
- [ ] **Math**
          catch underflows?
- [ ] **Testing**


tasks/deploy/hyperdrive.ts
```diff
@@ -23,7 +23,8 @@ HyperdriveDeployBaseTask(
 
         // deploy the registry
         await run("deploy:registry", {
-            name: `${network.name.toUpperCase()}_REGISTRY`,
+            // TODO: Generalize this.
+            name: "ElementDAO Hyperdrive Registry",
```
jrhea, line 27: why elementdao? this is our registry?  why not delv Hyperdrive registry?

tasks/deploy/lib/constants.ts
```diff
@@ -23,3 +26,5 @@ export const MAINNET_STETH_WHALE =
 
 export const MAINNET_RETH_WHALE =
     "0x1bee69b7dfffa4e2d53c2a2df135c388ad25dcd2" as Address;
+
+export const SIX_MONTHS = "182 days";
```
jrhea, line 30: why not 525,600/2 minutes?

jrhea: ed3e1892-63bf-4a63-8028-2a5cc4543613_text

tasks/deploy/hyperdrive.ts
jalextowle: Yeah, that's fair. I'll change it.

mcclurejt (approved):

tasks/deploy/lib/constants.ts
jalextowle: It's not divisible by a day

jrhea (approved):

PR #1080, opened 2024-07-02
mainnet "clone"

mcclurejt:
# Resolved Issues

- Adds script for generating a mainnet "clone" with `artifacts/addresses.json` pointing to the current mainnet instance.
- Generalizes account funding script
- Fix unresolved factory address in `artifacts/addresses.json`
- Isolate configuration imports to their respective `hardhat.config.ts` files

# Description

# Review Checklists

be removed.

## [[Reviewer Name]]

- [ ] **Tokens**
              token spend?
- [ ] **Reentrancy**
- [ ] **Gas Optimizations**
          memory variables?
          issues?
- [ ] **Visibility**
- [ ] **Math**
          catch underflows?
- [ ] **Testing**


hardhat.config.mainnet_fork.ts
```diff
@@ -30,8 +30,8 @@ const config: HardhatUserConfig = {
     networks: {
         mainnet_fork: {
             live: false,
-            url: env.HYPERDRIVE_ETHEREUM_URL ?? "http://anvil:8545",
-            accounts: [env.DEPLOYER_PRIVATE_KEY ?? DEFAULT_PK],
+            url: env.RPC_URL ?? "http://anvil:8545",
```
jalextowle, line 33: Why didn't this change in all of the config files?

jalextowle (approved):

hardhat.config.mainnet_fork.ts
mcclurejt: Good catch, was initially considering updating some/all of the environment variables to match what is used in `hyperdrive-infra`, however when going through it plus the `hyperdrive-infra` integration it seemed sizeable enough to do in a separate PR.

Reverted to what it was originally (as well as updated clone config to match)

PR #1094, opened 2024-07-12
Added a Morpho Blue integration

jalextowle:
# Resolved Issues

# Description

TODO

# Review Checklists

be removed.

## [[Reviewer Name]]

- [ ] **Tokens**
              token spend?
- [ ] **Reentrancy**
- [ ] **Gas Optimizations**
          memory variables?
          issues?
- [ ] **Visibility**
- [ ] **Math**
          catch underflows?
- [ ] **Testing**


contracts/src/deployers/morpho-blue/MorphoBlueTarget4Deployer.sol
```diff
@@ -0,0 +1,40 @@
...
+/// @custom:disclaimer The language used in this code is for coding convenience
+///                    only, and is not intended to, and does not, have any
+///                    particular legal or regulatory significance.
+contract MorphoBlueTarget4Deployer is IHyperdriveTargetDeployer {
+    /// @notice Deploys a target4 instance with the given parameters.
+    /// @param _config The configuration of the Hyperdrive pool.
+    /// @param _extraData The extra data for the Morpho instance. This contains
+    ///        the market parameters that weren't specified in the config.
+    /// @param _salt The create2 salt used in the deployment.
+    /// @return The address of the newly deployed MorphoBlueTarget3 instance.
```
jrhea, line 21:
```suggestion
    /// @return The address of the newly deployed MorphoBlueTarget4 instance.
```

codegen/templates/deployers/HyperdriveCoreDeployer.sol.jinja
```diff
@@ -19,6 +19,7 @@ contract {{ name.capitalized }}HyperdriveCoreDeployer is IHyperdriveCoreDeployer
     /// @param _target1 The target1 address.
     /// @param _target2 The target2 address.
     /// @param _target3 The target3 address.
+    /// @param _target4 The target4 address.
```
mcclurejt, line 22: boooooo 👎

contracts/src/instances/morpho-blue/MorphoBlueBase.sol
```diff
@@ -0,0 +1,221 @@
...
+    /// @param _baseAmount The base amount to deposit.
+    /// @param _extraData Additional data to pass to the Morpho vault. This
+    ///        should be zero if it is unused.
+    /// @return The shares that were minted in the deposit.
+    /// @return The amount of ETH to refund. Since this yield source isn't
+    ///         payable, this is always zero.
+    function _depositWithBase(
+        uint256 _baseAmount,
+        bytes calldata _extraData
+    ) internal override returns (uint256, uint256) {
```
mcclurejt, line 76: nit: I feel like any function with multiple return values should use named values (especially if they're the same type)

mcclurejt (approved):
lgtm, comments aren't blocking

contracts/src/instances/morpho-blue/MorphoBlueBase.sol
jalextowle: I like named returns, but I don't agree with this in general. For functions like `openShort`, we need the flexibility to manage the stack more carefully than this would allow. In this case, I'm fine with using named returns, so I'll add them.

contracts/src/instances/morpho-blue/MorphoBlueConversions.sol
```diff
@@ -0,0 +1,121 @@
...
+                _lltv
+            );
+
+        return _shareAmount.toAssetsDown(totalSupplyAssets, totalSupplyShares);
+    }
+
+    /// @dev Convert an amount of base to an amount of vault shares.
+    /// @param _baseToken The base token underlying the Aave vault.
+    /// @param _vault The Aave vault.
+    /// @param _baseToken The collateral token for this Morpho Blue market.
```
jrhea, line 58:
```suggestion
    /// @param _collateralToken The collateral token for this Morpho Blue market.
```

contracts/src/instances/morpho-blue/MorphoBlueConversions.sol
```diff
@@ -0,0 +1,121 @@
...
+                _oracle,
+                _irm,
+                _lltv
+            );
+
+        return _shareAmount.toAssetsDown(totalSupplyAssets, totalSupplyShares);
+    }
+
+    /// @dev Convert an amount of base to an amount of vault shares.
+    /// @param _baseToken The base token underlying the Aave vault.
```
jrhea, line 56:
```suggestion
```

jrhea, line 56:
```suggestion
    /// @param _vault The Aave vault.
```

contracts/src/instances/morpho-blue/MorphoBlueConversions.sol
```diff
@@ -0,0 +1,121 @@
...
+    /// @param _vault The Aave vault.
```
jrhea, line 57:
```suggestion
    /// @param _baseToken The base token underlying the Morpho Blue vault.
```

contracts/src/instances/morpho-blue/MorphoBlueConversions.sol
```diff
@@ -0,0 +1,121 @@
...
+                _oracle,
+                _irm,
+                _lltv
+            );
+
+        return _baseAmount.toSharesDown(totalSupplyAssets, totalSupplyShares);
+    }
+
+    /// @dev Gets the Morpho Blue supply balances after accruing interest.
+    /// @param _vault The Aave vault.
```
jrhea, line 90:
```suggestion
    /// @param _vault The Morpho Blue vault.
```

PR #1096, opened 2024-07-16
Everlong interface + skeleton contract

mcclurejt:
# Resolved Issues

- closes https://github.com/delvtech/hyperdrive/issues/1082

# Description

This PR adds  VERY basic interfaces and a skeleton contract for Everlong.

The interfaces included are not exhaustive and intentionally limited to methods critical to core functionality. I fully expect that additional features will be added as development progresses. This way, entire features (interface + implementation + tests) can be encapsulated in single pull requests... hopefully resulting in a nice clean changelog 🤞 

# Review Checklists

## @jalextowle 

- [ ] **Tokens**
              token spend?
- [ ] **Reentrancy**
- [ ] **Gas Optimizations**
          memory variables?
          issues?
- [ ] **Visibility**
    
## @jrhea  

- [ ] **Tokens**
              token spend?
- [ ] **Reentrancy**
- [ ] **Gas Optimizations**
          memory variables?
          issues?
- [ ] **Visibility**


contracts/src/interfaces/IEverlongPositions.sol
```diff
@@ -0,0 +1,43 @@
...
+        uint128 bondAmount,
+        uint256 index
+    );
+
+    /// @notice Emitted when an existing position's `bondAmount` is modified.
+    event PositionUpdated(
+        uint128 indexed maturityTime,
+        uint128 newBondAmount,
+        uint256 index
+    );
```
jalextowle, line 24: I feel like something like `Deposit` and `Withdrawal` might be clearer.

contracts/src/everlong/EverlongPositions.sol
```diff
@@ -0,0 +1,34 @@
+// SPDX-License-Identifier: Apache-2.0
+pragma solidity 0.8.20;
+
+import { IEverlongPositions } from "../interfaces/IEverlongPositions.sol";
+import { DoubleEndedQueue } from "openzeppelin/utils/structs/DoubleEndedQueue.sol";
+
+contract EverlongPositions is IEverlongPositions {
+    using DoubleEndedQueue for DoubleEndedQueue.Bytes32Deque;
+    DoubleEndedQueue.Bytes32Deque internal _positions;
```
jalextowle, line 9: I'm not convinced we need a deque. I think we should start with a mapping and switch to this if necessary. I only see us needing to traverse backwards, if at all. We can do this with a queue.

contracts/src/everlong/EverlongPositions.sol
```diff
@@ -0,0 +1,34 @@
+// SPDX-License-Identifier: Apache-2.0
+pragma solidity 0.8.20;
+
+import { IEverlongPositions } from "../interfaces/IEverlongPositions.sol";
+import { DoubleEndedQueue } from "openzeppelin/utils/structs/DoubleEndedQueue.sol";
+
+contract EverlongPositions is IEverlongPositions {
```
jalextowle, line 7: Let's add contract-level Natspec to all of these contracts. This would explain what the contract is used for. In this case, it should clearly identify what positions are stored in Everlong and what they are used for.

contracts/src/everlong/EverlongERC4626.sol
```diff
@@ -0,0 +1,96 @@
...
+    /// @notice Decimals used by the underlying Hyperdrive asset.
+    uint8 internal immutable _decimals;
+
+    /// @notice Name of the Everlong token.
+    string internal _name;
+
+    /// @notice Symbol of the Everlong token.
+    string internal _symbol;
+
+    constructor(
```
jalextowle, line 26: We're missing Natspec here.

contracts/src/everlong/EverlongERC4626.sol
```diff
@@ -0,0 +1,96 @@
...
+    address internal immutable _underlying;
+
+    /// @notice Decimals used by the underlying Hyperdrive asset.
+    uint8 internal immutable _decimals;
+
+    /// @notice Name of the Everlong token.
+    string internal _name;
+
+    /// @notice Symbol of the Everlong token.
+    string internal _symbol;
```
jalextowle, line 24: We need `kind` too.

contracts/src/everlong/EverlongERC4626.sol
```diff
@@ -0,0 +1,96 @@
...
+
+contract EverlongERC4626 is ERC4626 {
+    /// @notice Whether virtual shares will be used to mitigate the inflation attack.
+    bool public constant useVirtualShares = true;
+
+    /// @notice Used to reduce the feasibility of an inflation attack.
+    uint8 public constant decimalsOffset = 3;
+
+    /// @notice Address of the underlying Hyperdrive instance.
+    address internal immutable _underlying;
```
jalextowle, line 15: IMO, it would be clearer for this to be `hyperdrive`.

contracts/src/everlong/EverlongERC4626.sol
```diff
@@ -0,0 +1,96 @@
...
+        _symbol = symbol_;
+    }
+
+    /// @dev Address of the underlying Hyperdrive instance.
+    ///
+    /// - MUST be an ERC20 token contract.
+    /// - MUST NOT revert.
+    function asset() public view virtual override returns (address) {
+        return _underlying;
+    }
```
jalextowle, line 46: Hyperdrive isn't an ERC20 token contract. It may make more sense for this to be the idle liquidity. I was thinking of this recently, but Everlong may not fit that well into the ERC4626 standard because of withdrawal shares. Just keep that in mind.

contracts/src/everlong/EverlongERC4626.sol
```diff
@@ -0,0 +1,96 @@
...
+        override
+        returns (uint8)
+    {
+        return _decimals;
+    }
+
+    /// @dev A non-zero value used to make the inflation attack even more unfeasible.
+    /// Only used when {_useVirtualShares} returns true.
+    ///
+    /// - MUST NOT revert.
```
jalextowle, line 82: Our style for this kind of comment is to make an extra `dev` comment for this.

contracts/src/everlong/EverlongAdmin.sol
```diff
@@ -0,0 +1,22 @@
...
+    modifier onlyAdmin() {
+        if (msg.sender != admin) {
+            revert IEverlongAdmin.Unauthorized();
+        }
+        _;
+    }
+
+    /// @inheritdoc IEverlongAdmin
+    function setAdmin(address _admin) external onlyAdmin {
+        admin = _admin;
```
jalextowle, line 20: We'll want an event for this.

contracts/src/interfaces/IEverlongPositions.sol
jalextowle: I think I misread this initially. Since this is referring to bonds, why don't we just call it `IEverlongBonds` or something like that? This would make it immediately clear what we're talking about.

contracts/src/everlong/EverlongERC4626.sol
```diff
@@ -0,0 +1,96 @@
...
+        string memory name_,
+        string memory symbol_
+    ) {
+        _underlying = underlying_;
+
+        (bool success, uint8 result) = _tryGetAssetDecimals(underlying_);
+        _decimals = success ? result : _DEFAULT_UNDERLYING_DECIMALS;
+
+        _name = name_;
+        _symbol = symbol_;
```
jalextowle, line 37: We have three code blocks and zero comments here. Let's make sure to preserve our style.

contracts/src/everlong/EverlongERC4626.sol
```diff
@@ -0,0 +1,96 @@
+// SPDX-License-Identifier: Apache-2.0
+pragma solidity 0.8.20;
+
+// import { IEverlongERC4626 } from "../interfaces/IEverlongERC4626.sol";
+import { ERC4626 } from "solady/tokens/ERC4626.sol";
+
+contract EverlongERC4626 is ERC4626 {
+    /// @notice Whether virtual shares will be used to mitigate the inflation attack.
+    bool public constant useVirtualShares = true;
```
jalextowle, line 9: I think this comment could be more concrete since we ARE using virtual shares.

```suggestion
    /// @notice Everlong uses virtual shares to mitigate inflation attacks
    bool public constant useVirtualShares = true;
```

contracts/src/everlong/EverlongERC4626.sol
```diff
@@ -0,0 +1,96 @@
...
+
+    /// @notice Used to reduce the feasibility of an inflation attack.
+    uint8 public constant decimalsOffset = 3;
```
jalextowle, line 12: Some justification for why it's 3 would be good. I understand that we'll need to test this to find the real value, so you can just leave a TODO.

jalextowle (commented):
It's looking good. This is the perfect PR size btw

contracts/src/everlong/EverlongPositions.sol
mcclurejt: One case that seems like it would greatly benefit from a double-sided queue is when multiple longs are opened at different times in the same checkpoint interval. 

Without a pointer to the last item, the nieve approach for handling this would be to append a new position to the queue (im sure there's a better way but don't know it yet)... however this comes at the cost of more iterations when rebalancing is performed.

As a bonus,  a double-ended queue would be very handy to have in place should we choose to start targeting an avg. time to maturity.

Could totally be missing something, but it seems like the storage variables required for a single ended queue are identical to those used for a double ended queue.

[**Single-Ended Queue**](https://ethereum.stackexchange.com/questions/129668/how-to-efficiently-implement-a-fifo-array-queue-in-solidity)
```solidity
pragma solidity >=0.8.4;
contract Queue {
    mapping (uint256 => uint256 /* or any other type */ ) queue;
    uint256 first = 1;
    uint256 last = 1;
    function enqueue(uint256 data) public {
        last += 1;
        queue[uint256] = data;
    }

    function dequeue() public returns (uint256) {
        uint256 data;
        require(last > first);
        data = queue[first];
        delete queue[first];
        first += 1;
        return data
    }
    function length() public view returns (uint256) {
        return last - first;
    }
}
```

[**Double-Ended Queue**](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/utils/structs/DoubleEndedQueue.sol#L29-L33)
```solidity
    struct Bytes32Deque {
        uint128 _begin;
        uint128 _end;
        mapping(uint128 index => bytes32) _data;
    }
    function pushBack(Bytes32Deque storage deque, bytes32 value) internal {
        unchecked {
            uint128 backIndex = deque._end;
            if (backIndex + 1 == deque._begin) Panic.panic(Panic.RESOURCE_ERROR);
            deque._data[backIndex] = value;
            deque._end = backIndex + 1;
        }
    }
    function popFront(Bytes32Deque storage deque) internal returns (bytes32 value) {
        unchecked {
            uint128 frontIndex = deque._begin;
            if (frontIndex == deque._end) Panic.panic(Panic.EMPTY_ARRAY_POP);
            value = deque._data[frontIndex];
            delete deque._data[frontIndex];
            deque._begin = frontIndex + 1;
        }
    }    
    function length(Bytes32Deque storage deque) internal view returns (uint256) {
        unchecked {
            return uint256(deque._end - deque._begin);
        }
    }
```
 
If the difference is only in implementing the methods for the double-ended queue, then why not use it?  Especially since OZ seems to have done a great job writing it for us 🤓 

FWIW I'm totally on board with the simplicity of the single sided approach and was initially going to go that route, just seemed  to me like the double-sided might come with some benefits at little-to-no cost.

contracts/src/everlong/EverlongERC4626.sol
mcclurejt: Totally chose this at random, will mark as TODO

contracts/src/everlong/EverlongERC4626.sol
mcclurejt: Gave this more thought, and totally agree. It seems like the `_underlying` variable should actually be the `baseToken` or `vaultSharesToken` for the underlying Hyperdrive instance since it will be used as the token for deposits/mints, idle liquidity, and bond purchases.

contracts/src/interfaces/IEverlongPositions.sol
mcclurejt: I'm worried that the "bond" wording might be a bit generic and get confusing when the interaction logic for opening/closing longs gets included (seeming like this will be meaty enough to be worth encapsulating in its own contract). 

Was thinking this interface + associated contract would cover higher level and aggregate actions and information on Everlong's bond positions.

Maybe something like `IEverlongFundManagement` or `IEverlongLiquidityManagement` would be appropriate and a bit more specific?

jalextowle: If we're worried about it being generic, "Fund," "Liquidity," and "Portfolio" are all more generic/less concrete than "Bond" in this context. I hear what you're saying though since the word "Long" is different than "Bond". Let's just stick with "Position"

contracts/src/interfaces/IEverlongPositions.sol
```diff
@@ -0,0 +1,49 @@
+// SPDX-License-Identifier: Apache-2.0
+pragma solidity 0.8.20;
+
+interface IEverlongPositions {
+    /// @dev Tracks the total amount of bonds managed by Everlong
+    ///      with the same maturityTime.
+    ///      TODO: Update this comment if multiple positions are needed to obtain
+    ///            the total amount of bonds managed by Everlong for a maturity time.
```
jalextowle, line 8: This won't be needed. Under the hood, Hyperdrive positions are grouped together by maturity time. There is no reason why we shouldn't do that here. It's cheaper from a storage perspective, and it reduces DOS vectors.

contracts/src/everlong/EverlongERC4626.sol
jalextowle: That makes sense to me too. The big thing I'm unsure about is whether or not we can really support ERC4626 fully due to withdrawal shares, but we can circle back on that.

contracts/src/everlong/Everlong.sol
```diff
@@ -0,0 +1,71 @@
...
+///
+/// @author DELV
+/// @title Everlong
+/// @notice A money market powered by Hyperdrive.
+/// @custom:disclaimer The language used in this code is for coding convenience
+///                    only, and is not intended to, and does not, have any
+///                    particular legal or regulatory significance.
+contract Everlong is EverlongAdmin, EverlongERC4626, EverlongPositions {
+    constructor(
+        address underlying_,
```
jalextowle, line 67: Nit: Name is the first argument for most of our other contracts (should be all but the factory for historical reasons). Can we put name and symbol first?

contracts/src/everlong/EverlongAdmin.sol
```diff
@@ -0,0 +1,29 @@
...
+
+/// @author DELV
+/// @title EverlongAdmin
+/// @notice Permissioning for Everlong.
+/// @custom:disclaimer The language used in this code is for coding convenience
+///                    only, and is not intended to, and does not, have any
+///                    particular legal or regulatory significance.
+contract EverlongAdmin is IEverlongAdmin {
+    /// @inheritdoc IEverlongAdmin
+    address public admin;
```
jalextowle, line 14: This is fine for now, but it might be easier long term to have an EverlongStorage contract. Just want to put the idea on your radar

contracts/src/everlong/EverlongERC4626.sol
```diff
@@ -0,0 +1,125 @@
...
+    address internal immutable _baseAsset;
+
+    /// @notice Decimals used by the {_baseAsset}.
+    uint8 internal immutable _decimals;
+
+    /// @notice Name of the Everlong token.
+    string internal _name;
+
+    /// @notice Kind of the Everlong contract.
+    string internal constant _kind = "Everlong";
```
jalextowle, line 39: Nit: Can you add this to `contracts/src/libraries/Constants.sol`?

jalextowle (approved):
LGTM. Just a few nits

PR #1097, opened 2024-07-16
Everlong testing harness

mcclurejt:
# Resolved Issues

- closes https://github.com/delvtech/hyperdrive/issues/1083

# Description

Includes a base test harness for Everlong that extends the existing `HyperdriveTest.sol` harness.

# Review Checklists

## @jalextowle 

- [ ] **Tokens**
              token spend?
- [ ] **Reentrancy**
- [ ] **Gas Optimizations**
          memory variables?
          issues?
- [ ] **Visibility**
- [ ] **Math**
          catch underflows?
- [ ] **Testing**

## @jrhea  

- [ ] **Tokens**
              token spend?
- [ ] **Reentrancy**
- [ ] **Gas Optimizations**
          memory variables?
          issues?
- [ ] **Visibility**
- [ ] **Math**
          catch underflows?
- [ ] **Testing**


contracts/src/everlong/EverlongAdmin.sol
```diff
@@ -1,12 +1,16 @@
 // SPDX-License-Identifier: Apache-2.0
 pragma solidity 0.8.20;
 
-import { IEverlongAdmin } from "../interfaces/IEverlongAdmin.sol";
+import { IEverlongAdmin } from "contracts/src/interfaces/IEverlongAdmin.sol";
```
jalextowle, line 4: Why are we making these absolute? Everywhere else in the project, we use relative for contracts in `contracts/src` and absolute elsewhere. The reason why we do this is annoying and historical, but I don't want to change unless we change all of the files.

test/utils/EverlongTest.sol
```diff
@@ -0,0 +1,41 @@
...
+import { HyperdriveTest } from "test/utils/HyperdriveTest.sol";
+import { IEverlong } from "contracts/src/interfaces/IEverlong.sol";
+import { Everlong } from "contracts/src/everlong/Everlong.sol";
+
+contract EverlongTest is HyperdriveTest {
+    IEverlong internal everlong;
+
+    function setUp() public virtual override {
+        super.setUp();
+        vm.startPrank(alice);
```
jalextowle, line 15: What's the point of this? `deployEverlong` immediately switches to the `deployer`.

test/utils/EverlongTest.sol
```diff
@@ -0,0 +1,41 @@
...
+pragma solidity 0.8.20;
+
+import { VmSafe } from "forge-std/Vm.sol";
+import { BaseTest } from "test/utils/BaseTest.sol";
+import { HyperdriveTest } from "test/utils/HyperdriveTest.sol";
+import { IEverlong } from "contracts/src/interfaces/IEverlong.sol";
+import { Everlong } from "contracts/src/everlong/Everlong.sol";
+
+contract EverlongTest is HyperdriveTest {
+    IEverlong internal everlong;
```
jalextowle, line 11: I love this. Using interfaces from the start will make our lives much easier.

jalextowle (commented):
It's looking good. Just some style nits.

contracts/src/everlong/EverlongAdmin.sol
mcclurejt: Got it, didn't make the connection on where rel/abs paths were being used, will put it back to relative

contracts/src/everlong/EverlongERC4626.sol
```diff
@@ -1,7 +1,7 @@
 // SPDX-License-Identifier: Apache-2.0
 pragma solidity 0.8.20;
 
-// import { IEverlongERC4626 } from "../interfaces/IEverlongERC4626.sol";
+import { IHyperdrive } from "contracts/src/interfaces/IHyperdrive.sol";
```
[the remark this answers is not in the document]
mcclurejt: https://github.com/delvtech/hyperdrive/pull/1097/files#diff-90f83f6ad18083e73ae495e1be72e15412314e5db271c2ec8121904305a82fbdL54

PR #1111, opened 2024-07-22
Tests for the SUSDe Integration

jalextowle:
# Resolved Issues

Fixes: #1105.

# Description

To integrate with Ethena's SUSDe instance, we will be using the `ERC4626Hyperdrive` instance contract. To sanity check this idea, this PR adds a comprehensive test suite that can also be used for future ERC4626 integrations.

# Review Checklists

be removed.

## [[Reviewer Name]]

- [ ] **Tokens**
              token spend?
- [ ] **Reentrancy**
- [ ] **Gas Optimizations**
          memory variables?
          issues?
- [ ] **Visibility**
- [ ] **Math**
          catch underflows?
- [ ] **Testing**

### DIFF
--- a/test/instances/susde/SUSDe.t.sol
+++ b/test/instances/susde/SUSDe.t.sol
@@ -1,9 +1,6 @@
 // SPDX-License-Identifier: Apache-2.0
 pragma solidity 0.8.20;
 
-// FIXME
-import { console2 as console } from "forge-std/console2.sol";
-
 import { stdStorage, StdStorage } from "forge-std/Test.sol";
 import { ERC4626HyperdriveCoreDeployer } from "contracts/src/deployers/erc4626/ERC4626HyperdriveCoreDeployer.sol";
 import { ERC4626HyperdriveDeployerCoordinator } from "contracts/src/deployers/erc4626/ERC4626HyperdriveDeployerCoordinator.sol";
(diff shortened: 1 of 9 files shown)


test/instances/susde/SUSDe.t.sol
```diff
@@ -0,0 +1,736 @@
...
+            // NOTE: SUSDe currently has a cooldown on withdrawals which
+            // prevents users from withdrawing as base instantaneously. We still
+            // support withdrawing with base since the cooldown can be disabled
+            // in the future.
+            baseWithdrawError: abi.encodeWithSelector(
+                OperationNotAllowed.selector
+            )
+        });
+
+    /// @dev Instantiates the Instance testing suite with the configuration.
```
jrhea, line 78: weird capitalization in this comment

test/instances/susde/SUSDe.t.sol
```diff
@@ -0,0 +1,736 @@
...
+    }
+
+    function test_round_trip_long_maturity() external {
+        // Bob opens a long with base.
+        uint256 basePaid = hyperdrive.calculateMaxLong();
+        IERC20(hyperdrive.baseToken()).approve(address(hyperdrive), basePaid);
+        (uint256 maturityTime, uint256 longAmount) = openLong(bob, basePaid);
+
+        // Advance the time and accrue a large amount of interest.
+        advanceTime(POSITION_DURATION, 137.123423e18);
```
jrhea, line 548: is there a specific reason you chose this variable interest rate?

test/utils/InstanceTest.sol
```diff
@@ -42,6 +42,8 @@ abstract contract InstanceTest is HyperdriveTest {
         bool enableShareDeposits;
         bool enableBaseWithdraws;
         bool enableShareWithdraws;
+        // TODO: Add variants of this for the other "enable" states.
```
jrhea, line 45: does this TODO need to be addressed?

jrhea (commented):
nice work on this! just some questions possible nits.

jrhea (approved):

test/utils/InstanceTest.sol
jalextowle: No, I left this so that we think about it in the future. That's why I left it as a TODO FWIW. Anytime there is something that I want to do in this PR, I make it a `FIXME`.

test/instances/susde/SUSDe.t.sol
jalextowle: No. I used this so that we were testing with a weird and large number. That said, I'll switch this to be a fuzz test since this is arbitrary, and it would be better to have a bound that we can actually take to the bank.

test/instances/susde/SUSDe.t.sol
jalextowle: Agreed

mcclurejt (approved):

PR #1113, opened 2024-07-23
morpho blue sepolia config + generic library resolution/deployment

mcclurejt:
# Resolved Issues

- Uses `MorphoBlue` coordinator + instance to deploy integration on Sepolia
- Generic Library resolution in deployment functions (rather than specific configuration for each library)

# Description

# Review Checklists

be removed.

## [[Reviewer Name]]

- [ ] **Tokens**
              token spend?
- [ ] **Reentrancy**
- [ ] **Gas Optimizations**
          memory variables?
          issues?
- [ ] **Visibility**
- [ ] **Math**
          catch underflows?
- [ ] **Testing**


hardhat.config.ts
```diff
@@ -80,6 +83,7 @@ const config: HardhatUserConfig = {
                     SEPOLIA_RETH_30DAY,
                     SEPOLIA_STETH_14DAY,
                     SEPOLIA_STETH_30DAY,
+                    SEPOLIA_MORPHO_BLUE_DAI_14DAY,
```
jalextowle, line 86: Why do we need to update this config considering that we have a specific Sepolia config?

scripts/deploy.sh
```diff
@@ -20,5 +22,9 @@ if [[ "${NETWORK}" != "anvil" && "${NETWORK}" != "hardhat" && "${NETWORK}" != "m
 	fi
 fi
 
-npx hardhat deploy:hyperdrive --show-stack-traces --network ${NETWORK} --config hardhat.config.${NETWORK}.ts
-npx hardhat deploy:verify --show-stack-traces --network ${NETWORK} --config hardhat.config.${NETWORK}.ts
+config_filename="hardhat.config.${NETWORK}.ts"
+if [[ "${NETWORK}" == "hardhat" ]]; then
+	config_filename="hardhat.config.ts"
```
jalextowle, line 27: Is this the reason why? Why are we doing this?

tasks/deploy/lib/environment-extensions.ts
```diff
@@ -589,13 +546,10 @@ extendEnvironment((hre) => {
...
             args as any,
-            {
-                gas: 5_000_000n,
-                value,
-            },
+            { gas: 10_000_000n, value },
         );
         let tx = await factory.write.deployAndInitialize(args as any, {
-            gas: 5_000_000n,
+            gas: 10_000_000n,
```
jalextowle, line 552: This is a big increase. Is there a tighter bound that we can use? I'm concerned that this will result in failures if the deployer doesn't have tons of ETH (but the tx would actually still succeed) on a mainnet deployment.

tasks/deploy/lib/environment-extensions.ts
```diff
@@ -589,13 +546,10 @@ extendEnvironment((hre) => {
         }
         let { result: address } = await factory.simulate.deployAndInitialize(
             args as any,
-            {
-                gas: 5_000_000n,
-                value,
-            },
+            { gas: 10_000_000n, value },
```
jalextowle, line 549: Same here

tasks/deploy/lib/environment-extensions.ts
```diff
@@ -530,13 +490,10 @@ extendEnvironment((hre) => {
...
                 args as any,
-                {
-                    gas: 5_500_000n,
-                    ...options.viemConfig,
-                },
+                { gas: 10_000_000n, ...options.viemConfig },
             );
             let tx = await factory.write.deployTarget(args as any, {
-                gas: 5_500_000n,
+                gas: 10_000_000n,
```
jalextowle, line 496: Same here

tasks/deploy/lib/environment-extensions.ts
```diff
@@ -530,13 +490,10 @@ extendEnvironment((hre) => {
             ];
             let { result: address } = await factory.simulate.deployTarget(
                 args as any,
-                {
-                    gas: 5_500_000n,
-                    ...options.viemConfig,
-                },
+                { gas: 10_000_000n, ...options.viemConfig },
```
jalextowle, line 493: Same here

tasks/deploy/lib/environment-extensions.ts
```diff
@@ -132,12 +132,29 @@ extendEnvironment((hre) => {
...
+                ).address;
+                libraries[libraryName] = libraryAddress;
+            }
+        }
+
         const instance = await hre.viem.deployContract(
             contract as ContractName<typeof contract>,
             args as any,
-            { ...viemConfig, gas: 5_000_000n },
+            { ...viemConfig, libraries, gas: 10_000_000n },
```
jalextowle, line 157: Same here

tasks/deploy/lib/environment-extensions.ts
jalextowle: I'd be surprised if Morpho really caused this big of a jump (especially considering that it's using an external library that we should have already deployed by this point).

tasks/deploy/lib/environment-extensions.ts
```diff
@@ -132,12 +132,29 @@ extendEnvironment((hre) => {
...
+
         console.log(`deploying ${name}...`);
 
+        // Deploy libraries if necessary and link them to the target bytecode.
+        let artifact = await hre.artifacts.readArtifact(contract);
+        let libraries: Record<string, `0x${string}`> = {};
+        for (let libraryFilename in artifact.linkReferences) {
+            for (let libraryName in artifact.linkReferences[
+                libraryFilename as keyof typeof artifact.linkReferences
+            ] as any) {
```
jalextowle, line 144: This code is a bit confusing to me. Why do we need to loop over this? Are there going to be multiple library names per library filename? Maybe I'm missing something.

hardhat.config.ts
mcclurejt: I was using it to deploy everything on a forked hh network before going to sepolia (faster iteration times). Also want to have some base config so `npx hardhat` can be run in the root of the repo to show all commands. Happy to remove these bits or document somewhere if desired

tasks/deploy/config/sepolia/morpho-blue-dai-14day.ts
```diff
@@ -0,0 +1,140 @@
...
+        {
+            morpho: "0x927A9E3C4B897eF5135e6B2C7689637fA8E2E0Bd" as `0x${string}`,
+            collateralToken:
+                "0xFF8AFe6bb92eB9D8e80c607bbe5bbb78BF1201Df" as `0x${string}`, // Hyperdrive Sepolia sDAI
+            oracle: "0x23F3A48121861b78f66BbE3DF60AD24A21c4DDad" as `0x${string}`,
+            irm: "0x0fB591F09ab2eB967c0EFB9eE0EF6642c2abe6Ab" as `0x${string}`,
+            lltv: BigInt("980000000000000000"),
+        },
+    ],
+);
```
jalextowle, line 59: I'm surprised that we need to be this verbose here. Is there not a way to do this with something in the Typescript contract wrapper or artifact? It's fine if this is the only way that we can do it, I just would have expected there to be a struct type that we could encode.

tasks/deploy/lib/environment-extensions.ts
mcclurejt: That's just how it's structured, unsure of a case where there would be multiple but figured looping again was the safe approach.

Here's an example from artifacts:

```json
"linkReferences": {
    "contracts/src/instances/morpho-blue/MorphoBlueConversions.sol": {
      "MorphoBlueConversions": [
        {
          "length": 20,
          "start": 16159
        },
        {
          "length": 20,
          "start": 16562
        }
      ]
    },
    "contracts/src/libraries/LPMath.sol": {
      "LPMath": [
        {
          "length": 20,
          "start": 7689
        }
      ]
    }
  }
  ```

File: tasks/deploy/config/sepolia/morpho-blue-dai-14day.ts [hunk 10]
```diff
@@ -0,0 +1,140 @@
...
+        coordinatorAddress: async (hre) =>
+            hre.hyperdriveDeploy.deployments.byName(
+                SEPOLIA_MORPHO_BLUE_COORDINATOR_NAME,
+            ).address,
+        deploymentId: toBytes32(SEPOLIA_MORPHO_BLUE_DAI_14DAY_NAME),
+        salt: toBytes32("0x420"),
+        extraData: enc,
+        contribution: CONTRIBUTION,
+        fixedAPR: parseEther("0.10"),
+        timestretchAPR: parseEther("0.10"),
```
jalextowle, line 74: Nit: We don't really need the ending `0`

tasks/deploy/config/sepolia/morpho-blue-dai-14day.ts
```diff
@@ -0,0 +1,140 @@
...
+            // mint some tokens for the contribution
+            tx = await baseToken.write.mint([CONTRIBUTION]);
+            await pc.waitForTransactionReceipt({ hash: tx });
+        },
+        poolDeployConfig: async (hre) => {
+            return {
+                baseToken:
+                    hre.hyperdriveDeploy.deployments.byName("DAI").address,
+                vaultSharesToken: zeroAddress,
+                circuitBreakerDelta: parseEther("0.6"),
```
jalextowle, line 114: This circuit breaker delta is too high. For mainnet pools, we're using the lower of `timeStretchAPR` and `10%`.

tasks/deploy/config/sepolia/morpho-blue-dai-14day.ts
```diff
@@ -0,0 +1,140 @@
...
+                ).address,
+                ...(await getLinkerDetails(
+                    hre,
+                    hre.hyperdriveDeploy.deployments.byName(
+                        SEPOLIA_FACTORY_NAME,
+                    ).address,
+                )),
+                fees: {
+                    curve: parseEther("0.01"),
+                    flat: normalizeFee(parseEther("0.0005"), "14 days"),
```
jalextowle, line 134: Might be good to make a constant for the position duration since we use `"14 days"` twice.

jalextowle (commented):
Aside from the other comments that I left, we'll want to bump the version in `contracts/src/libraries/Constants.sol` so that we can ship a new release once this lands. Any chance you could work on the mainnet config so that I can deploy it after I review and get up in the morning?

tasks/deploy/lib/environment-extensions.ts
mcclurejt: IIRC it needed ~6.5mil, wdyt about updating to 7.5mil?

scripts/deploy.sh
mcclurejt: We want to have the file `hardhat.config.ts` so that it plays nicely with editors trying to understand project structure as well as having the ability to run `npx hardhat` and see what commands are available.

Unfortunately, for the network `hardhat`, the resolved config would be `hardhat.config.hardhat.ts`, which is why I include this accomodation.

tasks/deploy/config/sepolia/morpho-blue-dai-14day.ts
mcclurejt: Will take a stab at resolving it from arguments like [here](https://github.com/delvtech/hyperdrive/blob/main/tasks/deploy/lib/types.ts#L210)

tasks/deploy/lib/environment-extensions.ts
jalextowle: Ah I see. Makes sense since it's a `[string]: Object` type. Carry on 🫡

tasks/deploy/lib/environment-extensions.ts
jalextowle: Could we do the closest bound that will work? These deployments should be entirely deterministic in cost, so I don't think it pays to add padding. Let me know if I'm missing something.

jalextowle: FWIW, I've been testing the deployments on a mainnet fork for sanity, so I'd see anything like this pop up there if the bound was too tight for some reason. Maybe 6.75 million would be a good compromise between exactly right and having a bit of padding?

scripts/deploy.sh
jalextowle: I think `hardhat.config.hardhat.ts` is fine if it works. It feels pretty consistent with the other config names, but I see why you might not like it.

hardhat.config.ts
jalextowle: Gotcha, I just thought it was weird that we had sepolia here rather than mainnet. What I've been doing for testing is using the same network (in my case, it's usually been `NETWORK=mainnet`), but I'll point the RPC URL to a node that is a mainnet fork. Could you do something like that for testing? That may obviate the need for this.

scripts/deploy.sh
mcclurejt: We need the filename `hardhat.config.ts` for `npx hardhat` to run without having to point it to a config file... also pretty sure the hardhat language server vscode integration depends on it being present for some features

hardhat.config.ts
mcclurejt: Been using it as a scratch file more or less so happy to remove the hyperdrive config from the file, however we still want this file to exist for reasons mentioned in other comments

scripts/deploy.sh
jalextowle: Right, but we don't need the sepolia stuff in the file, right?

hardhat.config.mainnet.ts
```diff
@@ -28,8 +30,13 @@ const config: HardhatUserConfig = {
                 coordinators: [
                     MAINNET_EZETH_COORDINATOR,
                     MAINNET_RETH_COORDINATOR,
+                    MAINNET_MORPHO_BLUE_COORDINATOR,
+                ],
+                instances: [
+                    MAINNET_EZETH_182DAY,
+                    MAINNET_RETH_182DAY,
+                    MAINNET_MORPHO_BLUE_DAI_14DAY,
```
jalextowle, line 38: Should be 182 days

tasks/deploy/config/mainnet/morpho-blue-susde-dai-182day.ts
```diff
@@ -0,0 +1,126 @@
...
+    SIX_MONTHS,
+    getLinkerDetails,
+    normalizeFee,
+    parseDuration,
+    toBytes32,
+} from "../../lib";
+import { MAINNET_FACTORY_NAME } from "./factory";
+import { MAINNET_MORPHO_BLUE_COORDINATOR_NAME } from "./morpho-blue-coordinator";
+
+export const MAINNET_MORPHO_BLUE_DAI_14DAY_NAME =
```
jalextowle, line 14: Should be 182 days

tasks/deploy/config/mainnet/morpho-blue-susde-dai-182day.ts
```diff
@@ -0,0 +1,126 @@
...
+    "ElementDAO 182 Day sUSDe/DAI Hyperdrive";
+
+const CONTRIBUTION = parseEther("100");
```
jalextowle, line 17: This is a large initial contribution

tasks/deploy/config/mainnet/morpho-blue-susde-dai-182day.ts
```diff
@@ -0,0 +1,126 @@
...
+                MAINNET_MORPHO_BLUE_COORDINATOR_NAME,
+            ).address,
+        deploymentId: toBytes32(
+            MAINNET_MORPHO_BLUE_DAI_14DAY_NAME.slice(0, 32),
+        ),
+        salt: toBytes32("0x420"),
+        extraData: morphoBlueParameters,
+        contribution: CONTRIBUTION,
+        fixedAPR: parseEther("0.083"),
+        timestretchAPR: parseEther("0.05"),
```
jalextowle, line 75: TODO: Double-check this. This seems potentially low, but it might be right for this market.

tasks/deploy/config/mainnet/morpho-blue-susde-dai-182day.ts
```diff
@@ -0,0 +1,126 @@
...
+                ).address,
+                CONTRIBUTION,
+            ]);
+            await pc.waitForTransactionReceipt({ hash: tx });
+        },
+        poolDeployConfig: async (hre) => {
+            return {
+                baseToken: MAINNET_DAI_ADDRESS,
+                vaultSharesToken: zeroAddress,
+                circuitBreakerDelta: parseEther("0.05"),
```
jalextowle, line 100: TODO: Double-check this. If we raise the timestretch, we should  also increase this.

tasks/deploy/config/mainnet/morpho-blue-susde-dai-182day.ts
```diff
@@ -0,0 +1,126 @@
...
+                minimumShareReserves: parseEther("10"),
```
jalextowle, line 101: We can use `1e15` for this too now.

hardhat.config.mainnet.ts
```diff
@@ -10,6 +10,8 @@ import {
     MAINNET_EZETH_182DAY,
     MAINNET_EZETH_COORDINATOR,
     MAINNET_FACTORY,
+    MAINNET_MORPHO_BLUE_COORDINATOR,
+    MAINNET_MORPHO_BLUE_DAI_14DAY,
```
jalextowle, line 14: Should be 182 days

hardhat.config.ts
jalextowle: Yeah for sure. I was only referring to the inclusion of the sepolia yield sources FWIW. I know that we need `hardhat.config.ts`

tasks/deploy/config/mainnet/morpho-blue-susde-dai-182day.ts
jalextowle: We increased this to 10% so that we can support rates up to ~40% with uniform liquidity and rates over 100% on the very high end.

tasks/deploy/config/mainnet/morpho-blue-susde-dai-182day.ts
jalextowle: We increased this to 10% so that we can support rates up to ~40% with uniform liquidity and rates over 100% on the very high end.

tasks/deploy/config/mainnet/morpho-blue-susde-dai-182day.ts
jalextowle: I changed this to `1e15`

tasks/deploy/config/mainnet/morpho-blue-susde-dai-182day.ts
jalextowle: Lowered this to 1 DAI

PR #1114, opened 2024-07-25
Updated the instance tests to support fuzzing of new integrations

jalextowle:
# Resolved Issues

Partially resolves: #1108.

# Description

This PR makes several functions in the `InstanceTest` harness public. This makes it possible to leverage this functionality on a mainnet fork to run fuzz tests against new integrations. This is the PR that should bring this system to the MCP (minimum crappy product). Some notable features that are absent are:

- The ability to fund arbitrary accounts. For now, these integrations will rely upon whale addresses, which also means that the fork block is important. 
- Advance time will increase the time internally as well as accrue interest. This PR doesn't add a separate `accrue` function, although that will not be hard to add in a later PR.

### DIFF
--- a/test/instances/aave/AaveHyperdrive.t.sol
+++ b/test/instances/aave/AaveHyperdrive.t.sol
@@ -48,27 +48,27 @@ contract AaveHyperdriveTest is InstanceTest {
     address[] internal baseTokenWhaleAccounts = [WETH_WHALE];
     address[] internal vaultSharesTokenWhaleAccounts = [AWETH_WHALE];
 
-    // The configuration for the instance testing suite.
-    InstanceTestConfig internal __testConfig =
-        InstanceTestConfig({
-            name: "Hyperdrive",
-            kind: "AaveHyperdrive",
-            baseTokenWhaleAccounts: baseTokenWhaleAccounts,
-            vaultSharesTokenWhaleAccounts: vaultSharesTokenWhaleAccounts,
-            baseToken: WETH,
-            vaultSharesToken: IERC20(address(AWETH)),
-            shareTolerance: 1e3,
-            minTransactionAmount: 1e15,
-            positionDuration: POSITION_DURATION,
-            enableBaseDeposits: true,
-            enableShareDeposits: true,
-            enableBaseWithdraws: true,
-            enableShareWithdraws: true,
-            baseWithdrawError: new bytes(0)
-        });
-
     /// @dev Instantiates the instance testing suite with the configuration.
-    constructor() InstanceTest(__testConfig) {}
+    constructor()
+        InstanceTest(
+            InstanceTestConfig({
+                name: "Hyperdrive",
+                kind: "AaveHyperdrive",
+                baseTokenWhaleAccounts: baseTokenWhaleAccounts,
+                vaultSharesTokenWhaleAccounts: vaultSharesTokenWhaleAccounts,
+                baseToken: WETH,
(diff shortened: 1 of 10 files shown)


test/instances/aave/AaveHyperdrive.t.sol
```diff
@@ -48,27 +48,27 @@ contract AaveHyperdriveTest is InstanceTest {
...
+                baseToken: WETH,
+                vaultSharesToken: IERC20(address(AWETH)),
+                shareTolerance: 1e3,
+                minTransactionAmount: 1e15,
+                positionDuration: POSITION_DURATION,
+                enableBaseDeposits: true,
+                enableShareDeposits: true,
+                enableBaseWithdraws: true,
+                enableShareWithdraws: true,
+                baseWithdrawError: new bytes(0)
```
slundqui, line 68: What does this parameter do?

slundqui (approved):
Thanks for putting this together! Everything here looks relatively straightforward to use

test/instances/aave/AaveHyperdrive.t.sol
jalextowle: This is used to support custom errors that might be thrown on base withdrawals. Right now, this is used with ethena because withdraws have to be processed via a cooldown period

test/utils/HyperdriveTest.sol
```diff
@@ -1,6 +1,8 @@
 // SPDX-License-Identifier: Apache-2.0
 pragma solidity 0.8.20;
 
+import { console2 as console } from "forge-std/console2.sol";
+
```
mcclurejt, line 5:
```suggestion
```

mcclurejt (approved):

test/instances/aave/AaveHyperdrive.t.sol
```diff
@@ -80,14 +80,14 @@ contract AaveHyperdriveTest is InstanceTest {
 
     /// @dev Gets the extra data used to deploy Hyperdrive instances.
     /// @return The extra data.
-    function getExtraData() internal pure override returns (bytes memory) {
+    function getExtraData() public pure override returns (bytes memory) {
         return new bytes(0);
     }
 
     /// @dev Converts base amount to the equivalent about in shares.
     function convertToShares(
```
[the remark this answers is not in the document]
jalextowle: Good question. That method isn’t available before deployHyperdrive is called, and this is used within deployHyperdrive. We might be able to find a workaround in the future, but for now, we need this.

test/instances/aave/AaveHyperdrive.t.sol
```diff
@@ -110,12 +110,12 @@ contract AaveHyperdriveTest is InstanceTest {
     /// @param _factory The address of the Hyperdrive factory.
     function deployCoordinator(
         address _factory
-    ) internal override returns (address) {
+    ) public override returns (address) {
         vm.startPrank(alice);
```
slundqui, line 114: I seem to be having trouble calling this contract's functions. My best guess is that I'm not using the `alice` account correctly from this function. Any chance we can either (1) expose the private key of `alice`, or (2) allow the account as an argument?

slundqui, line 114: I'm attempting to deploy this contract, then call this function to deploy the coordinator, but getting execution reverted on this line. Are cheatcodes supported when the contract is deployed?

jalextowle:
Closing this since we found that this wasn't a viable path to fuzzing the instances.

PR #1115, opened 2024-07-25
Added a getter for `totalShares`

jalextowle:
# Description

This PR adds a getter for `totalShares` that serves as a convenient way for the fuzzing tools (and other integrations) to get the amount of vault shares owned by Hyperdrive for each yield source.


contracts/src/external/HyperdriveTarget0.sol
```diff
@@ -442,6 +448,15 @@ abstract contract HyperdriveTarget0 is
         _revert(abi.encode(_convertToShares(_baseAmount)));
     }
 
+    /// @notice Gets the total amount of vault shares held by Hyperdrive.
+    /// @dev This is a convenience method that allows developers to get the
+    ///      total amount of vault shares without knowing the specifics of the
+    ///      integration.
+    /// @return The total amount of vault shares held by Hyperdrive.
+    function totalShares() external view returns (uint256) {
```
mcclurejt, line 456: out of curiousity, why isn't there any accompanying `_totalBase()`/`totalBase()` ?

mcclurejt (approved):

contracts/src/external/HyperdriveTarget0.sol
jrhea, line 456: can we get a simple test for this?

jrhea (approved):
lgtm. just one nit about a test

contracts/src/external/HyperdriveTarget0.sol
jrhea, line 456: should we update codegen with total shares?

jalextowle: We don’t need it for sweep (which is what this is used for), and it should always be 0 since 100% of the funds are allocated to the yield source

jalextowle: Yeah, for sure

PR #1116, opened 2024-07-25
Minimum Crappy Style Guide

jalextowle:
# Description

This PR adds the Minimum Crappy Product of a DELV Solidity Style Guide. This includes the basic elements of our style. It is likely missing many important style rules or heuristics. As we go through future PR review cycles and find things that aren't in the style guide, we should add them along with fixing whatever review feedback arises.

### DIFF
--- a/STYLE_GUIDE.md
+++ b/STYLE_GUIDE.md
@@ -1,0 +1,25 @@
+# DELV Solidity Style Guide
+
+## Commenting
+
+1. Use complete sentences and end them with the appropriate puntuaction (typically a period).
+2. Use correct spelling.
+3. Use correct capitalization.
+4. Comments should not go past the 80th column from the start of the line.
+5. Add  full Natspec to each contract, library, function, storage variable, struct, event, and error.
+6. Within function bodies, code should be broken up into logical chunks (hereafter called "comment blocks") that are clearly commented. These comments should logically explain what the block is accomplishing. Any comments interleaved within blocks should be `NOTE` or `TODO` comments.
+7. `FIXME` comments should always be removed before merging your PR.
+8. A `NOTE` comment should be added above any line that uses fixed point math or another function that rounds up or down to explain why the code rounds in a particular direction.
+9. Integrations should include detailed comments explaining what is different or special about the implementation. This includes things like how interest accrues, why base or vault shares may not be a supported deposit or withdrawal asset, how the `convertToBase` or `convertToShares` functions work, etc.
+
+## Imports
+
+1. Relative paths are preferred because they improve lookups for LSP providers.
+2. Imports must specify what they are importing.
+3. All imports must be used.
+
+## Interfaces
+
+1. Each immutable should have a corresponding getter.
+2. Each state variable should have a corresponding getter or should be reachable through a generalized getter (like `loads`).
+3. Each function should have an event that encodes the state changes that occurred within the function call. If possible, this event should be sufficient to fully recreate the state change that occurred from the previous state.
--- a/STYLE_GUIDE.md
+++ b/STYLE_GUIDE.md
@@ -2,10 +2,10 @@
 
 ## Commenting
 
-1. Use complete sentences and end them with the appropriate punctuaction (typically a period).
+1. Use complete sentences and end them with the appropriate punctuation (typically a period).
 2. Use correct spelling.
 3. Use correct capitalization.
-4. Comments should not go past the 80th column from the start of the line.
+4. Comments should not go past the 80th column from the start of the line. An exception can be made for multi-line math comments in Ascii or Latex.
 5. Add full Natspec to each contract, library, function, storage variable, struct, event, and error.
 6. @notice tags indicate that something is part of the public interface and should be used for giving a synopsis of a contract, public and external functions, public state variables, events and errors.
 7. @dev tags indicate something that developers will be interested in but isn't something that end users need to be concerned with. Each contract's Natspec should include a `@dev` section that explains the contract at a high-level and explains any gotchas that developers should be aware when integrating with or using the contracts.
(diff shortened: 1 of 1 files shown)


sentilesdal (approved):
nice

STYLE_GUIDE.md
```diff
@@ -0,0 +1,25 @@
+# DELV Solidity Style Guide
+
+## Commenting
+
+1. Use complete sentences and end them with the appropriate puntuaction (typically a period).
```
mcclurejt, line 5:
```suggestion
1. Use complete sentences and end them with the appropriate punctuation (typically a period).
```

STYLE_GUIDE.md
```diff
@@ -0,0 +1,25 @@
...
+2. Use correct spelling.
+3. Use correct capitalization.
+4. Comments should not go past the 80th column from the start of the line.
```
mcclurejt, line 8: I think this hard limit may become pretty irritating for longer-form commentary and math equations.

One alternative I like is prettier's [print width](https://prettier.io/docs/en/options.html#print-width) approach:

> In code styleguides, maximum line length rules are often set to 100 or 120. However, when humans write code, they don’t strive to reach the maximum number of columns on every line. Developers often use whitespace to break up long lines for readability. In practice, the average line length often ends up well below the maximum.
> 
> Prettier’s printWidth option does not work the same way. It is not the hard upper allowed line length limit. It is a way to say to Prettier roughly how long you’d like lines to be. Prettier will make both shorter and longer lines, but generally strive to meet the specified printWidth.
> 
> Remember, computers are dumb. You need to explicitly tell them what to do, while humans can make their own (implicit) judgements, for example on when to break a line.

STYLE_GUIDE.md
```diff
@@ -0,0 +1,25 @@
...
+5. Add  full Natspec to each contract, library, function, storage variable, struct, event, and error.
```
mcclurejt, line 9:
```suggestion
5. Add full Natspec to each contract, library, function, storage variable, struct, event, and error.
```

mcclurejt: One way to make this a bit more clear would be to highlight which [Natspec tags](https://docs.soliditylang.org/en/latest/natspec-format.html#tags) we would like to see for each context.

STYLE_GUIDE.md
```diff
@@ -0,0 +1,25 @@
...
+6. Within function bodies, code should be broken up into logical chunks (hereafter called "comment blocks") that are clearly commented. These comments should logically explain what the block is accomplishing. Any comments interleaved within blocks should be `NOTE` or `TODO` comments.
```
mcclurejt, line 10: > Any comments interleaved within blocks should be `NOTE` or `TODO` comments.

I'm a bit confused on what this is trying to convey... it reads to me like all comments within functions should be prepended with `NOTE` or `TODO`, but I feel like that isn't your intention.

STYLE_GUIDE.md
```diff
@@ -0,0 +1,25 @@
...
+7. `FIXME` comments should always be removed before merging your PR.
+8. A `NOTE` comment should be added above any line that uses fixed point math or another function that rounds up or down to explain why the code rounds in a particular direction.
+9. Integrations should include detailed comments explaining what is different or special about the implementation. This includes things like how interest accrues, why base or vault shares may not be a supported deposit or withdrawal asset, how the `convertToBase` or `convertToShares` functions work, etc.
```
mcclurejt, line 13: In a similar vein, all contracts should probably have
- a synopsis of their purpose
- an outline of their functionality

STYLE_GUIDE.md
```diff
@@ -0,0 +1,25 @@
...
+
+1. Relative paths are preferred because they improve lookups for LSP providers.
+2. Imports must specify what they are importing.
+3. All imports must be used.
+
+## Interfaces
+
+1. Each immutable should have a corresponding getter.
+2. Each state variable should have a corresponding getter or should be reachable through a generalized getter (like `loads`).
+3. Each function should have an event that encodes the state changes that occurred within the function call. If possible, this event should be sufficient to fully recreate the state change that occurred from the previous state.
```
mcclurejt, line 25:
```suggestion
3. Each function that changes state should have an event that encodes the state changes that occurred. If possible, this event should be sufficient to fully recreate the state change that occurred from the previous state.
```

mcclurejt (commented):
comments + questions + suggestions

STYLE_GUIDE.md
jrhea: oh good call. for example @notice on public/external and @dev for private/internal

STYLE_GUIDE.md
jrhea: also,
- legal disclaimer 
- license

jrhea (commented):
nice job! just a few additional comments

STYLE_GUIDE.md
jalextowle: We already do this throughout the codebase. I will add an exception for math, but otherwise, we’ve been getting along fine with this.

mcclurejt: Yeah wow after poking around and checking column numbers this is actually upheld super well.... expected there to be way more lines in the 80->90 col range

jalextowle: Yeah, if you do vertical splits on a 16 inch screen, it becomes very nice/helpful to follow this convention. Anything much larger will start to wrap.

STYLE_GUIDE.md
jalextowle: It’s trying to convey that the comment at the start of the block should be the only “naked” comment and that any comments within the block should be NOTE or TODO.

For example, this follows this:

```
// blah blah blah
uint256 a = foo();
a += b - c;
// NOTE: We have to reduce a because…
a = a.mulDown(0.5e18);
```

Whereas this doesn’t:

```
// blah blah blah
uint256 a = foo();
a += b - c;
// We have to reduce a because…
a = a.mulDown(0.5e18);
```

jalextowle: It just helps you visually see the start of the code block by looking for the starting comment. Then it’s obvious that any comments inside the block are at a lower-level of detail.

jalextowle: If this clears it up, I can just include this example. My thought was that we could add examples wherever they were needed, but I wasn’t going to do that for all of the rules because this is the minimum crappy style guide

STYLE_GUIDE.md
jalextowle: What irony 😂

STYLE_GUIDE.md
jalextowle: Oh good call. Yeah, I’ll add some more explicit rules for Natspec

PR #1118, opened 2024-07-26
AAVE L2 Integration

sentilesdal:
Adds the AAVE L2Pool integration.  Instead of using the codegen I followed the Aave mainnet code as the template since they are mostly the same contract.  The L2 version has some functions that can be found here: 
https://github.com/aave/aave-v3-core/blob/master/contracts/protocol/pool/L2Pool.sol


test/instances/aave/AaveL2Hyperdrive.t.sol
```diff
@@ -0,0 +1,461 @@
...
+    /// @dev Converts share amount to the equivalent amount in base.
+    function convertToBase(
+        uint256 shareAmount
+    ) internal view override returns (uint256) {
+        return
+            shareAmount.mulDivDown(
+                POOL.getReserveNormalizedIncome(address(WETH)),
+                1e27
+            );
+    }
```
jalextowle, line 142: It would be good to use the `AaveL2Conversions` library here

test/instances/aave/AaveL2Hyperdrive.t.sol
```diff
@@ -0,0 +1,461 @@
...
+        uint256 normalizedTime = timeDelta.divDown(365 days);
+        reserveNormalizedIncome = variableRate >= 0
+            ? reserveNormalizedIncome +
+                reserveNormalizedIncome.mulDown(uint256(variableRate)).mulDown(
+                    normalizedTime
+                )
+            : reserveNormalizedIncome -
+                reserveNormalizedIncome.mulDown(uint256(-variableRate)).mulDown(
+                    normalizedTime
+                );
```
jalextowle, line 440: We've been switching to using `HyperdriveUtils.calculateInterest` in the other yield source tests. Could you update this to use that function?

jalextowle (commented):
It's looking super solid. Just need to update `supply`, `withdraw`, and a couple of other things and it should be good to go.

contracts/src/instances/aave-l2/AaveL2Base.sol
```diff
@@ -0,0 +1,240 @@
...
+        );
+        amountWithdrawn = _vault.withdraw(withdrawParams);
+        _baseToken.transfer(_destination, amountWithdrawn);
+        // amountWithdrawn = _vault.withdraw(
+        //     address(_baseToken), // asset
+        //     // NOTE: Withdrawals are processed in base, so we have to convert
+        //     // the share amount to a base amount.
+        //     _convertToBase(_shareAmount), // amount
+        //     _destination // onBehalfOf
+        // );
```
jalextowle, line 124: Nit: Remove this.

```suggestion
```

contracts/src/instances/aave-l2/AaveL2Base.sol
```diff
@@ -0,0 +1,240 @@
...
+            msg.sender,
+            address(this),
+            // NOTE: The AToken interface transfers in base, so we have to
+            // convert the share amount to a base amount.
+            _convertToBase(_shareAmount)
+        );
+    }
+
+    // FIXME: This still needs to be updated.
+    //
```
jalextowle, line 101: Remove this.

jalextowle (approved):
Sorry for the delay on this. A couple of nits, but otherwise, this LGTM

PR #1123, opened 2024-08-07
Relaxed the version requirements for interfaces, libraries, and test contracts

jalextowle:
# Description

If other projects (internally or externally) need to make use of our libraries or interfaces, versioning could be a problem. Fixing the version is a good safeguard that production contracts are compiled with the right version of Solidity, but it's appropriate for libraries to be more flexible since they may be used long after the original contracts are deployed. This PR relaxes the constraints on our libraries, interfaces, and test utilities without changing the version pragmas of the core contracts that will be deployed.


mcclurejt (approved):
Might want to add `test/debug/Debug.t.sol` but otherwise looks good!

sentilesdal (approved):

PR #1124, opened 2024-08-07
Upgrades the Solidity version to `0.8.22`

jalextowle:
# Description

This PR upgrades the Solidity version of contracts that aren't libraries, interfaces, or test utilities to `0.8.22` and removes the unchecked loop counters that are no longer necessary.


mcclurejt (approved):
Might need to remove one more unchecked block at `contracts/src/internal/HyperdriveCheckpoint.sol:131` but could totally be missing something.

Otherwise, awesome sauce!

jalextowle:
> Might need to remove one more unchecked block at `contracts/src/internal/HyperdriveCheckpoint.sol:131` but could totally be missing something.
> 
> Otherwise, awesome sauce!

Good callout. I thought about this but ultimately decided against it since it isn't a normal loop counter and I thought the loop looked weird if I added it to the for-loop.

PR #1127, opened 2024-08-08
Added a test suite for the Morpho wstETH/USDC market

jalextowle:
# Description

This PR adds a test suite for the Morpho wstETH/USDC market to ensure that precision issues don't result in problems. This is necessary since USDC only has 6 decimals and Morpho has an abnormally low share price of ~1e12.


test/utils/InstanceTest.sol
```diff
@@ -86,7 +86,7 @@ abstract contract InstanceTest is HyperdriveTest {
         super.setUp();
 
         // Initial contribution.
-        uint256 contribution = 5_000e18;
```
jalextowle, line 89: These changes aren't going to work for the other yield sources. What I think would fix all of these issues would be to pass in the decimals as part of the config, and then we could calculate the contribution amounts as:

```
uint256 contribution = 5_000 * 10 ** config.decimals;
```

contracts/src/instances/morpho-blue/MorphoBlueBase.sol
```diff
@@ -125,19 +125,22 @@ abstract contract MorphoBlueBase is HyperdriveBase {
...
-                irm: _irm,
-                lltv: _lltv
-            }),
-            _convertToBase(_shareAmount),
-            0,
-            address(this),
-            _destination
-        );
+        uint256 baseAmount = _convertToBase(_shareAmount);
+        if (baseAmount > 0) {
```
jalextowle, line 129: This was the fix I mentioned the other day.

test/instances/ezETH/EzETHHyperdrive.t.sol
```diff
@@ -60,22 +60,33 @@ contract EzETHHyperdriveTest is InstanceTest {
...
-            1e6,
-            1e15,
-            POSITION_DURATION_15_DAYS,
-            false,
-            true,
-            false,
-            true,
-            abi.encodeWithSelector(IHyperdrive.UnsupportedToken.selector)
-        );
+        InstanceTestConfig({
```
jrhea, line 63: i like the named struct params. nice

test/instances/morpho-blue/MorphoBlue_wstETH_USDC_Hyperdrive.t.sol
```diff
@@ -0,0 +1,818 @@
...
+            vaultSharesTokenWhaleAccounts: new address[](0),
+            baseToken: IERC20(LOAN_TOKEN),
+            vaultSharesToken: IERC20(address(0)),
+            // NOTE: The share tolerance is quite high for this integration
+            // because the vault share price is ~1e12, which means that just
+            // multiplying or dividing by the vault is an imprecise way of
+            // converting between base and vault shares. We included more
+            // assertions than normal to the round trip tests to verify that
+            // the calculations satisfy our expectations of accuracy.
+            shareTolerance: 1e8,
```
jrhea, line 89: seems like this is lower than the other morpho test share tolerances

test/instances/morpho-blue/MorphoBlue_wstETH_USDC_Hyperdrive.t.sol
```diff
@@ -0,0 +1,818 @@
...
+                    oracle: ORACLE,
+                    irm: IRM,
+                    lltv: LLTV
+                })
+            );
+        assertApproxEqAbs(totalSupplyAssets, totalBaseBefore - baseProceeds, 1);
+        assertApproxEqAbs(
+            totalSupplyShares,
+            totalSharesBefore - hyperdrive.convertToShares(baseProceeds),
+            1e6
```
jrhea, line 319: Is this tolerance higher due to imprecision with convertToShares?

test/instances/morpho-blue/MorphoBlue_wstETH_USDC_Hyperdrive.t.sol
```diff
@@ -0,0 +1,818 @@
...
+            withdrawalShares
+        );
+        assertEq(withdrawalSharesRedeemed, withdrawalShares);
+
+        // Bob should receive base approximately equal in value to his present
+        // value.
+        assertApproxEqAbs(
+            baseProceeds,
+            withdrawalShares.mulDown(lpSharePrice),
+            1e3
```
jrhea, line 515: okay so this must have been one of the tolerances you had to tweak for a 6 decimal place USDC token. I believe this was 1e12 for the other tokens (which obviously wouldn't work here); however, an error of 12/18 > an error of 3/6 so this seems acceptable proportionally

test/instances/aave/AaveHyperdrive.t.sol
```diff
@@ -43,27 +43,36 @@ contract AaveHyperdriveTest is InstanceTest {
...
+            InstanceTestConfig({
+                name: "Hyperdrive",
+                kind: "AaveHyperdrive",
+                decimals: 18,
+                baseTokenWhaleAccounts: baseTokenWhaleAccounts,
+                vaultSharesTokenWhaleAccounts: vaultSharesTokenWhaleAccounts,
+                baseToken: WETH,
+                vaultSharesToken: IERC20(address(AWETH)),
+                shareTolerance: 1e3,
+                minimumShareReserves: 1e15,
```
mcclurejt, line 58: Do we want to use constants for this, `shareTolerance` and/or `minimumTransactionAmount` like with position duration?

test/instances/morpho-blue/MorphoBlue_wstETH_USDC_Hyperdrive.t.sol
```diff
@@ -0,0 +1,818 @@
...
+        AccountBalances memory hyperdriveBalancesBefore = getAccountBalances(
+            address(hyperdrive)
+        );
+
+        // Bob closes his long with base as the target asset.
+        uint256 baseProceeds = closeShort(bob, maturityTime, _shortAmount);
+
+        // Bob should receive less base than he paid since no time as passed
+        // and the fees are non-zero.
+        assertLt(baseProceeds, basePaid);
```
jrhea, line 703: okay so on these instantaneous tests you are just making sure that the proceeds are less than the basePaid. i think this is a better way to test it anyways. we should probably add this type of assertion to the other morpho tests if we havent already

test/utils/InstanceTest.sol
```diff
@@ -463,7 +522,16 @@ abstract contract InstanceTest is HyperdriveTest {
         }
 
         // Contribution in terms of base.
-        uint256 contribution = 1_000e18;
+        uint256 contribution;
+        if (!isBaseETH) {
```
mcclurejt, line 526: maybe add some comments around branches of this if statement (or describe in block comment)

test/instances/morpho-blue/MorphoBlue_wstETH_USDC_Hyperdrive.t.sol
```diff
@@ -0,0 +1,818 @@
...
+
+        // Bob closes his long with base as the target asset.
+        uint256 baseProceeds = closeShort(bob, maturityTime, _shortAmount);
+
+        // Bob should receive almost exactly the interest that accrued on the
+        // bonds that were shorted.
+        assertApproxEqAbs(
+            baseProceeds,
+            _shortAmount.mulDown(_variableRate),
+            1e10
```
jrhea, line 754: ?

test/utils/InstanceTest.sol
```diff
@@ -209,9 +267,14 @@ abstract contract InstanceTest is HyperdriveTest {
 
         // Alice gives approval to the deployer coordinator to fund the market.
         if (asBase && !isBaseETH) {
```
mcclurejt, line 269: finding the branch logic a bit hard to follow at first glance... might be helpful to have some comments around it or restructure

test/utils/InstanceTest.sol
```diff
@@ -123,6 +150,25 @@ abstract contract InstanceTest is HyperdriveTest {
         factory.addDeployerCoordinator(deployerCoordinator);
 
         // Deploy all Hyperdrive contracts using deployer coordinator contract.
+        uint256 contribution;
+        if (config.enableShareDeposits && !config.isRebasing) {
```
mcclurejt, line 154: could use some comments around branching logic here

test/utils/InstanceTest.sol
```diff
@@ -927,8 +1001,15 @@ abstract contract InstanceTest is HyperdriveTest {
...
         // Ensure Bob is credited the correct amount of bonds.
-        assertLe(baseProceeds, longAmount);
-        assertApproxEqAbs(baseProceeds, longAmount, 10);
+        assertLe(
+            baseProceeds,
+            longAmount.mulDown(ONE - hyperdrive.getPoolConfig().fees.flat)
+        );
+        assertApproxEqAbs(
+            baseProceeds,
+            longAmount.mulDown(ONE - hyperdrive.getPoolConfig().fees.flat),
```
mcclurejt, line 1010: was this able to pass earlier due to the flat fee being smaller than the "approx" tolerance?

good catch!!!!

mcclurejt (commented):
Nits and q's, nothing major

mcclurejt (approved):

jrhea (approved):
Great work!! I am going to approve, but I do have a couple of questions and a suggestion to tighten the error bounds on `test_round_trip_short_maturity`

test/instances/aave/AaveHyperdrive.t.sol
jalextowle: I don't think so. Two things: (1) it obscures the values and (2) the values are different for most of the tests.

test/utils/InstanceTest.sol
jalextowle: Yeah, it was

test/instances/morpho-blue/MorphoBlue_wstETH_USDC_Hyperdrive.t.sol
jalextowle: You mean higher relative to the token decimals? Yes it seems to be a precision issue, but I'll take another look to see if I can bring it down.

test/instances/morpho-blue/MorphoBlue_wstETH_USDC_Hyperdrive.t.sol
jalextowle: I'd say that an error of 1e12 is better than 1e3 since 1e3 only leaves 3 decimals of precision and 1e12 leaves 6. That said, There are some confounding variables caused by using `mulDown` instead of `mulDivDown` with the present value and the lp total supply. I'll try to tighten the bound.

jalextowle: I was able to drop this to `100`.

test/instances/morpho-blue/MorphoBlue_wstETH_USDC_Hyperdrive.t.sol
jalextowle: Yep, I added these to the other tests. I still kept the `assertApproxEqAbs`, but I added a `assertLt`.

test/instances/morpho-blue/MorphoBlue_wstETH_USDC_Hyperdrive.t.sol
jalextowle: I'll update the tolerance. Awesome catch!

test/instances/morpho-blue/MorphoBlue_wstETH_USDC_Hyperdrive.t.sol
jalextowle: I was able to drop this to 1e3 which seems acceptable.

test/instances/morpho-blue/MorphoBlue_wstETH_USDC_Hyperdrive.t.sol
jalextowle: I was able to drop this to 10.

test/instances/aave/AaveHyperdrive.t.sol
jalextowle: I'd feel differently if we used it in multiple places, but the point of the config object is that we will use the configs value throughout the test.

Issue #1104, opened 2024-07-17
wstETH L2 Integration


jalextowle:
Integrate the L2 version of wstETH with Hyperdrive

PR #1135, opened 2024-08-17
Chainlink Integration

jalextowle:
# Resolved Issues

Fixes: #1104

# Description

This PR adds a chainlink integration. This can be used with general price feeds, although the intended use is with price feeds like the `wsteth-eth reference rate` on Gnosis Chain. This integration uses the output of a given Chainlink aggregator as the vault share price. It doesn't support base as a deposit asset since the base token may not even exist. Some documentation is included in the implementation that describes how Chainlink downtime or deprecation will impact the instance.

# Review Checklists

be removed.

## [[Reviewer Name]]

- [ ] **Tokens**
              token spend?
- [ ] **Reentrancy**
- [ ] **Gas Optimizations**
          memory variables?
          issues?
- [ ] **Visibility**
- [ ] **Math**
          catch underflows?
- [ ] **Testing**


contracts/src/deployers/chainlink/ChainlinkHyperdriveDeployerCoordinator.sol
```diff
@@ -0,0 +1,158 @@
...
+    /// @param _extraData The extra data containing the Chainlink aggregator
+    ///        and the instance's decimals.
+    function _checkPoolConfig(
+        IHyperdrive.PoolDeployConfig memory _deployConfig,
+        bytes memory _extraData
+    ) internal view override {
+        // Perform the default checks.
+        super._checkPoolConfig(_deployConfig, _extraData);
+
+        // Ensure that the base token is zero.
```
jrhea, line 110: oh interesting. we will need to let the frontend team know about this wrinkle

contracts/src/interfaces/IChainlinkAggregatorV3.sol
```diff
@@ -0,0 +1,34 @@
+// SPDX-License-Identifier: Apache-2.0
+pragma solidity ^0.8.0;
```
jrhea, line 2:
```suggestion
pragma solidity ^0.8.22;
```

contracts/src/libraries/SafeCast.sol
```diff
@@ -26,6 +26,18 @@ library SafeCast {
         y = uint128(x);
     }
 
+    // FIXME: Add unit tests for this.
```
jrhea, line 29: Did you add a test for this?

test/instances/chainlink/ChainlinkTest.t.sol
```diff
@@ -0,0 +1,691 @@
...
+            false
+        );
+        uint256 baseProceeds = hyperdrive.convertToBase(vaultSharesProceeds);
+
+        // Bob should receive almost exactly the interest that accrued on the
+        // bonds that were shorted.
+        assertApproxEqAbs(
+            baseProceeds,
+            _shortAmount.mulDown(_variableRate),
+            1e9
```
jrhea, line 643: This error can be lowered to 1e4 (i tested it with 10k fuzz runs)

jrhea (approved):
Nice work! I just had a couple of nits, but I will go ahead and approve. I would consider having @DannyDelott take a look at the PR in case he has any issues with the lack of a baseToken.

contracts/src/interfaces/IChainlinkAggregatorV3.sol
jalextowle: We've been doing `^0.8.20` since a lot of contracts still use `0.8.20`. Regardless, it's a good catch, and I'll update it.

contracts/src/deployers/chainlink/ChainlinkHyperdriveCoreDeployer.sol
```diff
@@ -0,0 +1,62 @@
...
+/// @title ChainlinkHyperdriveCoreDeployer
+/// @notice The core deployer for the ChainlinkHyperdrive implementation.
+/// @custom:disclaimer The language used in this code is for coding convenience
+///                    only, and is not intended to, and does not, have any
+///                    particular legal or regulatory significance.
+contract ChainlinkHyperdriveCoreDeployer is IHyperdriveCoreDeployer {
+    /// @notice Deploys a Hyperdrive instance with the given parameters.
+    /// @param __name The name of the Hyperdrive pool.
+    /// @param _config The configuration of the Hyperdrive pool.
+    /// @param _extraData The extra data containing the Chainlink aggregator.
```
mcclurejt, line 19: nice

contracts/src/instances/chainlink/ChainlinkHyperdrive.sol
```diff
@@ -0,0 +1,99 @@
...
+/// SSS\\\\\\\\000000000000////////|:::::0x666:::::|\\\\\\\\00000000000//////////SSS
+/// SSS|||||||||||||||||||||||||||||:::::::::::::::||||||||||||||||||||||||||||||SSS
+/// SSSSSSSS|_______________|______________||_______________|______________|SSSSSSSS
+/// SSSSSSSS                                                                SSSSSSSS
+/// SSSSSSSS                                                                SSSSSSSS
+///
+/// @author DELV
+/// @title ChainlinkHyperdrive
+/// @notice A Hyperdrive instance that uses a Chainlink vault as the yield source.
+/// @dev This instance pulls the vault share price from a Chainlink aggregator.
```
mcclurejt, line 52: Maybe worth noting here the two peculiarities of deploying this contract:
- baseToken must be zero
- extraData should specify the chainlink aggregator

test/instances/chainlink/ChainlinkTest.t.sol
```diff
@@ -0,0 +1,691 @@
+// SPDX-License-Identifier: Apache-2.0
+pragma solidity 0.8.22;
+
+import { stdStorage, StdStorage } from "forge-std/Test.sol";
+import { ChainlinkHyperdriveCoreDeployer } from "contracts/src/deployers/chainlink/ChainlinkHyperdriveCoreDeployer.sol";
```
mcclurejt, line 5: relative import paths

test/instances/chainlink/ChainlinkTest.t.sol
```diff
@@ -0,0 +1,691 @@
...
+            address(IChainlinkHyperdrive(address(hyperdrive)).aggregator()),
+            address(CHAINLINK_AGGREGATOR_PROXY)
+        );
+        (, uint256 totalShares) = getTokenBalances(address(hyperdrive));
+        assertEq(hyperdrive.totalShares(), totalShares);
+    }
+
+    /// LP ///
+
+    function test_round_trip_lp_instantaneous(uint256 _contribution) external {
```
mcclurejt, line 267: would like to see high-level summary comments on what each of these tests are doing but can totally understand if that's not something we want to tackle rn

test/instances/chainlink/ChainlinkTest.t.sol
```diff
@@ -0,0 +1,691 @@
...
+    // The configuration for the Instance testing suite.
+    InstanceTestConfig internal __testConfig =
+        InstanceTestConfig({
+            name: "Hyperdrive",
+            kind: "ChainlinkHyperdrive",
+            baseTokenWhaleAccounts: new address[](0),
+            vaultSharesTokenWhaleAccounts: vaultSharesTokenWhaleAccounts,
+            baseToken: IERC20(address(0)),
+            vaultSharesToken: WSTETH,
+            // FIXME
```
mcclurejt, line 62: dangling fixme

mcclurejt (approved):
awesome stuff! coolest integration yet imo

test/instances/chainlink/ChainlinkTest.t.sol
jalextowle: I'll do this when I clean up the integration test suite.

PR #1136, opened 2024-08-20
Implemented an Admin Controller

jalextowle:
# Description

Simplified the admin configuration of Hyperdrive instances. This feature wires up Hyperdrive to look up the admin configuration from an "admin controller." This admin controller must provide the following data: `hyperdriveGovernance`, `feeCollector`, `sweepCollector`, `checkpointRewarder`, and `defaultPausers`. These are all of the fields that can currently be updated by governance on the Hyperdrive factory, and the intention is that the factory will be the admin controller for deployed pools.

This feature is fully backward-compatible. `setPauser` and `setGovernance` were updated to be stub functions since the Hyperdrive factory calls these functions during the `deployAndInitialize` routine. `IHyperdrive.PoolConfig` wasn't updated to include the `adminController` to avoid changing the interface, and all of the interface changes live inside the deployers and the contract constructors. Handling the changes at this level avoids any deployer coordinator interface changes since the deployer coordinators know which factory they are connected to.


contracts/src/interfaces/IHyperdriveAdminController.sol
```diff
@@ -0,0 +1,30 @@
...
+    /// @return The sweep collector.
+    function sweepCollector() external view returns (address);
+
+    /// @notice Returns the checkpoint rewarder that can pay out rewards to
+    ///         checkpoint minters.
+    /// @return The checkpoint rewarder.
+    function checkpointRewarder() external view returns (address);
+
+    // TODO: A better interface would be `isPauser`, but this can't be changed
+    //       without swapping out the factory.
```
jrhea, line 24: Why would isPauser be better?

contracts/src/internal/HyperdriveStorage.sol
```diff
@@ -139,9 +131,17 @@ abstract contract HyperdriveStorage is ReentrancyGuard {
 
     /// Constructor ///
 
+    // TODO: It would be good to update `PoolConfig` when we make other breaking
+    //       changes.
```
jrhea, line 135: I know I'm not telling you anything you don't know, but any changes to poolConfig are going to cause issues with the frontend. They will have to find an efficient way to load different SDKs for different pools. This could bloat the bundle size unless they come up with something efficient.

contracts/src/external/HyperdriveTarget0.sol
```diff
@@ -35,12 +36,31 @@ abstract contract HyperdriveTarget0 is
...
 
     /// Admin ///
 
+    // TODO: This function doesn't do anything anymore and is only here for
+    // backwards compatability. This can be removed when the factory is upgraded.
+    //
+    /// @notice A stub for the old setPauser functions that doesn't do anything
+    ///         anymore.
+    /// @dev Don't call this. It doesn't do anything.
+    function setGovernance(address) external {}
```
jrhea, line 54: I'm not totally following this. Maybe we can do a walkthrough. 

One question, why not just have it revert if it is called?

contracts/src/interfaces/IHyperdriveCore.sol
```diff
@@ -190,27 +206,6 @@ interface IHyperdriveCore is IMultiTokenCore {
     /// @param _status True to pause all deposits and false to unpause them.
     function pause(bool _status) external;
 
-    /// @notice Allows governance to transfer the fee collector role.
-    /// @param _who The new fee collector address.
-    function setFeeCollector(address _who) external;
```
jrhea, line 195: Is the idea that we just use the Factory's, for example, `updateFeeCollector` function to set the fee collector now?

contracts/src/internal/HyperdriveAdmin.sol
```diff
@@ -30,16 +30,16 @@ abstract contract HyperdriveAdmin is IHyperdriveEvents, HyperdriveBase {
         _checkOptions(_options);
 
         // Ensure that the destination is set to the fee collector.
-        address feeCollector = _feeCollector;
+        address feeCollector = _adminController.feeCollector();
```
jrhea, line 33: so one side effect of this change is that all feeCollectors, pausers, etc will always be the same for all pools. is that right? if so, is that what we want?

contracts/src/internal/HyperdriveAdmin.sol
```diff
@@ -30,16 +30,16 @@ abstract contract HyperdriveAdmin is IHyperdriveEvents, HyperdriveBase {
...
             revert IHyperdrive.InvalidFeeDestination();
         }
 
         // Ensure that the caller is authorized to collect fees.
         if (
-            !_pausers[msg.sender] &&
             msg.sender != feeCollector &&
-            msg.sender != _governance
+            msg.sender != _adminController.hyperdriveGovernance() &&
+            !_isPauser(msg.sender)
```
jrhea, line 42: i'm sure i'm thinking about this wrong, but it doesn't make sense to me as to why we care who calls `_collectGovernanceFees()`.  as long as the fees go to the feeCollector, we should allow anyone to call it, right?

contracts/src/internal/HyperdriveAdmin.sol
```diff
@@ -144,11 +81,11 @@ abstract contract HyperdriveAdmin is IHyperdriveEvents, HyperdriveBase {
...
     function _sweep(IERC20 _target) internal nonReentrant {
         // Ensure that the caller is authorized to sweep tokens.
-        address sweepCollector = _sweepCollector;
+        address sweepCollector = _adminController.sweepCollector();
         if (
-            !_pausers[msg.sender] &&
             msg.sender != sweepCollector &&
-            msg.sender != _governance
+            msg.sender != _adminController.hyperdriveGovernance() &&
+            !_isPauser(msg.sender)
```
jrhea, line 88: Same with this, should we care who call this as long as the sweepCollector gets the tokens?

jrhea (commented):
if we decide to go this route, it would be good to thoroughly document in the operations playbook how this is used and what existing pools will have to still be manually updated

contracts/src/external/HyperdriveTarget0.sol
jalextowle: It’s called during deployAndInitialize, so that would break backward compatibility and require a new factory to be deployed

contracts/src/interfaces/IHyperdriveAdminController.sol
jalextowle: It would be more efficient. Right now, we have to scan through the full list of pausers to check if an address is a pauser. This places constraints on the amount of pausers that the factory can have.

jrhea (approved):

contracts/src/internal/HyperdriveStorage.sol
jalextowle: Yep, 100%. I get it. This is just to remind us to update the pool config with this information when we feel that it is appropriate.

contracts/src/internal/HyperdriveAdmin.sol
jalextowle: We talked about this offline. I think it's good to have basic admin controls on this function in case the treasury is compromised or there is something sensitive to think about.

contracts/src/internal/HyperdriveAdmin.sol
jalextowle: Similar story to the `collectGovernanceFees` functions.

contracts/src/internal/HyperdriveAdmin.sol
jalextowle: Yep, as mentioned offline, I think that this is a practically going to be fine for the foreseeable future as long as it's clearly communicated. For most of the parameters, the address can be set to an access controller or to a contract with a similar architecture to the checkpoint rewarder to support more bespoke functionality.

PR #1146, opened 2024-08-27
consistent registry addresses

mcclurejt:
# Resolved Issues

# Description

- Use CREATE3 to deploy the registry so that we can maintain consistent addresses across all chains

# Review Checklists

be removed.

## [[Reviewer Name]]

- [ ] **Tokens**
              token spend?
- [ ] **Reentrancy**
- [ ] **Gas Optimizations**
          memory variables?
          issues?
- [ ] **Visibility**
- [ ] **Math**
          catch underflows?
- [ ] **Testing**


jrhea (commented):
this lgtm. i'll wait for @jalextowle to give the final approval

contracts/src/factory/HyperdriveCreate3Factory.sol
```diff
@@ -0,0 +1,41 @@
...
+contract HyperdriveCreate3Factory {
+    /// @notice Deploy the contract with the provided creation code to a
+    ///         precomputable address determined by only the deployer address
+    ///         and salt.
+    /// @param _salt Salt to use for the deployment.
+    /// @param _creationCode Creation code of the contract to deploy.
+    function deploy(
+        bytes32 _salt,
+        bytes memory _creationCode
+    ) external payable returns (address deployed) {
```
jalextowle, line 22: To fix the issue that you had with admin controls, let's add a `bytes memory _initializationData` parameter. If this is non-empty, you'll call back into the deployed contract. This will let you update the admin address (for example).

jalextowle (approved):
Nice! LGTM once the initialization logic is updated.

contracts/src/factory/HyperdriveCreate3Factory.sol
```diff
@@ -0,0 +1,41 @@
+// SPDX-License-Identifier: Apache-2.0
+pragma solidity 0.8.22;
+
+import { CREATE3 } from "solmate/utils/CREATE3.sol";
+
+/// @author DELV
+/// @title HyperdriveCreate3Factory
+/// @notice Uses CREATE3 to deploy a contract to a precomputable address that
+///         is only dependent on the deployer's address and the provided salt.
```
jalextowle, line 9: Let's add a `@dev` comment. This kind of deployment isn't suitable for high-volume contracts since it uses a proxy. I'm fine with using it for the registry, but every call will cost an additional 2700+ gas because of the extra call.

PR #1151, opened 2024-08-29
Improved the instance tests

jalextowle:
# Description

This PR refactors the instance tests to reduce the amount of changes that must be made for each new instance. This also paves the way for future changes to improve the integration developer experience.


jrhea (approved):
This is great!!

test/instances/aave/AaveHyperdrive.t.sol
```diff
@@ -58,18 +58,40 @@ contract AaveHyperdriveTest is InstanceTest {
...
                 enableShareWithdraws: true,
                 baseWithdrawError: new bytes(0),
                 isRebasing: true,
-                fees: IHyperdrive.Fees({
-                    curve: 0,
-                    flat: 0,
-                    governanceLP: 0,
-                    governanceZombie: 0
-                })
+                // The base test tolerances.
```
mcclurejt, line 73: I think using a builder pattern off a default configuration might make this more digestable by highlighting which config fields are out of the ordinary....

Maybe something like... 

```solidity
DefaultInstanceTestConfig
    .fees(0)
    .roundTripLongInstantaneousWithBaseUpperBoundTolerance(1e3)
    ...
```



Not something that needs to get done now, just food for thought

mcclurejt (approved):
Love to see this... awesome work!

image

test/instances/aave/AaveHyperdrive.t.sol
jalextowle: Yeah, this could be nice. I don't know how well this would work considering that this is a constructor param, but worth playing around with in the future.

PR #1154, opened 2024-08-30
stUSD Integration

jalextowle:
# Description

This PR includes an instance test for the stUSD integration. This is an ERC4626 integration, so no new contracts needed to be added. This also includes the deployment script.


tasks/deploy/config/mainnet/stusd-182day.ts
```diff
@@ -0,0 +1,84 @@
...
+    coordinatorAddress: async (hre) =>
+        hre.hyperdriveDeploy.deployments.byName(
+            MAINNET_ERC4626_COORDINATOR_NAME,
+        ).address,
+    deploymentId: keccak256(toBytes(MAINNET_STUSD_182DAY_NAME)),
+    salt: toBytes32("0x69420"),
+    extraData: "0x",
+    contribution: CONTRIBUTION,
+    fixedAPR: parseEther("0.0829"),
+    timestretchAPR: parseEther("0.05"),
```
jrhea, line 35: we've seen this go as high as 35%. it would be good to test to see that it can hit that APY with this timestretch. i am pretty sure it can, but it won't be smooth.

test/instances/erc4626/StUSD.t.sol
```diff
@@ -0,0 +1,135 @@
...
+        vm.store(
+            address(STUSD),
+            lastUpdateLocation,
+            bytes32((block.timestamp << 208) | uint256(STUSD.rate()))
+        );
+        (totalAssets, ) = totalAssets.calculateInterest(
+            variableRate,
+            timeDelta
+        );
+        bytes32 balanceLocation = keccak256(abi.encode(address(STUSD), 51));
```
jrhea, line 132: image

jrhea (approved):
Approved. Great work! I think we should just discuss if we want to maybe bump the ts up a little, but it is good to merge imo

tasks/deploy/config/mainnet/stusd-182day.ts
jalextowle: Gotcha, when I looked on DeFiLlama, I didn't see anything above 10%. I'll update it, but let's follow up with this next week. I must have missed something.

PR #1167, opened 2024-09-10
cbETH Integration on Base

jalextowle:
# Description

This PR adds tests and deployment tools for the cbETH integration on Base. This integration makes use of the `ChainlinkHyperdrive` instance, and as part of this PR, I refactored the Chainlink instance tests to make it easier to add more chainlink integrations.


contracts/src/interfaces/IChainlinkAggregatorV3.sol
```diff
@@ -2,6 +2,8 @@
 pragma solidity ^0.8.20;
 
 interface IChainlinkAggregatorV3 {
+    function aggregator() external view returns (address);
```
mcclurejt, line 5: natspec?

tasks/deploy/lib/constants.ts
```diff
@@ -127,6 +127,23 @@ export const WSTETH_WHALE_MAINNET =
...
+    "0x2Ae3F1Ec7F1F5012CFEab0185bfc7aa3cf0DEc22" as Address;
+
+export const CHAINLINK_AGGREGATOR_CBETH_ETH_PROXY_BASE =
+    "0x868a501e68F3D1E89CfC0D22F6b22E8dabce5F04" as Address;
+
+// ╭─────────────────────────────────────────────────────────╮
+// │ Base Whales                                             │
+// ╰─────────────────────────────────────────────────────────╯
+
+export const CBETH_WHALE_GNOSIS =
```
mcclurejt, line 144:
```suggestion
export const CBETH_WHALE_BASE =
```

will need to update elsewhere

test/utils/BaseTest.sol
```diff
@@ -7,30 +7,31 @@ import { IERC20 } from "../../contracts/src/interfaces/IERC20.sol";
...
-
-    address minter;
-    address deployer;
-    address feeCollector;
-    address sweepCollector;
-    address governance;
-    address pauser;
-    address registrar;
-    address rewardSource;
+    address internal alice;
```
mcclurejt, line 10: why internal? also realize it would be a huge PITA but doesn't style guide say internal vars should have `_` prefix?

mcclurejt (approved):
LGTM all comments are nits

contracts/src/interfaces/IChainlinkAggregatorV3.sol
jalextowle: None of the other interfaces have natspec. I don't think we need this here.

test/utils/BaseTest.sol
jalextowle: There is a compiler warning for "visibility" should be specified. Yeah, the style guide does say that, but the style guide has not been applied consistently towards the Hyperdrive test suites. I don't think it's a bad idea to address this at some point, but this PR is not the time for that.

test/instances/chainlink/CbETHBase.t.sol
```diff
@@ -0,0 +1,91 @@
...
+                minimumShareReserves: 1e15,
+                minimumTransactionAmount: 1e15,
+                positionDuration: POSITION_DURATION,
+                fees: IHyperdrive.Fees({
+                    curve: 0,
+                    flat: 0,
+                    governanceLP: 0,
+                    governanceZombie: 0
+                }),
+                enableBaseDeposits: false,
```
jrhea, line 45: what's the issue with base deposits btw?

jalextowle: We’re using the Chainlink integration which doesn’t support base deposits and withdrawals. The reason is that chain link oracles don’t give us a conversion mechanism from base to vault shares or vice versa. 

There may be a different way to integrate with Chainlink on base, but it seemed to me like this was the way to go (and is consistent with our wsteth integration on Gnosis Chain).

jrhea: that's right. i wrote the comment, went to eat lunch and realized it

test/instances/chainlink/ChainlinkHyperdriveInstanceTest.t.sol
```diff
@@ -224,10 +144,10 @@ contract ChainlinkHyperdriveTest is InstanceTest {
             ,
             uint256 updatedAt,
 
-        ) = CHAINLINK_AGGREGATOR_PROXY.latestRoundData();
+        ) = _chainlinkAggregator.latestRoundData();
         uint256 answer_ = uint256(answer);
 
-        // Accrue interest in the Chainlink wstETH market. We do this by
+        // Accrue interest in the Chainlink aggregator. We do this by
```
jrhea, line 150: thumbs up

jrhea (approved):

PR #1171, opened 2024-09-13
Corn Integration

jalextowle:
# Description

Adds a `Corn` integration that makes it possible to launch a Hyperdrive pool with any of the assets allowed by the Corn Silo, adds an instance test for a Corn instance that uses LBTC, and adds deployment tools that deploy with LBTC.

### DIFF
--- a/codegen/example/config.yaml
+++ b/codegen/example/config.yaml
@@ -4,16 +4,16 @@
 name:
   # Capitalized version of the name to be used i.e. for contract names, file
   # names, and comments.
-  capitalized: "Corn"
+  capitalized: "ExampleName"
 
   # All upper case name to be used for constants.
-  uppercase: "CORN"
+  uppercase: "EXAMPLE_NAME"
 
   # All lower case name to be used for directories.
-  lowercase: "corn"
+  lowercase: "example-name"
 
   # Camel case name to be used i.e. for variables.
-  camelcase: "corn"
+  camelcase: "exampleName"
 
 # Configuration parameters for the hyperdrive instance.
 contract:
--- a/codegen/example/config.yaml
+++ b/codegen/example/config.yaml
@@ -4,19 +4,16 @@
 name:
   # Capitalized version of the name to be used i.e. for contract names, file
   # names, and comments.
-  capitalized: "MorphoBlue"
+  capitalized: "Corn"
 
   # All upper case name to be used for constants.
-  uppercase: "MORPHO_BLUE"
-
-  # All upper case name to be used for constants.
-  uppercase: "TEST_ETH"
+  uppercase: "CORN"
 
   # All lower case name to be used for directories.
-  lowercase: "morpho-blue"
+  lowercase: "corn"
 
   # Camel case name to be used i.e. for variables.
-  camelcase: "morphoBlue"
+  camelcase: "corn"
 
 # Configuration parameters for the hyperdrive instance.
 contract:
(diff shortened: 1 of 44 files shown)


codegen/example/config.yaml
```diff
@@ -4,19 +4,16 @@
 name:
   # Capitalized version of the name to be used i.e. for contract names, file
   # names, and comments.
-  capitalized: "MorphoBlue"
+  capitalized: "Corn"
```
jrhea, line 7: did you mean to check this in?

contracts/src/instances/corn/CornBase.sol
```diff
@@ -0,0 +1,147 @@
...
+
+    /// @dev Accepts a deposit from the user in base.
+    /// @param _baseAmount The base amount to deposit.
+    /// @return The shares that were minted in the deposit.
+    /// @return The amount of ETH to refund. Since this yield source isn't
+    ///         payable, this is always zero.
+    function _depositWithBase(
+        uint256 _baseAmount,
+        bytes calldata // unused
+    ) internal override returns (uint256, uint256) {
```
jrhea, line 47: i imagine this specific implementation is only for the lbtc silo, right? there is another silo that accepts dai, sdai, usdc so we'd need to account for the multiple base types for those silos, right?

tasks/deploy/config/mainnet/corn-lbtc-91day.ts
```diff
@@ -0,0 +1,88 @@
...
+    prefix: "Corn",
+    coordinatorAddress: async (hre) =>
+        hre.hyperdriveDeploy.deployments.byName(MAINNET_CORN_COORDINATOR_NAME)
+            .address,
+    deploymentId: keccak256(toBytes(MAINNET_CORN_LBTC_91DAY_NAME)),
+    salt: toBytes32("0x69420"),
+    extraData: "0x",
+    contribution: CONTRIBUTION,
+    // FIXME
+    fixedAPR: parseEther("0.0666"),
```
jrhea, line 32: the pendle pt is trading at 9.1%, fyi

contracts/src/instances/corn/CornBase.sol
jalextowle: It should work for all of them. It’s made to be generic enough to handle all of the assets supported on the Corn Silo. For some assets, we’d need a zap, but I don’t think anything would need to change to support other assets.

contracts/src/instances/corn/CornHyperdrive.sol
```diff
@@ -0,0 +1,99 @@
...
+/// |||////////000000000000\\\\\\\\|:::::::::::::::|////////00000000000\\\\\\\\\\|||
+/// SSS\\\\\\\\000000000000////////|:::::0x666:::::|\\\\\\\\00000000000//////////SSS
+/// SSS|||||||||||||||||||||||||||||:::::::::::::::||||||||||||||||||||||||||||||SSS
+/// SSSSSSSS|_______________|______________||_______________|______________|SSSSSSSS
+/// SSSSSSSS                                                                SSSSSSSS
+/// SSSSSSSS                                                                SSSSSSSS
+///
+/// @author DELV
+/// @title CornHyperdrive
+/// @notice A Hyperdrive instance that uses a Corn vault as the yield source.
```
mcclurejt, line 54: Might be nice to have a lil blurb here on what a corn silo is (assuming it's a vault)

mcclurejt (approved):
aMAIZEing

PR #1179, opened 2024-09-25
Uni v3 Zap

jalextowle:
# Description

This PR adds a zap contract that supports generalized swapping between assets using Uniswap V3. Each Hyperdrive trading or LP function has a direct anologue in the zap contract that exposes all of the parameters as well as a few extra parameters for things like the Uniswap swap object (`ISwapRouter.ExactInputParams`) and special-case parameters to handle different edge-cases in the zap. 

Extensive tests were added that should test the different paths through these functions. Special care was taken to verify that ETH, WETH, and rebasing tokens are properly handled by the zap contract. Since Uniswap v3 doesn't really support ETH as a first-class citizen in all situations, there are several flows within the zap contract where ETH will be converted to WETH or vice-versa. For the most part, this is unopinionated and will support all use-cases, but in the case of zapping funds out of the zap, WETH will always be unwrapped into ETH before sending it to the user. 

Some notable things to watch out for are:

- Users may receive funds in the zapped asset rather than their original asset as a result of `openShortZap`'s lack of precision compared to the other flows. This is unavoidable because of the way that Hyperdrive's `openShort` logic functions. 
- Tokens can be "wrapped" before the Uniswap swap is executed. This makes it possible to support zaps from rebasing tokens like stETH and eETH into other assets. 
- This contract does not have any of its own events. When I talked to @ryangoree about the original requirements, he said that he didn't think events would be required. We can always revisit this if it turns out that we need them. For now, the only events will be the underlying Uniswap V3 and Hyperdrive events.


contracts/src/zaps/UniV3Zap.sol
```diff
@@ -0,0 +1,937 @@
...
+        // call. This means that the ether transfer should be the result of a
+        // Hyperdrive trade or unwrapping WETH.
+        if (!_reentrancyGuardEntered()) {
+            revert InvalidTransfer();
+        }
+    }
+
+    /// LPs ///
+
+    /// @notice Executes a swap on Uniswap and uses the proceeds to add
```
jrhea, line 89: could you just inherit this comment from the interface?

contracts/src/zaps/UniV3Zap.sol
```diff
@@ -0,0 +1,937 @@
...
+            ERC20(_zapInOptions.swapParams.path.tokenOut()).forceApprove(
+                address(_hyperdrive),
+                proceeds + 1
+            );
+        }
+
+        // Add liquidity using the proceeds of the trade. If the vault shares
+        // token is a rebasing token, the proceeds amount needs to be converted
+        // to vault shares.
+        if (!_hyperdriveOptions.asBase && _zapInOptions.isRebasing) {
```
jrhea, line 138: okay so even if the base token is rebasing there is no need to convert to shares bc it is base. sorry, this confused me for a second so im just typing it out.

contracts/src/zaps/UniV3Zap.sol
```diff
@@ -0,0 +1,937 @@
...
+        // Validate the zap parameters.
+        address proceedsAsset = _validateZapOut(
+            _hyperdrive,
+            _options,
+            _swapParams,
+            _shouldWrap
+        );
+
+        // Take custody of the LP shares.
+        _hyperdrive.transferFrom(
```
jrhea, line 193: i have a question about this. remind me to ask over voices.

contracts/src/zaps/UniV3Zap.sol
```diff
@@ -0,0 +1,937 @@
...
+        // Validate the zap parameters.
+        bool shouldConvertToETH = _validateZapIn(
+            _hyperdrive,
+            _hyperdriveOptions,
+            _zapInOptions
+        );
+
+        // Zap the funds that will be used to add liquidity and approve the pool
+        // to spend these funds.
+        uint256 proceeds = _zapIn(_zapInOptions, shouldConvertToETH);
```
jrhea, line 122: what do we do if tokenIn==tokenOut?

contracts/src/zaps/UniV3Zap.sol
```diff
@@ -0,0 +1,937 @@
...
+    /// @dev Zaps funds into this contract to open positions on Hyperdrive.
+    /// @param _zapInOptions The options for the zap.
+    /// @param _shouldConvertToETH A flag indicating whether or not the proceeds
+    ///        should be converted to ETH.
+    /// @return proceeds The amount of assets that were zapped into this
+    ///         contract.
+    function _zapIn(
+        IUniV3Zap.ZapInOptions memory _zapInOptions,
+        bool _shouldConvertToETH
+    ) internal returns (uint256 proceeds) {
```
jrhea, line 708: what do we do if tokenIn==tokenOut?

test/zaps/uni-v3/CloseLongZap.t.sol
```diff
@@ -0,0 +1,448 @@
...
+        );
+
+        // Execute the zap.
+        IHyperdrive hyperdrive = _hyperdrive; // avoid stack-too-deep
+        uint256 maturityTime = _maturityTime; // avoid stack-too-deep
+        uint256 bondAmount = _bondAmount; // avoid stack-too-deep
+        ISwapRouter.ExactInputParams memory swapParams = _swapParams; // avoid stack-too-deep
+        bool shouldWrap = _shouldWrap; // avoid stack-too-deep
+        bool asBase = _asBase; // avoid stack-too-deep
+        uint256 proceeds = zap.closeLongZap(
```
jrhea, line 401: weird that test coverage doesn't pick up on this

contracts/src/zaps/UniV3Zap.sol
```diff
@@ -0,0 +1,937 @@
...
+            return
+                ILido(_hyperdrive.vaultSharesToken()).getSharesByPooledEth(
+                    _baseAmount
+                );
+        }
+        // If this is a mainnet deployment and the address is the legacy stETH
+        // pool, we have to convert the proceeds to shares manually using Lido's
+        // `getSharesByPooledEth` function.
+        else if (
+            block.chainid == 1 && address(_hyperdrive) == LEGACY_SDAI_HYPERDRIVE
```
jrhea, line 925: I didn't see a specific test for this case - might be a good idea

test/zaps/uni-v3/OpenShortZap.t.sol
```diff
@@ -0,0 +1,678 @@
...
+                sourceAmount: 0.3882e18,
+                shouldWrap: true,
+                isRebasing: false
+            }),
+            10e18, // most of this should be refunded
+            false // as base
+        );
+    }
+
+    /// @notice Ensure that zapping into `openLong` with base succeeds with
```
jrhea, line 310:
```suggestion
    /// @notice Ensure that zapping into `openShort` with base succeeds with
```

test/zaps/uni-v3/OpenShortZap.t.sol
```diff
@@ -0,0 +1,678 @@
...
+                sourceAmount: 1_000e6,
+                shouldWrap: false,
+                isRebasing: true
+            }),
+            10e18, // this should be completely refunded
+            true // as base
+        );
+    }
+
+    /// @notice Ensure that zapping into `openLong` with vault shares
```
jrhea, line 336:
```suggestion
    /// @notice Ensure that zapping into `openShort` with vault shares
```

test/zaps/uni-v3/OpenShortZap.t.sol
```diff
@@ -0,0 +1,678 @@
...
+                sourceAmount: 1_000e6,
+                shouldWrap: false,
+                isRebasing: true
+            }),
+            0,
+            false // as base
+        );
+    }
+
+    /// @notice Ensure that zapping into `openLong` with base succeeds with
```
jrhea, line 368:
```suggestion
    /// @notice Ensure that zapping into `openShort` with base succeeds with
```

test/zaps/uni-v3/OpenShortZap.t.sol
```diff
@@ -0,0 +1,678 @@
...
+                sourceAmount: 1_000e6,
+                shouldWrap: false,
+                isRebasing: false
+            }),
+            0,
+            true // as base
+        );
+    }
+
+    /// @notice Ensure that zapping into `openLong` with vault shares
```
jrhea, line 394:
```suggestion
    /// @notice Ensure that zapping into `openShort` with vault shares
```

test/zaps/uni-v3/CloseLongZap.t.sol
```diff
@@ -0,0 +1,448 @@
...
+                // This will automatically be adjusted down.
+                amountIn: 1_000e18,
+                amountOutMinimum: 0.38e18
+            }),
+            false, // should wrap
+            true // as base
+        );
+    }
+
+    /// @notice Ensure that zapping out of `removeLiquidity` with vault shares
```
jrhea, line 312:
```suggestion
    /// @notice Ensure that zapping out of `closeLong` with vault shares
```

test/zaps/uni-v3/CloseShortZap.t.sol
```diff
@@ -0,0 +1,456 @@
...
+                // This will automatically be adjusted down.
+                amountIn: 1_000e18,
+                amountOutMinimum: 0.38e18
+            }),
+            false, // should wrap
+            true // as base
+        );
+    }
+
+    /// @notice Ensure that zapping out of `removeLiquidity` with vault shares
```
jrhea, line 317:
```suggestion
    /// @notice Ensure that zapping out of `closeShort` with vault shares
```

jrhea (commented):
Great work! I just had a few qs and some nits

contracts/src/zaps/UniV3Zap.sol
jalextowle: Yeah that's right. And then the other two cases are:

1. Non-rebasing and `asBase = true`: The proceeds amount is in base, so no conversion is necessary.
2. Non-rebasing and `asBase = false`: The proceeds amount is in vault shares, so no conversion is necessary.

contracts/src/zaps/UniV3Zap.sol
jalextowle: I'm glad that you bring this up. We definitely can, but I personally don't love that kind of inheritance because it moves the documentation away from the code that it is documenting. It is definitely good for maintainability and documentation duplication, but it's bad for readability IMO.

I think we need to collectively agree on what tradeoff we want to make. This is my personal preference, but there may be strong arguments in the other direction that would change my mind.

contracts/src/zaps/UniV3Zap.sol
jalextowle: Good question. My approach so far was "do nothing and let Uniswap handle it," but I can add an explicit error and failure for this for zapping in and out.

contracts/src/zaps/UniV3Zap.sol
jalextowle: Yeah, good call.

test/zaps/uni-v3/CloseLongZap.t.sol
jalextowle: Yeah it definitely is. Idk why

jrhea (approved):

contracts/src/zaps/UniV3Zap.sol
```diff
@@ -0,0 +1,952 @@
...
+        // If the deposit isn't in ETH, we need to set an approval on Hyperdrive.
+        address tokenOut = _zapInOptions.swapParams.path.tokenOut();
+        if (!shouldConvertToETH) {
+            // NOTE: We increase the required approval amount by 1 wei so that the
+            // pool ends with an approval of 1 wei. This makes future approvals
+            // cheaper by keeping the storage slot warm.
+            ERC20(tokenOut).forceApprove(address(_hyperdrive), proceeds + 1);
+        }
+
+        // Open a long using the proceeds of the trade.
```
mcclurejt, line 455:
```suggestion
        // Open a short using the proceeds of the trade.
```

mcclurejt (approved):
Looks awesome, really great work on the documentation... super digestable

PR #1189, opened 2024-10-11
Moonwell EURC

jalextowle:
# Description

This PR adds testing for the Moonwell EURC yield source. Unlike most of our other yield sources, this yield source has a base decimals of `6` and vault share decimals of `18`. Furthermore, the vault share price is quite low at around ~1e6. With this in mind, some of the tolerances are significantly larger than what we'd expect to see with other yield sources.

PR #1190, opened 2024-10-14
snARS Integration

jalextowle:
# Description

This PR adds the snARS integration. This yield source is a staking vault for the nARS stable coin. Some interesting things about this yield source are that `asBase = true` is not supported on withdrawal since the user will not receive adequate base (nARS is instead locked in an escrow contract, which we don't want to subject our users to).

### DIFF
--- a/deployments.json
+++ b/deployments.json
@@ -1453,41 +1453,6 @@
             "address": "0xFcdaF9A4A731C24ed2E1BFd6FA918d9CF7F50137",
             "timestamp": "2024-09-23T17:24:41.079Z"
         },
-        "ElementDAO ERC4626 Hyperdrive Deployer Coordinator_ERC4626HyperdriveCoreDeployer": {
-            "contract": "ERC4626HyperdriveCoreDeployer",
-            "address": "0xb1c456824cb526a0b57f6cea309785752ec015d7",
-            "timestamp": "2024-10-09T20:36:51.355Z"
-        },
-        "ElementDAO ERC4626 Hyperdrive Deployer Coordinator_ERC4626Target0Deployer": {
-            "contract": "ERC4626Target0Deployer",
-            "address": "0xd171c68eaa3fbd25c1fa34c1a5e63e193df39aa6",
-            "timestamp": "2024-10-09T20:36:58.302Z"
-        },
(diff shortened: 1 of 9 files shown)


deployments.json
```diff
@@ -1453,41 +1453,6 @@
             "address": "0xFcdaF9A4A731C24ed2E1BFd6FA918d9CF7F50137",
             "timestamp": "2024-09-23T17:24:41.079Z"
         },
-        "ElementDAO ERC4626 Hyperdrive Deployer Coordinator_ERC4626HyperdriveCoreDeployer": {
```
mcclurejt, line 1456: are these going to get replaced?

mcclurejt (approved):
LGTM

tasks/deploy/config/base/snars-182day.ts
```diff
@@ -0,0 +1,87 @@
...
+            "HyperdriveFactory",
+            hre.hyperdriveDeploy.deployments.byName(BASE_FACTORY_NAME).address,
+        );
+        return {
+            baseToken: NARS_ADDRESS_BASE,
+            vaultSharesToken: SNARS_ADDRESS_BASE,
+            circuitBreakerDelta: parseEther("0.075"),
+            minimumShareReserves: parseEther("0.001"),
+            minimumTransactionAmount: parseEther("0.001"),
+            positionDuration: parseDuration(SIX_MONTHS),
```
jrhea, line 67: didn't we talk about this being 1 month or 3 months?

hardhat.config.base.ts
```diff
@@ -40,6 +41,7 @@ const config: HardhatUserConfig = {
                     BASE_MORPHO_BLUE_CBETH_USDC_182DAY,
                     BASE_MOONWELL_ETH_182DAY,
                     BASE_STK_WELL_182DAY,
+                    BASE_SNARS_182DAY,
```
jrhea, line 44: if the term is going to be 1 month or 3 months we should rename this

jrhea (changes requested):

jrhea (approved):

jalextowle:
Closed because these changes were incorporated into #1189

PR #1201, opened 2024-10-26
Hyperdrive Matching Engine

jalextowle:
# Description

This PR adds a matching engine that can be used to match traders on Hyperdrive directly. Due to the way it was designed, this matching engine allows a long and short trader to trade directly with each other without needing to worry about whether or not there is enough LP provided liquidity in the pool. The system uses Morpho as a flash lender since Morpho provides zero-cost flash loans on the funds that it holds.

In order for two users to be directly matched, one needs to open a long and the other needs to open a short. Because of the way that Hyperdrive's intra-checkpoint netting system works, a long and a short that have the same size will offset each other. This is the core insight that enables direct matching to work because it means that directly matching a long against a short doesn't impact liquidity if they are sized properly. The way that the users are directly matched follows this flow:

1. The long buyer and short buyer create order intents and approve the matching engine to spend their base tokens.
2. The executor of the match calls `matchOrders` with the order intents and specifies some other configuration parameters, including the amount of liquidity to flash borrow.
3. The requested funds are flash borrowed into the matching engine.
4. The flash borrowed funds are LP'd into Hyperdrive.
5. One of the trades is executed on Hyperdrive.
6. The opposing trade is executed on Hyperdrive.
7. The liquidity is removed from Hyperdrive.
8. Any surplus base and withdrawal shares in the matching engine are sent to the fee recipient. 
9. The flash loan is repaid. 

A few simplifications were made to speed up the development process and reduce complexity:

- `asBase` must be true for both orders and for the add and remove liquidity options. This simplifies the logic and avoids having to deal with edge cases like `addLiquidity` using `asBase` as `true` and `removeLiquidity` using `asBase` as `false`. 
- ETH is not a supported deposit or withdrawal asset. 
- There are no partial fills. This logic is challenging to design for this kind of matching engine because Hyperdrive trades aren't linear. This means that chopping up an order and executing it in pieces could have very different results (and different slippage properties) than executing it in one shot.


contracts/src/interfaces/IHyperdriveMatchingEngine.sol
```diff
@@ -0,0 +1,168 @@
...
+    ///         opinionated to keep the implementation simple.
+    error InvalidSettlementAsset();
+
+    /// @notice Thrown when the signature for an order intent doesn't recover to
+    ///         the expected signer address.
+    error InvalidSignature();
+
+    /// @notice Thrown when the long and short orders don't refer to the same
+    ///         Hyperdrive instance.
+    error MismatchedHyperdrive();
```
mcclurejt, line 41: NIT: Maybe move these to the end? `IHyperdrive` isn't a great reference since it's so broken up, but a common order for interfaces I've seen elsewhere seems to be:
```
events

structs

stateful fn's

getter fn's

errors
```

contracts/src/interfaces/IHyperdriveMatchingEngine.sol
```diff
@@ -0,0 +1,168 @@
...
+
+    /// @notice Emitted when orders are cancelled
+    event OrdersCancelled(address indexed trader, bytes32[] orderHashes);
```
mcclurejt, line 44: Why the array of order hashes as opposed to `longOrderHash` and `shortOrderHash`?

contracts/src/interfaces/IHyperdriveMatchingEngine.sol
```diff
@@ -0,0 +1,168 @@
...
+    function version() external view returns (string memory);
+
+    /// @notice Get the Morpho flash loan provider.
+    /// @return The Morpho contract address.
+    function morpho() external view returns (IMorpho);
+
+    /// @notice Check if an order has been cancelled.
+    /// @param orderHash The hash of the order to check.
+    /// @return True if the order was cancelled.
+    function cancels(bytes32 orderHash) external view returns (bool);
```
mcclurejt, line 116: NIT: imo `isCanceled(..)` seems a bit more straightforward

contracts/src/interfaces/IHyperdriveMatchingEngine.sol
```diff
@@ -0,0 +1,168 @@
...
+    error InvalidSignature();
+
+    /// @notice Thrown when the long and short orders don't refer to the same
+    ///         Hyperdrive instance.
+    error MismatchedHyperdrive();
+
+    /// @notice Emitted when orders are cancelled
+    event OrdersCancelled(address indexed trader, bytes32[] orderHashes);
+
+    /// @notice Emitted when orders are matched
```
mcclurejt, line 46:
```suggestion
    /// @notice Emitted when orders are matched.
```

contracts/src/interfaces/IHyperdriveMatchingEngine.sol
```diff
@@ -0,0 +1,168 @@
...
+
+    /// @notice Thrown when the signature for an order intent doesn't recover to
+    ///         the expected signer address.
+    error InvalidSignature();
+
+    /// @notice Thrown when the long and short orders don't refer to the same
+    ///         Hyperdrive instance.
+    error MismatchedHyperdrive();
+
+    /// @notice Emitted when orders are cancelled
```
mcclurejt, line 43:
```suggestion
    /// @notice Emitted when orders are cancelled.
```

contracts/src/interfaces/IHyperdriveMatchingEngine.sol
```diff
@@ -0,0 +1,168 @@
...
+    /// @return The version string.
+    function version() external view returns (string memory);
+
+    /// @notice Get the Morpho flash loan provider.
+    /// @return The Morpho contract address.
+    function morpho() external view returns (IMorpho);
+
+    /// @notice Check if an order has been cancelled.
+    /// @param orderHash The hash of the order to check.
+    /// @return True if the order was cancelled.
```
mcclurejt, line 115: Assuming false if otherwise? What if the `orderHash` wasn't a valid `orderHash`? Still false?

contracts/src/matching/HyperdriveMatchingEngine.sol
```diff
@@ -0,0 +1,472 @@
...
+import { ECDSA } from "openzeppelin/utils/cryptography/ECDSA.sol";
+import { EIP712 } from "openzeppelin/utils/cryptography/EIP712.sol";
+import { ReentrancyGuard } from "openzeppelin/utils/ReentrancyGuard.sol";
+import { IHyperdrive } from "../interfaces/IHyperdrive.sol";
+import { IHyperdriveMatchingEngine } from "../interfaces/IHyperdriveMatchingEngine.sol";
+import { AssetId } from "../libraries/AssetId.sol";
+import { HYPERDRIVE_MATCHING_ENGINE_KIND, VERSION } from "../libraries/Constants.sol";
+import { FixedPointMath } from "../libraries/FixedPointMath.sol";
+
+// TODO: Document the simplifications that were made.
```
mcclurejt, line 17: Not sure if you meant to do this in this PR

contracts/src/matching/HyperdriveMatchingEngine.sol
```diff
@@ -0,0 +1,472 @@
...
+    using FixedPointMath for uint256;
+    using SafeERC20 for ERC20;
+
+    /// @notice The EIP712 typehash of the OrderIntent struct.
+    bytes32 public constant ORDER_INTENT_TYPEHASH =
+        keccak256(
+            "OrderIntent(address trader,address hyperdrive,uint256 amount,uint256 slippageGuard,uint256 minVaultSharePrice,Options options,uint8 orderType,uint256 expiry,bytes32 salt)"
+        );
+
+    /// @notice The EIP712 typehash of the Options struct.
```
mcclurejt, line 40:
```suggestion
    /// @notice The EIP712 typehash of the IHyperdrive.Options struct.
```

contracts/src/interfaces/IHyperdriveMatchingEngine.sol
```diff
@@ -0,0 +1,168 @@
...
+    function cancels(bytes32 orderHash) external view returns (bool);
+
+    /// @notice Get the EIP712 typehash for the OrderIntent struct.
+    /// @return The typehash.
+    function ORDER_INTENT_TYPEHASH() external view returns (bytes32);
+
+    /// @notice Get the EIP712 typehash for the Options struct.
```
mcclurejt, line 122:
```suggestion
    /// @notice Get the EIP712 typehash for the IHyperdrive.Options struct.
```

contracts/src/matching/HyperdriveMatchingEngine.sol
```diff
@@ -0,0 +1,472 @@
...
+    /// @dev We exclude extra data from the options hashing since it has no
```
mcclurejt, line 41: love this helpful tidbit

contracts/src/matching/HyperdriveMatchingEngine.sol
```diff
@@ -0,0 +1,472 @@
...
+        // to the fee recipient.
+        if (proceeds > lpAmount) {
+            baseToken.safeTransfer(feeRecipient, proceeds - lpAmount);
+        }
+
+        // Approve Morpho Blue to take back the assets that were provided.
+        baseToken.forceApprove(address(morpho), lpAmount);
+    }
+
+    /// @notice Hashes an order intent according to EIP-712
```
mcclurejt, line 238:
```suggestion
    /// @notice Hashes an order intent according to EIP-712.
```

contracts/src/matching/HyperdriveMatchingEngine.sol
```diff
@@ -0,0 +1,472 @@
...
+    /// @param _order The order intent to hash
```
mcclurejt, line 239:
```suggestion
    /// @param _order The order intent to hash.
```

contracts/src/matching/HyperdriveMatchingEngine.sol
```diff
@@ -0,0 +1,472 @@
...
+    /// @return The hash of the order intent
```
mcclurejt, line 240:
```suggestion
    /// @return The hash of the order intent.
```

contracts/src/interfaces/IHyperdriveMatchingEngine.sol
```diff
@@ -0,0 +1,168 @@
...
+        ///      `OpenLong`, this is the minimum output in bonds, and in the
+        ///      case of `OpenShort`, this is the maximum deposit in base.
+        uint256 slippageGuard;
+        /// @dev The minimum vault share price. This protects traders against
+        ///      the sudden accrual of negative interest in a yield source.
+        uint256 minVaultSharePrice;
+        /// @dev The options that configure how the trade will be settled.
+        ///      `asBase` is required to be true, the `destination` is the
+        ///      address that receives the long or short position that is
+        ///      purchased, and the extra data is configured for the yield
```
mcclurejt, line 81: It's mentioned in `HyperdriveMatchingEngine.sol` that the `extraData` param is not used, maybe it's worth mentioning here as well? Reading this gives me the impression that `extraData` is being considered.

contracts/src/matching/HyperdriveMatchingEngine.sol
```diff
@@ -0,0 +1,472 @@
...
+        }
+
+        // Ensure that both orders refer to the same Hyperdrive pool.
+        if (_longOrder.hyperdrive != _shortOrder.hyperdrive) {
+            revert MismatchedHyperdrive();
+        }
+
+        // Ensure that all of the transactions should be settled with base.
+        if (
+            !_longOrder.options.asBase ||
```
mcclurejt, line 414: Since `asBase` has to be true, is `IHyperdrive.Options` only part of the `OrderIntent` struct for its `destination` param? 

Asking because it seems silly to ask a user to pass in a parameter that only has one valid answer... but can also see why it'd be convenient to have the entire options struct when it comes time to `openLong`/`openShort`

contracts/src/matching/HyperdriveMatchingEngine.sol
```diff
@@ -0,0 +1,472 @@
...
+            !_shortOrder.options.asBase ||
+            !_addLiquidityOptions.asBase ||
+            !_removeLiquidityOptions.asBase
+        ) {
+            revert InvalidSettlementAsset();
+        }
+
+        // Ensure that the order's cross. We can calculate a worst-case price
```
mcclurejt, line 422:
```suggestion
        // Ensure that the orders cross. We can calculate a worst-case price
```

contracts/src/matching/HyperdriveMatchingEngine.sol
```diff
@@ -0,0 +1,472 @@
...
+        // for the long and short using the `amount` and `slippageGuard` fields.
+        // In order for the orders to cross, the price of the long should be
+        // equal to or higher than the price of the short. This implies that the
+        // long is willing to buy bonds at a price equal or higher than the
+        // short is selling bonds, which ensures that the trade is valid.
+        if (
+            _longOrder.slippageGuard != 0 &&
```
mcclurejt, line 429: Wanting to clarify my own understanding.... so if the long order doesn't have a slippage guard, the order is valid?

contracts/src/matching/HyperdriveMatchingEngine.sol
```diff
@@ -0,0 +1,472 @@
...
+            _shortOrder.slippageGuard < _shortOrder.amount &&
+            _longOrder.amount.divDown(_longOrder.slippageGuard) <=
+            (_shortOrder.amount - _shortOrder.slippageGuard).divDown(
+                _shortOrder.amount
+            )
```
mcclurejt, line 434: Should this inequality just be `<` instead of `<=`? The way i'm reading it, if both sides are equal, then the long's max buy price equals the shorts min sell price, which should be a match no?

contracts/src/matching/HyperdriveMatchingEngine.sol
```diff
@@ -0,0 +1,472 @@
...
+        ) {
+            revert InvalidMatch();
+        }
+
+        // Hash both orders
```
mcclurejt, line 439:
```suggestion
        // Hash both orders.
```

test/integrations/matching/DirectMatchTest.t.sol
```diff
@@ -0,0 +1,500 @@
...
+        (v, r, s) = vm.sign(
+            HEDGER_PK,
+            matchingEngine.hashOrderIntent(shortOrder)
+        );
+        shortOrder.signature = abi.encodePacked(r, s, v);
+
+        // Update the short order's extra data after signing. This simulates
+        // what the DFB UI will do to ensure an up-to-date quote. Since the
+        // extra data isn't baked into the hash, this will not cause issues with
+        // signature verification.
```
mcclurejt, line 288: NIT: This tidbit about `extraData`'s relevance to the DFB UI might be a nice addition to the interface... seeing that it wasn't included in the hash made me think it wasn't relevant or was being ignored.

test/integrations/matching/DirectMatchTest.t.sol
```diff
@@ -0,0 +1,500 @@
...
+        shortOrder.options.extraData = hex"deadbeefbabefade6660";
+
+        // Match the two counterparties using flash loans.
+        matchingEngine.matchOrders(
+            // long order
+            longOrder,
+            // short order
+            shortOrder,
+            // flash loan amount
+            18_500_000e6,
```
mcclurejt, line 298: how did you determine this amount?

test/units/matching/HyperdriveMatchingEngineTest.t.sol
```diff
@@ -0,0 +1,1535 @@
...
+    ) external view returns (bytes4) {
+        if (shouldVerifySignature) {
+            return this.isValidSignature.selector;
+        } else {
+            return bytes4(0);
+        }
+    }
+}
+
+// FIXME: Fully Natspec this whole test suite.
```
mcclurejt, line 88: Dangling FIXME

test/units/matching/HyperdriveMatchingEngineTest.t.sol
```diff
@@ -0,0 +1,1535 @@
...
+                IHyperdrive.Options({
+                    asBase: true,
+                    destination: address(matchingEngine),
+                    extraData: ""
+                }),
+                // fee recipient
+                celine,
+                true
+            );
+            vm.revertTo(snapshotId);
```
mcclurejt, line 967:
```suggestion
```

mcclurejt (commented):
Awesome work dude! As a whole it looks awesome.

Lots of nits + questions for my own understanding...

Want to give it one more pass before approve just to make sure I'm not missing anything :)

contracts/src/matching/HyperdriveMatchingEngine.sol
jalextowle: I just left this here as a reminder to include the simplifications in the PR description

contracts/src/matching/HyperdriveMatchingEngine.sol
jalextowle: Yeah because this implies a worst-case price that is infeasible, which will match against any order on the short side. If the long worst-case price is infinite (or even just 1) or the short worst-case price is 0, the orders will trivially cross.

I don't have a problem with a user specifying a slippage tolerance of 0, but it's not recommended.

contracts/src/matching/HyperdriveMatchingEngine.sol
jalextowle: Yeah, that's a good point

test/integrations/matching/DirectMatchTest.t.sol
jalextowle: It's a manual process at the moment. Increasing this decreases slippage => lower realized rate. I had a rate that I wanted to get below (5.11% for the short), so the process of finding this number was just trial and error.

contracts/src/interfaces/IHyperdriveMatchingEngine.sol
jalextowle: I'm not sure what you mean. This includes the hashes of all of the orders that were cancelled. It could be 1, 2, or 100 orders, so I'm not sure how this would fit into a long and short order hash structure.

contracts/src/interfaces/IHyperdriveMatchingEngine.sol
jalextowle: Good call

contracts/src/interfaces/IHyperdriveMatchingEngine.sol
jalextowle: I'll add something about what false means.

What would make an order hash invalid? This is just a mapping from `bytes32` to `bool`. In theory, there are orders that hash to every `bytes32` value, we just don't know what they are because of the hash function's pre-image resistance.

test/units/matching/HyperdriveMatchingEngineTest.t.sol
jalextowle: I add these to avoid issues if we add more cases. I'm going to leave this.

contracts/src/interfaces/IHyperdriveMatchingEngine.sol
mcclurejt: Okay awesome! Not a huge deal and didn't really have a great source of truth on this one, solidity docs to the rescue!

contracts/src/interfaces/IHyperdriveMatchingEngine.sol
mcclurejt: When I wrote this I hadn't read the implementation yet, totally get you on the hash-validity front

contracts/src/matching/HyperdriveMatchingEngine.sol
mcclurejt: cool makes sense

contracts/src/interfaces/IHyperdriveMatchingEngine.sol
mcclurejt: Derp sorry this definitely doesn't make sense... wrote this before reading the rest of the PR and forgot to remove.

mcclurejt (approved):
Looks great!

PR #1208, opened 2024-10-31
Morpho mainnet WBTC/USDC deployment

MazyGio:
# Resolved Issues

# Description

# Review Checklists

be removed.

## [[Reviewer Name]]

- [ ] **Tokens**
              token spend?
- [ ] **Reentrancy**
- [ ] **Gas Optimizations**
          memory variables?
          issues?
- [ ] **Visibility**
- [ ] **Math**
          catch underflows?
- [ ] **Testing**


tasks/deploy/config/mainnet/morpho-blue-wbtc-usdc-182day.ts
```diff
@@ -0,0 +1,138 @@
...
+            morpho: "0xBBBBBbbBBb9cC5e90e3b3Af64bdAF62C37EEFFCb" as `0x${string}`,
+            collateralToken: WBTC_ADDRESS_MAINNET,
+            oracle: "0xDddd770BADd886dF3864029e4B377B5F6a2B6b83" as `0x${string}`,
+            irm: "0x870aC11D48B15DB9a138Cf899d20F13F79Ba00BC" as `0x${string}`,
+            lltv: BigInt("860000000000000000"),
+        },
+    ],
+);
+
+export const MAINNET_MORPHO_BLUE_WBTC_USDC_182DAY: HyperdriveInstanceConfig<"MorphoBlue"> =
```
jrhea, line 68: in other PRs we started adding a comment with a link to the market so that we could easily find the source of the rate we are initializing to

jrhea (changes requested):

mcclurejt (approved):
Nice Work!

tasks/deploy/config/mainnet/morpho-blue-wbtc-usdc-182day.ts
```diff
@@ -0,0 +1,140 @@
...
+            hre.hyperdriveDeploy.deployments.byName(
+                MAINNET_MORPHO_BLUE_COORDINATOR_NAME,
+            ).address,
+        deploymentId: keccak256(
+            toBytes(MAINNET_MORPHO_BLUE_WBTC_USDC_182DAY_NAME),
+        ),
+        salt: toBytes32("0x42080085"),
+        extraData: morphoBlueParameters,
+        contribution: CONTRIBUTION,
+        // NOTE: The latest variable rate on the Morpho Blue market is 2.93% APY:
```
jrhea, line 82: the latest rate isn't 2.93%. is this a copy paste thing?

MazyGio: Yes it is totally a copy paste thing. Updated!

jrhea (approved):

PR #1209, opened 2024-11-01
Morpho cbBTC/USDC Mainnet Integration.

sentilesdal:
Adds the Hyperdrive integration for the Morpho cbBTC/USDC market for mainnet.


mcclurejt (approved):
LGTM, would like to get a sanity check on the poolconfig from @jrhea or @jalextowle before full go-ahead

jrhea (approved):
The params look good. The only thing I would suggest is re-checking the APY before we deploy. It has gone up quite a bit this past week

PR #1212, opened 2024-11-11
Fix hyperdrivetypes pypi upload error. Includes version bump to 1.0.20.11

slundqui:
Uploading to pypi breaks when calling the uploading workflow from another workflow (https://github.com/pypa/gh-action-pypi-publish/issues/299). Fixing by consolidating workflow into one, and only upload when pyproject changes are made (removed upload on tag).

Minor fixes:
- Avoid running test hyperdrivetypes twice in PR


dpaiton (approved):
awesome!

PR #1220, opened 2024-11-20
Adds `mint` and `burn` functions

jalextowle:
# Description

This PRs adds two functions that together allow traders to completely bypass Hyperdrive's AMM system while still allowing them to open long and short positions. In order to bypass the AMM and avoid the use of LP funds, `mint` and `burn` provide the low-level plumbing required for traders to be matched directly against each other. These functions provide a way to directly mint longs and shorts equal in value to a given amount of base and directly burn longs and shorts for the amount of value underlying the bonds. 

## Mint

The operation of minting takes in a share deposit $\Delta z$ and outputs an amount of bonds $\Delta y$ that were minted. The bonds are split between long and short tokens that are both minted with amounts equal to $\Delta y$. 

One of the goals for the `mint` function is to interoperate with the existing accounting mechanisms. Long and short tokens that are created through calling the `mint` function must be able to be closed by calling `closeLong` or `closeShort` without negatively impacting the system invariants like solvency. Since offsetting positions are created, the price shouldn't change when positions are minted. 

To mint matched long and short positions, a few things must be paid for:

- Underlying value of the bond in shares, equal to $\tfrac{\Delta y}{c}$
- Pre-paid variable interest in shares, equal to $\tfrac{\max(c, c_0) - c_0}{c_0 \cdot c} \cdot \Delta y$ (the max component ensures that this is never negative)
- Prepaid flat fee, equal to $\tfrac{\phi_f}{c} \cdot \Delta y$
- Governance fees, by convention equal to $2 \cdot \tfrac{\phi_g \cdot \phi_f}{c} \cdot \Delta y$

Given all of this information, we can set up the equality relating the shares deposited to the bond amount:

$$
\begin{aligned}
\Delta z & = \tfrac{\Delta y}{c} + \tfrac{\max(c, c_0) - c_0}{c_0 \cdot c} \cdot \Delta y + \tfrac{\phi_f}{c} \cdot \Delta y + 2 \cdot \tfrac{\phi_g \cdot \phi_f}{c} \cdot \Delta y \\
& \implies \\
\Delta y & = \frac{\Delta z}{\tfrac{1}{c} + \tfrac{\max(c, c_0) - c_0}{c_0 \cdot c} + \tfrac{\phi_f}{c} + 2 \cdot \tfrac{\phi_g \cdot \phi_f}{c}} \\ 
& \implies \\
\Delta y & = \frac{c \cdot \Delta z}{\tfrac{\max(c, c_0)}{c_0} + \phi_f + 2 \cdot \phi_g \cdot \phi_f}
\end{aligned}
$$

Since the trade isn't happening on the curve and conceptually the long and short trades are perfectly matched, the only accounting updates the need to be made is to update the average maturity times, the longs outstanding, the shorts outstanding, and the amount of governance fees that have accrued.  

## Burn

`burn` does the opposite of `mint`. It takes in $\Delta y$ long and short positions with the same maturity and redeems them for the amount of capital underlying the position. Similar to the `mint` function, `burn` should be able to interoperate seamlessly with positions opened by `openLong` and `openShort`. 

When positions are burned, several things must be accounted for:

1. The underlying value of the bonds in shares, equal to $\tfrac{c_1}{c_0 \cdot c} \cdot \Delta y$
2. The flat fee pre-paid by the short, equal to $\tfrac{\phi_f}{c} \cdot \Delta y$.
3. The flat fees owed to the LPs. This shouldn't be required, but it's a hard component to get around because of the way that checkpointing automatically closes positions and applies the flat fee. This is equal to $2 \cdot \tfrac{\phi_f \cdot t_m}{c}$.
4. The governance fees paid when burning the positions, equal by convention to $2 \cdot \tfrac{\phi_g \cdot \phi_f}{c} \cdot \Delta y$.

The proceeds in shares is given by:

$$
\Delta z = \tfrac{c_1}{c_0 \cdot c} \cdot \Delta y + \tfrac{\phi_f}{c} \cdot \Delta y - 2 \cdot \tfrac{\phi_f \cdot t_m}{c} - 2 \cdot \tfrac{\phi_g \cdot \phi_f}{c} \cdot \Delta y
$$

The share and bond reserves aren't updated since the long and short trades perfectly cancel out. The only state changes that must be made are updating the average maturity times, the longs outstanding, the shorts outstanding, and the amount of governance fees accrued. 

# Todos

- [ ] `mint` and `burn`
   - [x] Write the `mint` function.
   - [x] Write unit tests for `mint`.
   - [x] Write integration tests for `mint`.
   - [x] Write the `burn` function.
   - [x] Write unit tests for `burn`.
   - [x] Write integration tests for `burn`.
   - [x] Ensure that negative interest and zombie interest are properly
      - [x] Add a negative interest test for `mint`.
      - [x] Zombie interest
   - [x] Add tests in the remaining major test suites.
      - [ ] Look into why the LP share tolerance needs to be 1e10 in the
            `test_lp_withdrawal_pair_redemption`.
   - [ ] Add `mint` and `burn` to the instance test framework.
   - [ ] Fix the code size.
   - [ ] Wire this up with the agent0 fuzzing framework


contracts/src/internal/HyperdrivePair.sol
```diff
@@ -0,0 +1,308 @@
...
+            // underestimates the bond amount.
+            (ONE +
+                (_vaultSharePrice - _openVaultSharePrice).divUp(
+                    _openVaultSharePrice
+                ) +
+                _flatFee +
+                2 *
+                _flatFee.mulUp(_governanceLPFee))
+        );
+        // FIXME: What should the rounding be here?
```
jrhea, line 301: Seems to me that rounding down what you want to do to be conservative

contracts/src/internal/HyperdrivePair.sol
```diff
@@ -0,0 +1,308 @@
...
+        uint256 longAssetId = AssetId.encodeAssetId(
+            AssetId.AssetIdPrefix.Long,
+            maturityTime
+        );
+        uint256 shortAssetId = AssetId.encodeAssetId(
+            AssetId.AssetIdPrefix.Short,
+            maturityTime
+        );
+        _mint(longAssetId, _options.longDestination, bondAmount);
+        _mint(shortAssetId, _options.shortDestination, bondAmount);
```
jrhea, line 118: In what situation would the long and short destination be different?

contracts/src/internal/HyperdrivePair.sol
```diff
@@ -0,0 +1,308 @@
...
+
+        // Enforce the minimum vault share price.
+        if (vaultSharePrice < _minVaultSharePrice) {
+            revert IHyperdrive.MinimumSharePrice();
+        }
+
+        // Perform a checkpoint.
+        uint256 latestCheckpoint = _latestCheckpoint();
+        uint256 openVaultSharePrice = _applyCheckpoint(
+            latestCheckpoint,
```
Sean329, line 89: @jalextowle Quick question: why here the `latestCheckpoint` gets passed in as the 1st input param? The `_applyCheckpoint(uint256 _checkpointTime,uint256 _vaultSharePrice,uint256 _maxIterations,bool _isTrader)` says "@param _checkpointTime The time of the checkpoint to create.", so intuitively shouldn't it be `block.timestamp` or something?

contracts/src/internal/HyperdrivePair.sol
```diff
@@ -0,0 +1,308 @@
...
+        //
+        // sharesDeposited * vaultSharePrice = (
+        //    bondAmount + bondAmount * (c - c0) / c0 + bondAmount * flatFee +
+        //    2 * bondAmount * flatFee * governanceFee
+        // )
+        //
+        // This implies that
+        //
+        // bondAmount = shareDeposited * vaultSharePrice / (
+        //     1 + (c - c0) / c0 + flatFee + 2 * flatFee * governanceFee
```
Sean329, line 285: @jalextowle 
Informational: This part `1 + (c - c0) / c0` if we further shrink it to just `c/c0`, then in the calculation we will mitigate the potential underflow risk of `c<c0`. Although I'm not sure what's the likelihood of `c<c0` happening in reality, but I think using `c/c0` is always safer. Thoughts?

contracts/src/internal/HyperdrivePair.sol
jalextowle: We talked about this offline

contracts/src/internal/HyperdrivePair.sol
jalextowle: Good call!

contracts/src/external/HyperdriveTarget4.sol
```diff
@@ -86,6 +88,55 @@ abstract contract HyperdriveTarget4 is
             );
     }
 
+    /// Pairs ///
+
+    // FIXME: Where does this fit?
```
jrhea, line 93: can you delete this?

contracts/src/external/HyperdriveTarget4.sol
```diff
@@ -86,6 +88,55 @@ abstract contract HyperdriveTarget4 is
...
+    function mint(
+        uint256 _amount,
+        uint256 _minOutput,
+        uint256 _minVaultSharePrice,
+        IHyperdrive.PairOptions calldata _options
+    ) external payable returns (uint256 maturityTime, uint256 bondAmount) {
+        return _mint(_amount, _minOutput, _minVaultSharePrice, _options);
+    }
+
+    // FIXME: Where does this fit?
```
jrhea, line 119: same here

contracts/src/internal/HyperdrivePair.sol
```diff
@@ -0,0 +1,564 @@
...
+
+        // If the positions haven't matured, apply the accounting updates that
+        // result from closing the pair to the reserves.
+        if (block.timestamp < _maturityTime) {
+            // Apply the state changes caused by burning the offsetting longs and
+            // shorts.
+            //
+            // NOTE: Since the spot price doesn't change, we don't update the
+            // weighted average spot price in this transaction.
+            _applyBurn(_maturityTime, _bondAmount, flatFee, governanceFee);
```
jrhea, line 218: sorry for the dumb question here, but doesn't the flat fee impact the share reserves? won't that cause a small change in the spot price?

jrhea: eh nevermind. zeta changes too so that should cancel it out.

test/integrations/hyperdrive/BurnTest.t.sol
```diff
@@ -0,0 +1,294 @@
...
+                    .mulUp(hyperdrive.getPoolConfig().fees.flat)
+                    .mulDown(
+                        ONE - hyperdrive.calculateTimeRemaining(maturityTime)
+                    )
+                    .mulDown(
+                        ONE - hyperdrive.getPoolConfig().fees.governanceLP
+                    ),
+            1e7
+        );
+        assertLe(hyperdrive.idle(), baseToken.balanceOf(address(hyperdrive)));
```
jrhea, line 207: nice

contracts/src/internal/HyperdrivePair.sol
```diff
@@ -0,0 +1,564 @@
...
+                true
+            );
+        }
+
+        // Burn the longs and shorts that are being closed.
+        uint256 longAssetId = AssetId.encodeAssetId(
+            AssetId.AssetIdPrefix.Long,
+            _maturityTime
+        );
+        _burn(longAssetId, msg.sender, _bondAmount);
```
Sean329, line 195: Hmmm. when new positions are minted, they are minted to the destinations of the long & short traders; but why when burning the positions we burn it from the msg.sender?
If it's the execution is via the MatchEngine, then the call stack could be:
The 3rd party matcher => MatchEngine => Hyperdrive =>(delegate to) target4 => HyperdrivePair , so the msg.sender will be the Hyperdrive instance address....I didn't get it how does this mapping change work?

contracts/src/internal/HyperdrivePair.sol
Sean329: Uh nvm... it did get updated as expected. I didn't see that line earlier.

contracts/src/internal/HyperdrivePair.sol
Sean329: Got it now. It's a generalized design for burning positions instead of designed particular for the match engine.
So for the match engine to use it, the contract needs to take custody of the traders' positions first, before calling the `burn`.

jrhea (approved):
🔥

test/utils/InstanceTest.sol
```diff
@@ -3060,6 +3060,942 @@ abstract contract InstanceTest is HyperdriveTest {
...
+            baseProceeds,
+            true,
+            totalBaseSupplyBefore,
+            totalShareSupplyBefore,
+            bobBalancesBefore,
+            hyperdriveBalancesBefore
+        );
+    }
+
+    // FIXME
```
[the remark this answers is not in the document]
jalextowle: They should all be fixed now

contracts/src/internal/HyperdrivePair.sol
jalextowle: Sweet!

PR #1224, opened 2024-12-05
Deploy scripts for `UniV3Zap` and the matching engine

jalextowle:
# Description

This PR adds deployment scripts for the zap and matching engine contracts.


tasks/deploy/lib/constants.ts
```diff
@@ -3,11 +3,21 @@ import { Address } from "viem";
 export const ETH_ADDRESS =
     "0xEeeeeEeeeEeEeeEeEeEeeEEEeeeeEeeeeeeeEEeE" as Address;
 
+// The salt for the Create 3 deployment of the Hyperdrive Matching Engine. This will
+// generate the address "0x6662b6e771facd61e33ccafdc23be16b4ead0666".
```
mcclurejt, line 7: 😈

tasks/deploy/hyperdrive-matching-engine.ts
```diff
@@ -0,0 +1,92 @@
...
+            let tc = await viem.getTestClient();
+            await tc.setBalance({
+                value: parseEther("1"),
+                address: CREATE_X_FACTORY_DEPLOYER,
+            });
+            let wc = await viem.getWalletClient(CREATE_X_FACTORY_DEPLOYER);
+            await wc.sendRawTransaction({
+                serializedTransaction: CREATE_X_PRESIGNED_TRANSACTION,
+            });
+        }
```
mcclurejt, line 47: will this revert if the CreateX factory already exists? wondering if we should have a check here...

also seeing this logic block in a couple places, wdyt of consolidating into a shared function?

mcclurejt (approved):
Really awesome stuff, minor nits but overall well done!

tasks/deploy/hyperdrive-matching-engine.ts
jalextowle: Good catch!

PR #1225, opened 2024-12-09
WIP PR -- Hyperdrive Matching Engine V2 on top of mint() & burn()

### LINKED ISSUE #1222: Hyperdrive Matching Engine
# Hyperdrive Matching Engine

## Background

In order to support large fixed borrow transactions, it is important that Hyperdrive can support the direct matching of long and short traders. #1201 added an extension contract that makes direct matching possible by using flash loans to add liquidity right before the trades are executed, but this approach has several issues. First, the orders must line up perfectly to ensure that the flash loan can be repaid, and these parameters are quite brittle. If even a small trade is executed before the orders are matched, the add liquidity amount will need to be adjusted, which is currently a manual and time consuming process. The second issue with the system is that it relies on flash loan liquidity. This places an upper-bound on the amount of volume that can feasibly flow through the system. The third issue is that Hyperdrive trades aren't linear, so the current matching engine system isn't set up to support partial fills. 

We can solve all of these issues by making it possible for the matching engine to circumvent the AMM logic. Engineering is underway for `mint` and `burn` functions in #1220 that will make it easier to develop new execution mechanisms for Hyperdrive. `mint` takes in base or vault shares that will underly a set of bonds. Once the `bondAmount` has been calculated, `mint` will create an equal amount of long and short tokens in the current checkpoint and send the proceeds to a `longDestination` and `shortDestination`. `burn` does the opposite. Given an equal amount of long and short tokens in the same checkpoint, `burn` will release the capital underlying the bonds. Using these functions, we can build a matching engine that is simpler and also more powerful. 

Our new matching engine should plug into `mint` and `burn` to support trades of multiple different types. Traders should be able to post their intent to buy or sell either the long or the short position in different checkpoints. Different combinations of these orders can be matched to give a pair of traders what they each want.

## Requirements

The goal of this task is to build a matching engine with the following requirements:

1. A simple schema for order intents that specifies:
    - The size of the trade in bonds
    - The price or fixed rate the trader should receive
    - The position they are buying or selling (long or short)
    - The asset they are using to buy or sell (base or vault shares)
    - The order's expiry timestamp.
    - The Hyperdrive pool
    - The direction of the trade (buy or sell)
2. Order validation logic that verifies:
    - The orders signature. This should support EOA signatures and EIP1271 signatures.
    - If the order has expired or not.
    - If the order has been cancelled or filled.
3. A function to cancel orders
4. A function to match orders
    - This should support all the different types of matches (buy long <> sell long, buy short <> sell short, buy long <> buy short, and sell long <> sell short)
    - Supports partial matches
    - Supports ETH and ERC20 tokens
    - Think about how to support different combintations of `asBase = true` and `asBase = false`.

# Examples
     
## Ex. Imperfect Match (price matches, size doesn't)

Order 1: Alex wants to buy 1000 longs for 990 USDC
Order 2: Sean wants to buy 500 shorts for 5 USDC
     
We will partially fill Alex's order and fully fill Sean's order.

Alex will get 500 longs for 495 USDC
Sean will get 500 shorts for 5 USDC

495 + 5 = 500 -> mint will mint 500 longs and 500 shorts

### Ex 1. Feasible

Order 1: Alex wants to buy 1000 longs for 990 USDC
Order 2: Sean wants to buy 1000 shorts for 11 USDC

990 + 11 > 1000, they can both be matched

Alex gets 1000 longs for 990 USDC
Sean gets 1000 shorts for 11 USDC

There is 1 leftover USDC. This can go to either Alex, Sean, or the person that calls `match`

### Ex 2. Infeasible

Order 1: Alex wants to buy 1000 longs for 990 USDC
Order 2: Sean wants to buy 1000 shorts for 9 USDC

This doesn't work. 990 + 9 = 999 < 1000, so we don't have enough money to mint the bonds

990 * (999 / 1000) = 989.01
9 * (999 / 1000) = 8.991

989.01 + 8.991 = 998.001 < 999

## Different types of orders

The following situations shows the different kinds of orders that can be matched. Some types of matches require each trader to have USDC whereas other types of matches require one or both traders have long or short positions.
 
Order 1: Alex wants 1000 longs for 990 USDC
Order 2: Sean wants 1000 shorts for 10 USDC

Order 1: Alex wants 1000 longs for 990 USDC
Order 2: Sean wants 990 USDC for 1000 longs 

Order 1: Alex wants 1000 shorts for 10 USDC
Order 2: Sean wants 10 USDC for 1000 shorts 

Order 1: Alex wants 990 USDC for 1000 longs
Order 2: Sean wants 10 USDC for 1000 shorts

Sean329:
resolves #1222 

The `matchOrders` function will be called by a 3rd party when he finds two hanging orders can actually be matched. So, both of these 2 orders are actually the maker, and the surplus goes to the 3rd party caller (frontrun competition possible). 

The `fillOrders` function will be called by the taker directly. The take does not need to sign a full `orderIntent`. He only grabs the maker's `orderIntent` and provide some other particular inputs along to the function call.

Checklist before finalization:
1. the contract needs to support `asBase==false` as the fund token as well;
2. apply the style: https://github.com/delvtech/hyperdrive/blob/main/STYLE_GUIDE.md.


contracts/src/matching/HyperdriveMatchingEngine.sol
```diff
@@ -1,464 +1,347 @@
...
-                _shortOrder,
-                _addLiquidityOptions,
-                _removeLiquidityOptions,
-                _feeRecipient,
-                _isLongFirst
-            )
-        );
+            // Calculate the amount of base tokens to transfer based on the bondMatchAmount
+            uint256 baseTokenAmountOrder1 = order1.amount.mulDivDown(bondMatchAmount, order1.slippageGuard);
+            uint256 baseTokenAmountOrder2 = order2.slippageGuard.mulDivDown(bondMatchAmount, order2.amount);
```
jalextowle, line 99: I have a few thoughts/questions about this logic:

1. What is `slippageGuard` doing now? Aren't we always going to enforce that the order is executed at the price specified by `order.amount / order.bondAmount`? It seems to me like we could remove it, but I might be missing something. 
2. We'll need to support both base and vault shares deposits. This is something that we'll probably want to think more about because I think that it could end up being a fairly complicated issue, but I want to bring this up now. `mint` expects the deposit to be made in either base or vault shares, but the best experience would allow the two makers to specify their deposit in either base or vault shares. Let me know if you want to talk more about this. We might end up needing to change the order schema.

contracts/src/matching/HyperdriveMatchingEngine.sol
```diff
@@ -468,4 +351,89 @@ contract HyperdriveMatchingEngine is
...
+        else {
+            revert InvalidOrderCombination();
+        }
+            
+        matchBondAmount = _order1BondAmount < _order2BondAmount ? 
+            _order1BondAmount : 
+            _order2BondAmount;
+    }
+
+    function _handleOpenLongShort(
```
jalextowle, line 389: What about calling this `mint` or `mintBonds`?

contracts/src/matching/HyperdriveMatchingEngine.sol
```diff
@@ -1,464 +1,347 @@
...
 
-        // If the long should be executed first, execute the long and then the
-        // short.
-        if (isLongFirst) {
-            _openLong(hyperdrive, baseToken, longOrder);
-            _openShort(hyperdrive, baseToken, shortOrder);
+        //TODOs
+        else if (_order1.orderType == OrderType.CloseLong && _order2.orderType == OrderType.CloseShort) {
+            // Case 2: Long + Short closing using burn()
+            _handleCloseLongShort(_order1, _order2, matchAmount, baseToken, hyperdrive);
```
jalextowle, line 143: What about calling this `_handleBurn`?

contracts/src/matching/HyperdriveMatchingEngine.sol
```diff
@@ -1,464 +1,347 @@
...
+                _order1, 
+                _order2, 
+                baseTokenAmountOrder1, 
+                baseTokenAmountOrder2, 
+                bondMatchAmount, 
+                baseToken, 
+                hyperdrive);
+            
+            // Update order bond amount used again to be accurate
+            orderBondAmountUsed[order1Hash] += bondAmount - bondMatchAmount;
```
jalextowle, line 121: Shouldn't this just be:

```suggestion
            orderBondAmountUsed[order1Hash] += bondAmount;
```

contracts/src/matching/HyperdriveMatchingEngine.sol
```diff
@@ -1,464 +1,347 @@
...
+            orderBondAmountUsed[order2Hash] += bondAmount - bondMatchAmount;
```
jalextowle, line 122: Same here. Shouldn't this be?

```suggestion
            orderBondAmountUsed[order2Hash] += bondAmount;
```

contracts/src/matching/HyperdriveMatchingEngine.sol
jalextowle: I see. This is correcting for what was done earlier. Is there a reason not to just do this once? The `bondAmount` will probably be lower than the `bondMatchAmount`, so I think that this could fail due to a lack of liquidity.

Sean329: Hi Alex, in lines# 90 and 91 I've already updated the mapping like
```
        orderBondAmountUsed[order1Hash] += bondMatchAmount;
        orderBondAmountUsed[order2Hash] += bondMatchAmount;
```
And, I do this here before the execution really goes into the `if(xxx){_handle...}` blocks mainly to follow the CEI pattern to prevent reentrancy. I know there is a reentrancy guard modifier for this function, but the guard won't for cross-contract reentrancy. So I tend to use both the guard and the CEI patter. Therefore, if we have the Line#90 and 91, then later I will need something like this `orderBondAmountUsed[order1Hash] += bondAmount - bondMatchAmount;` to revise the update (account for the differences, and ofc the real `bondAmount ` must always be larger than `bondMatchAmount ` as the `mint` checks the min output. So there shouldn't be underflow risk here)

contracts/src/matching/HyperdriveMatchingEngine.sol
Sean329: check my comments above

contracts/src/matching/HyperdriveMatchingEngine.sol
Sean329: Hey Alex, here are my answers:
1. For the 1st question, I kept using the vars of `_order1.amount` and `_order1.slippageGuard` only because I'm reusing the interface `IHyperdriveMatchingEngine` designed for the V1MatchingEngine. I try to do as little modification as possible to that interface. The `orderIntent` defined in the interface has these two vars named so, and always one of them refers to the "_fundAmount_" and another var refers to the "_bondAmount_". As for which one is exactly referring to what, it depends on the order types. You put some explanations in the Natspec already, and I summarize them as the followings:

openLong: `amount` is the amount of base to deposit; `sllippageGuard` is the min output in bonds
openShort: `amount` is the amount of bonds to short; `slippageGuard` is the max deposit in base
closeLong: `amount` is the amount of bonds to close; `slippageGuard` is the min output in base
closeShort: `amount` is the amount of base to receive; `slippageGuard` is the max bond to close

As you can see it above, based on different order types, sometimes the `amount` refers to _fundAmount_ and sometimes it refers to _bondAmount_, and `slippageGuard` takes the other role. If we do not have to keep the interface consistent as much as we can, then I'd be more happy to use names of `fundAmount` and `bondAmount` to replace `amount` and `slippageGuard`. We should discuss about this sometime.

Sean329: 2. For the 2nd question, yes I know we should also support vault shares as the fund. And I plan to discuss with you later on that part. Using the base is simpler to code so I'm only working on the `asBase==true` situation for now. Code change to support vault shares as the fund is a TODO after we have solved all puzzles for `asBase==true` situation.

contracts/src/matching/HyperdriveMatchingEngine.sol
Sean329: `_handleBurn` sounds nice.

contracts/src/matching/HyperdriveMatchingEngine.sol
Sean329: > I see. This is correcting for what was done earlier. Is there a reason not to just do this once? The `bondAmount` will probably be lower than the `bondMatchAmount`, so I think that this could fail due to a lack of liquidity.

Can it be? The `bondMatchAmount` is passed into the `mint` function call as the min output guard, so can the output of `bondAmount` still be lower than `bondMatchAmount`? I thought it couldn't otherwise `mint` fails?

contracts/src/matching/HyperdriveMatchingEngine.sol
jalextowle: 1. I’d prefer us to update the naming. Don’t worry about maintaining compatibility with the old matching engine’s interface. Let’s treat this as a fresh matching engine since it will work significantly differently and will be in use much longer. I like `fundAmount` and `bondAmount`.
2. That makes sense. Works for me, I just wanted to make sure that I mentioned that we will need support for this.

I’m wondering if we will want a slippage amount since the cost of minting at a given fixed rate is variable for the short (since the short is required to pre-pay the flat fee and pre-paid interest). We could consider this as part of the fixed rate that the short pays, but I don’t think this is quite right in the case of the pre-paid variable interest since this is repaid at maturity.

I like your design decision to keep things simple, but we will want to make sure that the frontend or data pipelines can easily extract the fixed rate (or equivalently, the spot price) that each trader receives from the trade. This should be able to account for things like the pre-paid variable interest. There is a good chance that this is handled by the vault share price being in the `Mint` event, but I want to call this out explicitly.

contracts/src/matching/HyperdriveMatchingEngine.sol
jalextowle: I need to think more about this second point. It’s late here, but I’ll take another look in the morning.

contracts/src/matching/HyperdriveMatchingEngine.sol
```diff
@@ -468,4 +351,89 @@ contract HyperdriveMatchingEngine is
...
+            uint256 _order1BondAmount = _order1.amount - orderBondAmountUsed[_order1Hash];
+            uint256 _order2BondAmount = _order2.slippageGuard - orderBondAmountUsed[_order2Hash];
+        }
+        else {
+            revert InvalidOrderCombination();
+        }
+            
+        matchBondAmount = _order1BondAmount < _order2BondAmount ? 
+            _order1BondAmount : 
+            _order2BondAmount;
```
jalextowle, line 386: Nit: We could just use `_order1BondAmount.min(order2BondAmount)`.

contracts/src/matching/HyperdriveMatchingEngine.sol
```diff
@@ -468,4 +351,89 @@ contract HyperdriveMatchingEngine is
             revert InvalidDestination();
         }
     }
-}
+
+    function _calculateMatchBondAmount(
```
jalextowle, line 355: Does this compile? I see that we're using `_calculateBondMatchAmount` above.

contracts/src/matching/HyperdriveMatchingEngine.sol
```diff
@@ -468,4 +351,89 @@ contract HyperdriveMatchingEngine is
...
+            uint256 _order2BondAmount = _order2.slippageGuard - orderBondAmountUsed[_order2Hash];
+        }
+        else {
+            revert InvalidOrderCombination();
+        }
+            
+        matchBondAmount = _order1BondAmount < _order2BondAmount ? 
+            _order1BondAmount : 
+            _order2BondAmount;
+    }
```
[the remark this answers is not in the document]
jalextowle: It would be good to add some test cases before putting this up for review. I'd like to see some tests that validate that the matching can be done successfully and some tests that stress the other cases.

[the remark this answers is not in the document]
jalextowle: I don't like the fact that this implementation doesn't guarantee the amount of bonds minted to the long and short will be what they specified in the order. Ideally, they would always receive exactly that amount of bonds. I think I'm going to push back against the notion that we don't want to simulate anything in this matching engine. It is relatively simple (basically just algebra) to set up an equation that solves for the total amount of funding that needs to be provided for the bond match amount. Then we can solve for the amount of funding that the long and short each need to provide, calculate the surplus that goes to the matcher, etc.

Let me know what you think about this.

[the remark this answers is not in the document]
Sean329: Implemented the idea

contracts/src/interfaces/IHyperdriveMatchingEngineV2.sol
```diff
@@ -0,0 +1,238 @@
...
+
+    /// @notice Thrown when an order is already expired.
+    error AlreadyExpired();
+
+    /// @notice Thrown when the counterparty doesn't match the counterparty
+    ///         signed into the order.
+    error InvalidCounterparty();
+
+    /// @notice Thrown when the destination for the add or remove liquidity
+    ///         options isn't configured to this contract.
```
jalextowle, line 20: This seems outdated. Do we even throw this error?

jalextowle: It would be good to go through this list and make sure that we are actually using all of the errors in the interface.

contracts/src/interfaces/IHyperdriveMatchingEngineV2.sol
```diff
@@ -0,0 +1,238 @@
...
+
+    /// @notice Emitted when the amount of funds used for an order is updated.
+    /// @param orderHash The hash of the order.
+    /// @param amountUsed The new total amount of funds used.
+    event OrderFundAmountUsedUpdated(bytes32 indexed orderHash, uint256 amountUsed);
+
+    /// @notice Emitted when the amount of bonds used for an order is updated.
+    /// @param orderHash The hash of the order.
+    /// @param amountUsed The new total amount of bonds used.
+    event OrderBondAmountUsedUpdated(bytes32 indexed orderHash, uint256 amountUsed);
```
jalextowle, line 82: I don't think we need these if we make the `OrdersMatched` event richer.

contracts/src/interfaces/IHyperdriveMatchingEngineV2.sol
```diff
@@ -0,0 +1,238 @@
...
+        /// @dev The amount to be used in the trade. In the case of `OpenLong` or 
+        ///      `OpenShort`, this is the amount of funds to deposit; and in the 
+        ///      case of `CloseLong` or `CloseShort`, this is the min amount of 
+        ///      funds to receive.
+        uint256 fundAmount;
+        /// @dev The minimum output amount expected from the trade. In the case of
+        ///      `OpenLong` or `OpenShort`, this is the min amount of bonds to 
+        ///      receive; and in the case of `CloseLong` or `CloseShort`, this is 
+        ///      the amount of bonds to close.
+        uint256 bondAmount;
```
jalextowle, line 129: I think we talked about framing this in terms of price (or rate) rather than in terms of the fund amount and bond amount. It's fine to do this later, but I am curious on your thoughts on this change. What are the pros and cons from the implementation perspective as you see it.

contracts/src/matching/HyperdriveMatchingEngineV2.sol
```diff
@@ -0,0 +1,659 @@
...
+contract HyperdriveMatchingEngineV2 is 
+    IHyperdriveMatchingEngineV2,
+    ReentrancyGuard,
+    EIP712
+{
+    using FixedPointMath for uint256;
+    using SafeERC20 for ERC20;
+
+    /// @notice The EIP712 typehash of the OrderIntent struct.
+    bytes32 public constant ORDER_INTENT_TYPEHASH =
```
jalextowle, line 35: It looks like this needs to be updated to include the new parameters like `fundAmount`, `bondAmount`, etc.

contracts/src/matching/HyperdriveMatchingEngineV2.sol
```diff
@@ -0,0 +1,659 @@
...
+        keccak256("Options(address destination,bool asBase)");
+
+    /// @notice The name of this matching engine
+    string public name;
+
+    /// @notice The kind of this matching engine
+    string public constant kind = HYPERDRIVE_MATCHING_ENGINE_KIND;
+
+    /// @notice The version of this matching engine
+    string public constant version = VERSION;
```
jalextowle, line 51: Nit: Make sure you're ending these sentences with periods. I'm seeing this elsewhere too, so it would be good for you to do a pass and fix this. This may seem pedantic, but it is about maintaining consistency.

```suggestion
    /// @notice The EIP712 typehash of the Options struct.
    bytes32 public constant OPTIONS_TYPEHASH =
        keccak256("Options(address destination,bool asBase)");

    /// @notice The name of this matching engine.
    string public name;

    /// @notice The kind of this matching engine.
    string public constant kind = HYPERDRIVE_MATCHING_ENGINE_KIND;

    /// @notice The version of this matching engine.
    string public constant version = VERSION;
```

contracts/src/matching/HyperdriveMatchingEngineV2.sol
```diff
@@ -0,0 +1,659 @@
...
+    uint256 public constant TOKEN_AMOUNT_BUFFER = 10;
+
+    /// @notice Mapping to track cancelled orders
+    mapping(bytes32 => bool) public isCancelled;
+
+    /// @notice Mapping to track the bond amount used for each order.
+    mapping(bytes32 => uint256) public orderBondAmountUsed;
+
+    /// @notice Mapping to track the amount of base used for each order.
+    mapping(bytes32 => uint256) public orderFundAmountUsed;
```
jalextowle, line 63: I haven't reviewed the logic too much, but I'm wondering if we could just use one of these (the bond amount makes more sense for me). Since orders are filled linearly, we should be able to derive one amount from the other.

contracts/src/matching/HyperdriveMatchingEngineV2.sol
```diff
@@ -0,0 +1,659 @@
...
+    function _handleShortTransfer() internal {}
+
+    /// @dev Get checkpoint and position durations from Hyperdrive contract
+    /// @param _hyperdrive The Hyperdrive contract to query
+    /// @return config The pool config
+    function _getHyperdriveDurationsAndFees(IHyperdrive _hyperdrive) internal view returns (
+        IHyperdrive.PoolConfig memory config
+    ) {
+        config = _hyperdrive.getPoolConfig();
+    }
```
jalextowle, line 644: What's the point of this function? Can't we just call `getPoolConfig` directly where we need to?

contracts/src/matching/HyperdriveMatchingEngineV2.sol
```diff
@@ -0,0 +1,659 @@
...
+    /// @param _order1 The first order to validate.
+    /// @param _order2 The second order to validate.
+    /// @return order1Hash The hash of the first order.
+    /// @return order2Hash The hash of the second order.
+    function _validateOrders(
+        OrderIntent calldata _order1,
+        OrderIntent calldata _order2
+    ) internal view returns (bytes32 order1Hash, bytes32 order2Hash) {
+        
+
```
jalextowle, line 390: Nit: Remove these newlines

```suggestion
```

contracts/src/matching/HyperdriveMatchingEngineV2.sol
```diff
@@ -0,0 +1,659 @@
...
+        // Hash orders
+        order1Hash = hashOrderIntent(_order1);
+        order2Hash = hashOrderIntent(_order2);
+
+        // Check if orders are cancelled
+        if (isCancelled[order1Hash] || isCancelled[order2Hash]) {
+            revert AlreadyCancelled();
+        }
+
```
jalextowle, line 451: Nit: Remove this newline

```suggestion
```

contracts/src/matching/HyperdriveMatchingEngineV2.sol
```diff
@@ -0,0 +1,659 @@
...
+                destination: address(this),
+                asBase: true,
+                extraData: ""
+            })
+        );
+        
+        // Transfer proceeds to traders
+        _baseToken.safeTransfer(_longOrder.options.destination, _minFundAmountLongOrder);
+        _baseToken.safeTransfer(_shortOrder.options.destination, _minFundAmountShortOrder);
+
```
jalextowle, line 630:
```suggestion
```

contracts/src/matching/HyperdriveMatchingEngineV2.sol
```diff
@@ -0,0 +1,659 @@
...
+        ) {
+            revert InvalidSettlementAsset();
+        }
+
+        // Verify valid maturity time
+        if (_order1.minMaturityTime > _order1.maxMaturityTime || 
+            _order2.minMaturityTime > _order2.maxMaturityTime 
+            ) {
+            revert InvalidMaturityTime();
+        }
```
jalextowle, line 427: We should also make sure these overlap. I also don't see any validation on these parameters in the `mint` flow. We'll need to make sure that the minted bonds satisfy both orders.

Sean329: The validation is at around Line#156 like these:
```
            // Calculate the maturity time of newly minted positions
            uint256 maturityTime = latestCheckpoint + config.positionDuration;

            // Check if the maturity time is within the range
            if (maturityTime < order1.minMaturityTime || maturityTime > order1.maxMaturityTime ||
                maturityTime < order2.minMaturityTime || maturityTime > order2.maxMaturityTime) {
                revert InvalidMaturityTime();
            }
```

Because it uses the `maturityTime` from the instance, so I put this validation inside the `matchOrders` function rather than the `_validateOrders`. 
The `_validateOrders` just does general validations; the case specific validations will be in each of the IF blocks in `matchOrders`.

contracts/src/interfaces/IHyperdriveMatchingEngineV2.sol
Sean329: I thought about it for a while and asking the traders to directly provide `fundAmount` and `bondAmount` makes it easier for the matching engine to code. The price/rate will be the factor that the traders make considerations in their head, and, in the end for whatever decision they've made, the price/rate can all be reflected in the combination of these 3 params:
1. input amount (money or positions)
2. min output amount (positions or money)
3. maturity time of the output or input positions

So, I put `fundAmount `, `bondAmount `, `minMaturityTime` and `maxMaturityTime` these four elements in the `orderIntent`. With them, the matching engine code can be straight forward; and, by carefully selecting the values for these four elements, the trader should be able to can convey their price expectations.

Let me know if this makes sense?

contracts/src/matching/HyperdriveMatchingEngineV2.sol
```diff
@@ -0,0 +1,659 @@
...
+            _order1, 
+            _order2
+        );
+
+        IHyperdrive hyperdrive = _order1.hyperdrive;
+
+        // Handle different order type combinations
+        if (_order1.orderType == OrderType.OpenLong && 
+            _order2.orderType == OrderType.OpenShort) {
+            // Case 1: Long + Short creation using mint()
```
jalextowle, line 92: Let's put this comment above the if statement. We can add similar comments above the `else if` and `else` blocks.

contracts/src/matching/HyperdriveMatchingEngineV2.sol
```diff
@@ -0,0 +1,659 @@
...
+            uint256 bondMatchAmount = _calculateBondMatchAmount(
+                order1, 
+                order2, 
+                order1Hash_, 
+                order2Hash_
+            );
+
+            
+            // Get the sufficient funding amount to mint the bonds.
+            // NOTE: Round the requred fund amount up to overestimate the cost.
```
jalextowle, line 128:
```suggestion
            // NOTE: Round the required fund amount up to overestimate the cost.
```

contracts/src/matching/HyperdriveMatchingEngineV2.sol
```diff
@@ -0,0 +1,659 @@
...
+            emit OrderFundAmountUsedUpdated(order2Hash_, orderFundAmountUsed[order2Hash_]);
+
+            // Calculate the maturity time of newly minted positions
+            uint256 maturityTime = latestCheckpoint + config.positionDuration;
+
+            // Check if the maturity time is within the range
+            if (maturityTime < order1.minMaturityTime || maturityTime > order1.maxMaturityTime ||
+                maturityTime < order2.minMaturityTime || maturityTime > order2.maxMaturityTime) {
+                revert InvalidMaturityTime();
+            }
```
jalextowle, line 163: Okay, I see the validation now

contracts/src/matching/HyperdriveMatchingEngineV2.sol
```diff
@@ -0,0 +1,659 @@
...
+                baseTokenAmountOrder1, 
+                baseTokenAmountOrder2,
+                cost, 
+                bondMatchAmount, 
+                baseToken, 
+                hyperdrive_);
+            
+            // Update order bond amount used
+            orderBondAmountUsed[order1Hash_] += bondAmount;
+            orderBondAmountUsed[order2Hash_] += bondAmount;
```
jalextowle, line 183: Shouldn't we put this before `_handleMatch` for CEI?

contracts/src/matching/HyperdriveMatchingEngineV2.sol
```diff
@@ -0,0 +1,659 @@
...
+            emit OrderBondAmountUsedUpdated(order1Hash_, orderBondAmountUsed[order1Hash_]);
+            emit OrderBondAmountUsedUpdated(order2Hash_, orderBondAmountUsed[order2Hash_]);
+
+            // Mark fully executed orders as cancelled 
+            if (orderBondAmountUsed[order1Hash_] >= order1.bondAmount || orderFundAmountUsed[order1Hash_] >= order1.fundAmount) {
+                isCancelled[order1Hash_] = true;
+            }
+            if (orderBondAmountUsed[order2Hash_] >= order2.bondAmount || orderFundAmountUsed[order2Hash_] >= order2.fundAmount) {
+                isCancelled[order2Hash_] = true;
+            }
```
jalextowle, line 193: Do we really need to do this? If the `orderBondAmountUsed` (or the `orderFundAmountUsed`) is greater than the order's specified parameter, wouldn't this fail anyways?

contracts/src/matching/HyperdriveMatchingEngineV2.sol
```diff
@@ -0,0 +1,659 @@
...
+
+            // Transfer the remaining base tokens back to the surplus recipient
+            baseToken.safeTransfer(
+                surplusRecipient,
+                baseToken.balanceOf(address(this))
+            );
+        } 
+
+
```
jalextowle, line 202:
```suggestion
```

contracts/src/matching/HyperdriveMatchingEngineV2.sol
```diff
@@ -0,0 +1,659 @@
...
+                bondMatchAmount,
+                baseToken,
+                hyperdrive
+            );
+            
+            // Update order fund amount used
+            orderFundAmountUsed[order1Hash] += minFundAmountOrder1;
+            orderFundAmountUsed[order2Hash] += minFundAmountOrder2;
+            emit OrderFundAmountUsedUpdated(order1Hash, orderFundAmountUsed[order1Hash]);
+            emit OrderFundAmountUsedUpdated(order2Hash, orderFundAmountUsed[order2Hash]);
```
jalextowle, line 253: Same here, shouldn't this be before `_handleBurn` for CEI? I also don't really see why we need to track the fund amount used since this can be derived from the bond amount used.

contracts/src/matching/HyperdriveMatchingEngineV2.sol
```diff
@@ -0,0 +1,659 @@
...
+
+            // Mark fully executed orders as cancelled
+            if (orderBondAmountUsed[order1Hash] >= _order1.bondAmount || 
+                orderFundAmountUsed[order1Hash] >= _order1.fundAmount) {
+                isCancelled[order1Hash] = true;
+            }
+            if (orderBondAmountUsed[order2Hash] >= _order2.bondAmount || 
+                orderFundAmountUsed[order2Hash] >= _order2.fundAmount) {
+                isCancelled[order2Hash] = true;
+            }
```
jalextowle, line 263: Same here, why do we need to do this?

contracts/src/matching/HyperdriveMatchingEngineV2.sol
```diff
@@ -0,0 +1,659 @@
...
+
+    /// @notice Hashes an order intent according to EIP-712
+    /// @param _order The order intent to hash
+    /// @return The hash of the order intent
+    function hashOrderIntent(
+        OrderIntent calldata _order
+    ) public view returns (bytes32) {
+        // Stack cycling to avoid stack-too-deep
+        // OrderIntent calldata order = _order;
+
```
jalextowle, line 326:
```suggestion
```

contracts/src/matching/HyperdriveMatchingEngineV2.sol
jalextowle: Okay, I missed that on my first pass. Makes sense.

contracts/src/matching/HyperdriveMatchingEngineV2.sol
```diff
@@ -0,0 +1,659 @@
...
+        // Approve Hyperdrive
+        // @dev Use balanceOf to get the total amount of base tokens instead of 
+        //      summing up the two amounts, in order to open the door for poential 
+        //      donation to help match orders.
+        uint256 totalBaseTokenAmount = _baseToken.balanceOf(address(this));
+        uint256 baseTokenAmountToUse = _cost + TOKEN_AMOUNT_BUFFER;
+        if (totalBaseTokenAmount < baseTokenAmountToUse) {
+            revert InsufficientFunding();
+        }
+        _baseToken.forceApprove(address(_hyperdrive), baseTokenAmountToUse);
```
jalextowle, line 543: Add 1 wei of approval so that the storage slot stays hot

contracts/src/matching/HyperdriveMatchingEngineV2.sol
```diff
@@ -0,0 +1,659 @@
...
+
+        // Create PairOptions
+        IHyperdrive.PairOptions memory pairOptions = IHyperdrive.PairOptions({
+            longDestination: _longOrder.options.destination,
+            shortDestination: _shortOrder.options.destination,
+            asBase: true,
+            extraData: ""
```
jalextowle, line 550: We need to add support for this. I'm not exactly sure what the best way to do this is, so let's just leave a TODO for now.

contracts/src/matching/HyperdriveMatchingEngineV2.sol
```diff
@@ -0,0 +1,659 @@
...
+        });
+
+        // Calculate minVaultSharePrice
+        uint256 minVaultSharePrice = _longOrder.minVaultSharePrice.min(_shortOrder.minVaultSharePrice);
```
jalextowle, line 554: This makes the check less strict. We want to take the max here to ensure that both traders are satisfied.

contracts/src/matching/HyperdriveMatchingEngineV2.sol
```diff
@@ -0,0 +1,659 @@
...
+
+        // Create PairOptions
+        IHyperdrive.PairOptions memory pairOptions = IHyperdrive.PairOptions({
+            longDestination: _longOrder.options.destination,
+            shortDestination: _shortOrder.options.destination,
+            asBase: true,
+            extraData: ""
+        });
+
+        // Calculate minVaultSharePrice
```
jalextowle, line 553: Let's add some detail to this comment. This doesn't add any useful information.

contracts/src/matching/HyperdriveMatchingEngineV2.sol
```diff
@@ -0,0 +1,659 @@
...
+            
+            // Get the sufficient funding amount to mint the bonds.
+            // NOTE: Round the requred fund amount up to overestimate the cost.
+            //       Round the flat fee calculation up and the governance fee
+            //       calculation down to match the rounding used in the other flows.
+            uint256 cost = bondMatchAmount.mulDivUp(
+                vaultSharePrice.max(openVaultSharePrice), 
+                openVaultSharePrice) + 
+                bondMatchAmount.mulUp(config.fees.flat) +
+                2 * bondMatchAmount.mulUp(config.fees.flat).mulDown(config.fees.governanceLP);
```
jalextowle, line 135: Looking at this and some of the other lines makes me think we need to run prettier. You can do that with `make prettier`

jalextowle: It would be good to add some rigorous tests for determining that the hardcoded buffer is sufficient for all cases. In particular, testing against some of our trickier instance tests would be good. Imprecise share accounting will drive a lot of this imprecision.

contracts/src/matching/HyperdriveMatchingEngineV2.sol
Sean329: Yeah that's right, needs to use the larger of the two to satisfy both orders

contracts/src/matching/HyperdriveMatchingEngineV2.sol
Sean329: Here are my thoughts:
1. This part is in the `_handleBurn` situation, so the input is the bondAmount, and the fundAmount is the output. Therefore, for the CEI purposes, putting the usedBondAmount mapping updates before the interaction step of `_handleBurn` is already sufficient. So, for the usedFundAmount update, it's OK to put it after the interaction, and potentially it can save some gas (if the revert happens during the burn interaction).
2. You are right that for the `_handleBurn` situation, this usedFundAmount mapping is not totally necessary because it can be calculated from the usedBondAmount mapping as the relationship remains proportional before and after the `burn` interaction. I chose to keep using the mapping here mainly because:
    - The usedFundAmount mapping is still necessary to be kept in this contract, because the `_handleMint` situation needs it, as in there the updates are not strictly proportional to the usedBond update before and after the `mint` interaction
    - Then, since we cannot totally get rid of that mapping anyway, I think maybe we should just use it in the `_handleBurn` as well, so the coding share some similarity with the `_handleMint` above. In addition, can save some lines of code repeating that calculation.

Let me know if the considerations above are valid?

contracts/src/matching/HyperdriveMatchingEngineV2.sol
Sean329: If we have these updates, then next time when such orders are used for matching again, the execution can revert at an earlier step in the `_validateOrders` due to the `isCanceled` flag. In comparison to letting it proceed to later steps when it starts to calculate the `bondMatchAmount` and revert here `revert NoBondMatchAmount()`, I think the former way can potentially save some gas?

contracts/src/interfaces/IHyperdriveMatchingEngineV2.sol
Sean329: We need this error code, because the `options.destination` will be the address to receive the output bonds or money. The input money or bonds come from the `trader` address, but output goes to `options.destination` address. 
I thought I added the zero address check for the destinations, but maybe I deleted the lines accidentally. Added them back and these lines use the `InvalidDestination` error code.

Sean329: Went thru the error code list again, and removed the following unused ones:
`InvalidFeeRecipient`,`InvalidMatch`,`InvalidOrderType`,`InvalidPoolConfig`

One exception: `InvalidSettlementAsset` will be used for now because we are coding for true+true or false+false only atm. In the future when the code is ready for mixed type match, it will be removed.

contracts/src/interfaces/IHyperdriveMatchingEngineV2.sol
Sean329: Agreed, code changed.

contracts/src/matching/HyperdriveMatchingEngineV2.sol
Sean329: Good catch! I overlooked this one. Fixed.
Question: it's OK we don't include the `feeRecipient` in the TYPEHASH, right? Although we keep it in the orderIntent struct, but the match engine V2 won't use it anywhere.

contracts/src/matching/HyperdriveMatchingEngineV2.sol
Sean329: I think there are cases we still need both. 
Explained in another comment here: https://github.com/delvtech/hyperdrive/pull/1225#discussion_r1931595246

contracts/src/matching/HyperdriveMatchingEngineV2.sol
Sean329: Oh yeah, initially I wrote this helper function because I was using low level staticcall and abi.decode...those stuff to get the values. After changing to use the HyperdriveRead interface, this does not need to be a separate function any more. Forgot to remove it.

contracts/src/matching/HyperdriveMatchingEngineV2.sol
Sean329: See my reply to another comment, similar explanations: https://github.com/delvtech/hyperdrive/pull/1225#discussion_r1931595246

contracts/src/matching/HyperdriveMatchingEngineV2.sol
Sean329: Yes that's right, it will fail at the step getting `bondMatchAmount` anyway in that case.
The reason why I still keep these lines here  --- see my reply to another comment, similar explanations: https://github.com/delvtech/hyperdrive/pull/1225#discussion_r1931602316

contracts/src/matching/HyperdriveMatchingEngineV2.sol
Sean329: Done

contracts/src/matching/HyperdriveMatchingEngineV2.sol
Sean329: Added a TODO mark, after the transfer situations and fillOrder are coded.

contracts/src/matching/HyperdriveMatchingEngineV2.sol
Sean329: Removed this part so avoided writing to new storage.
The related validations are moved into `_validateOrders ` so when such orders are matched again next time, it reverts there as `error AlreadyFullyExecuted();`

contracts/src/matching/HyperdriveMatchingEngineV2.sol
Sean329: Removed this part so avoided writing to new storage.
The related validations are moved into `_validateOrders ` so when such orders are matched again next time, it reverts there as `error AlreadyFullyExecuted();`

contracts/src/interfaces/IHyperdriveMatchingEngineV2.sol
jalextowle: Sounds good. I'll resolve it.

contracts/src/interfaces/IHyperdriveMatchingEngineV2.sol
jalextowle: Works for me. Would be good to poke Mihai about it again.

contracts/src/matching/HyperdriveMatchingEngineV2.sol
jalextowle: Yeah, let's remove fee recipient here and in the order. I'm not sure why we need it, but maybe I'm missing somethign.

contracts/src/matching/HyperdriveMatchingEngineV2.sol
```diff
@@ -0,0 +1,645 @@
...
+                _order2, 
+                order1Hash, 
+                order2Hash
+            );
+
+            // Get the min fund output according to the bondMatchAmount.
+            // NOTE: Round the required fund amount up to respect the order specified
+            //       min fund output.
+            uint256 minFundAmountOrder1 = (_order1.fundAmount - orderFundAmountUsed[order1Hash]).mulDivUp(bondMatchAmount, (_order1.bondAmount - orderBondAmountUsed[order1Hash]));
+            uint256 minFundAmountOrder2 = (_order2.fundAmount - orderFundAmountUsed[order2Hash]).mulDivUp(bondMatchAmount, (_order2.bondAmount - orderBondAmountUsed[order2Hash]));
```
jalextowle, line 212: These lines still look way too long. I think running `make prettier` would fix this

contracts/src/matching/HyperdriveMatchingEngineV2.sol
jalextowle: I think your reasoning is sound. This checks out to me

contracts/src/matching/HyperdriveMatchingEngineV2.sol
```diff
@@ -0,0 +1,645 @@
...
+
+    /// @notice The buffer amount used for cost related calculations.
+    /// @dev TODO: The buffer amount needs more testing.
+    uint256 public constant TOKEN_AMOUNT_BUFFER = 10;
+
+    /// @notice Mapping to track cancelled orders.
+    mapping(bytes32 => bool) public isCancelled;
+
+    /// @notice Mapping to track the bond amount used for each order.
+    mapping(bytes32 => uint256) public orderBondAmountUsed;
```
jalextowle, line 61: Why don’t we combine this with the fund amount mapping? If we put both fields in a struct sized with `uint128` we’d get significant gas savings

jalextowle (approved):
LGTM once the comments are addressed and CI is passing

contracts/src/matching/HyperdriveMatchingEngineV2.sol
Sean329: Done

contracts/src/matching/HyperdriveMatchingEngineV2.sol
Sean329: Added fuzzing unit test to make the `fundAmount1 + fundAmount2` happens to be equal to `cost + BUFFER` every time. If the BUFFER of 10 wei is sufficient, then the function call should not be reverted due to not meeting the min output in every fuzzing.

Sean329:
Another more comprehensive PR #1239 will be merged into main. This PR and its branch shall be discarded.

PR #1227, opened 2024-12-16
Ensure that Morpho is the caller of `onMorphoFlashLoan`

jalextowle:
# Description

Fixes a critical vulnerability in the matching engine. Previously anyone could call `onMorphoFlashLoan` which could be used to drain approvals by circumventing the validation logic.


mcclurejt (approved):

Sean329:
LGTM, 
@jalextowle  but a nit for thoughts: in the current code, if the immediate returned `proceeds` is smaller than `lpAmount`, then the trxn will revert with a quite general error code like `SafeERC20FailedOperation()`; if we `require(proceeds >= lpAmount)` within the body of `onMorphoFlashLoan` then we have the chance to give it a better clarified internal error code. 

However, it comes with a con which is that, we will lose the potential workaround to "help" some `matchOrder` to succeed by donating some base tokens into the match engine contract even when the liquidity amount returned from hyperdrive is not good enough.

jalextowle:
> LGTM, @jalextowle but a nit for thoughts: in the current code, if the immediate returned `proceeds` is smaller than `lpAmount`, then the trxn will revert with a quite general error code like `SafeERC20FailedOperation()`; if we `require(proceeds >= lpAmount)` within the body of `onMorphoFlashLoan` then we have the chance to give it a better clarified internal error code.
> 
> However, it comes with a con which is that, we will lose the potential workaround to "help" some `matchOrder` to succeed by donating some base tokens into the match engine contract even when the liquidity amount returned from hyperdrive is not good enough.

I'm glad you brought this up. It's good to think about this. As discussed offline, we'll keep as-is based on your last comment.

Sean329 (approved):

PR #1230, opened 2024-12-20
Adds integration for Savings USDS on L2 chains + tests on Base

MazyGio:
The Savings USDS (SUSDS) token on L2s doesn't have staking logic. Instead, staking functions as a swap through a PSM contract, which uses an underlying `rateProvider` oracle contract for the exchange rate.

This PR includes the integration code and an instance test for the SUSDS deployment on Base.


contracts/src/instances/savings-usds-l2/SavingsUSDSL2Base.sol
```diff
@@ -0,0 +1,175 @@
...
+        amountWithdrawn = _PSM.swapExactOut(
+            address(_vaultSharesToken),
+            address(_baseToken),
+            _convertToBase(_shareAmount) - 1, // minus 1 because the PSM swap function rounds up.
+            _shareAmount,
+            _destination,
+            0
+        );
+
+        return _convertToBase(amountWithdrawn);
```
Sean329, line 119: I think for the Line#110 here, we do not have to use the `_PSM.swapExactOut`, instead, we can still use the `swapExactIn` for the purpose of `_withdrawWithBase` because the input param used is still the `uint256 _shareAmount`. The good thing of using `swapExactIn` is that it makes the return value simpler.

The current code of using `_PSM.swapExactOut` and again for the return value it needs to do `_convertToBase(amountWithdrawn)`, this looks unnecessary. In addition, the exchange rate used at the swap and the exchange rate used later in the preview for the return value maybe different, I guess. Because one is referring to the spot conversion at the 1st swap, but the other one is simulating a consecutively 2nd swap? Could mess up the accounting.

MazyGio: You're right! Having fewer back-and-forth conversions did improve accounting precision. Pushing the change, thanks!

contracts/src/deployers/savings-usds-l2/SavingsUSDSL2HyperdriveDeployerCoordinator.sol
```diff
@@ -0,0 +1,171 @@
...
+        }
+
+        // Ensure that the base token address is properly configured.
+        if (address(_deployConfig.baseToken) != address(PSM.usds())) {
+            revert IHyperdriveDeployerCoordinator.InvalidBaseToken();
+        }
+
+        // Ensure that the minimum share reserves are equal to 1e15. This value
+        // has been tested to prevent arithmetic overflows in the
+        // `_updateLiquidity` function when the share reserves are as high as
```
jalextowle, line 149: This comment is outdated. It can support much larger share reserves

contracts/src/deployers/savings-usds-l2/SavingsUSDSL2HyperdriveDeployerCoordinator.sol
```diff
@@ -0,0 +1,171 @@
...
+    HyperdriveDeployerCoordinator
+{
+    using SafeERC20 for ERC20;
+
+    /// @notice The deployer coordinator's kind.
+    string public constant override kind =
+        SAVINGS_USDS_L2_HYPERDRIVE_DEPLOYER_COORDINATOR_KIND;
+
+    /// @notice The PSM contract.
+    IPSM public immutable PSM;
```
jalextowle, line 31: Why are we passing the PSM as extra data into the target deployers when we set it as an immutable here? Maybe I’m missing something, but we typically either use immutables everywhere or use extra data everything (including in the coordinator).

contracts/src/instances/savings-usds-l2/SavingsUSDSL2Base.sol
```diff
@@ -0,0 +1,176 @@
...
+    /// @dev Accepts a deposit from the user in base.
+    /// @param _baseAmount The base amount to deposit.
+    /// @return The shares that were minted in the deposit.
+    /// @return The amount of ETH to refund. Since this yield source isn't
+    ///         payable, this is always zero.
+    function _depositWithBase(
+        uint256 _baseAmount,
+        bytes calldata // unused
+    ) internal override returns (uint256, uint256) {
+        if (_baseAmount == 0) {
```
jalextowle, line 43: Do we need this? Wouldn’t this be caught by the minimum transaction amount check?

contracts/src/instances/savings-usds-l2/SavingsUSDSL2Base.sol
```diff
@@ -0,0 +1,176 @@
...
+
+        // Deposit the base into the yield source.
+        //
+        // NOTE: We increase the required approval amount by 1 wei so that
+        // the vault ends with an approval of 1 wei. This makes future
+        // approvals cheaper by keeping the storage slot warm.
+        ERC20(address(_baseToken)).forceApprove(address(_PSM), _baseAmount + 1);
+
+        // Depositing amounts to swapping USDS for SUSDS in the PSM.
+        uint256 sharesMinted = _PSM.swapExactIn(
```
jalextowle, line 64: Looks good, but I’ll give this another pass tomorrow to read through the PSM logic

contracts/src/instances/savings-usds-l2/SavingsUSDSL2Base.sol
```diff
@@ -0,0 +1,176 @@
...
+            address(_baseToken),
+            address(_vaultSharesToken),
+            _baseAmount,
+            _convertToShares(_baseAmount) - 1,
```
jalextowle, line 68: Can you add a comment for why we are subtracting 1 here? Are we doing this because tests broke or because we did this elsewhere? Why we’d do it makes sense if sUSDS was a rebasing token, but it isn’t, so I wouldn’t expect this to be necessary. Maybe I’m missing something though.

contracts/src/instances/savings-usds-l2/SavingsUSDSL2Base.sol
```diff
@@ -0,0 +1,176 @@
...
+
+    /// @param _shareAmount The amount of vault shares to withdraw.
+    /// @param _destination The destination of the withdrawal.
+    /// @return amountWithdrawn The amount of base withdrawn.
+    function _withdrawWithBase(
+        uint256 _shareAmount,
+        address _destination,
+        bytes calldata // unused
+    ) internal override returns (uint256 amountWithdrawn) {
+        if (_shareAmount <= 1) {
```
jalextowle, line 100: Did you run into this issue during testing? The 0 case is covered here, so we may want to update the comment: https://github.com/delvtech/hyperdrive/blob/02d3cd465afbf86cc5f490c618a3a850c6dc6bfe/contracts/src/internal/HyperdriveBase.sol#L106

contracts/src/instances/savings-usds-l2/SavingsUSDSL2Base.sol
```diff
@@ -0,0 +1,176 @@
...
+            // The PSM swap function will revert if the amount to swap is 0,
+            // but the output would just be 0 anyway.
+            return 0;
+        }
+        ERC20(address(_vaultSharesToken)).forceApprove(
```
jalextowle, line 105: Let’s either comment this or move it under the next comment block to follow the style guide

contracts/src/instances/savings-usds-l2/SavingsUSDSL2Conversions.sol
```diff
@@ -0,0 +1,54 @@
...
+        uint256 _baseAmount
+    ) internal view returns (uint256) {
+        /// Sky's internal accounting uses RAY units (1e27), so we want to
+        /// ensure that we're using the same precision.
+        return
+            _baseAmount.mulDivDown(
+                1e27,
+                IRateProvider(_PSM.rateProvider()).getConversionRate()
+            );
+        // return _PSM.previewSwapExactIn(
```
jalextowle, line 48: Did you mean to leave this in?

contracts/src/instances/savings-usds-l2/SavingsUSDSL2Hyperdrive.sol
```diff
@@ -0,0 +1,100 @@
...
+            _target1,
+            _target2,
+            _target3,
+            _target4
+        )
+        SavingsUSDSL2Base(_PSM)
+    {
+        // Approve the base token with 1 wei. This ensures that all of the
+        // subsequent approvals will be writing to a dirty storage slot.
+        ERC20(address(_config.baseToken)).forceApprove(address(this), 1);
```
jalextowle, line 97: What’s the point of these approvals? We’re approving this address to spend the tokens, but it should already be able to do that by just calling `transfer` (or using the `SafeERC20.safeTransfer` library function.

test/instances/savings-usds-l2/SavingsUSDSL2HyperdriveInstanceTest.t.sol
```diff
@@ -0,0 +1,215 @@
...
+        uint256 timeDelta,
+        int256 variableRate
+    ) internal override {
+        IRateProvider rateProvider = IRateProvider(PSM.rateProvider());
+        uint256 chi = rateProvider.getChi();
+        uint256 rho = rateProvider.getRho();
+        uint256 ssr = rateProvider.getSSR();
+        chi = (_rpow(ssr, block.timestamp - rho) * chi) / RAY;
+
+        // console.log("chi: ", chi);
```
jalextowle, line 138: Can we remove all of the commented out code?

contracts/src/instances/savings-usds-l2/SavingsUSDSL2Conversions.sol
MazyGio: Not at all, removed now :)

contracts/src/instances/savings-usds-l2/SavingsUSDSL2Base.sol
MazyGio: I'd run into this issue, but I'm not sure how/why. I've removed it now and the tests still pass.

contracts/src/instances/savings-usds-l2/SavingsUSDSL2Base.sol
MazyGio: I had to do that due to rounding issues from unnecessary back-and-forth conversions between shares and base. Sean pointed it out as well and it's no longer an issue, I've now removed this subtraction.

contracts/src/instances/savings-usds-l2/SavingsUSDSL2Base.sol
MazyGio: It should indeed. I can't recall why I added that in the first place, but it was likely while hunting down test failures. I've removed it now.

contracts/src/deployers/savings-usds-l2/SavingsUSDSL2HyperdriveDeployerCoordinator.sol
MazyGio: Oh, great point, it is absolutely not necessary! I've switched to just using the extraData.

contracts/src/deployers/savings-usds-l2/SavingsUSDSL2HyperdriveDeployerCoordinator.sol
MazyGio: Deleted the part about 200M shares.

contracts/src/instances/savings-usds-l2/SavingsUSDSL2Base.sol
MazyGio: Done!

contracts/src/instances/savings-usds-l2/SavingsUSDSL2Hyperdrive.sol
MazyGio: I'd seen them on other contracts, but didn't quite get why they were there. Removed these now.

test/instances/savings-usds-l2/SavingsUSDSL2HyperdriveInstanceTest.t.sol
MazyGio: Absolutely, didn't mean to leave that in. Thanks for catching!

Sean329:
Thanks for the previous fixes. One more nit here: since we've decided to only utilize PSM's `swapExactIn` related functions, then the interface `IPSM` can be modified to have the `swapExactOut` and `previewSwapExactOut` related code removed as they are not being used. 
But plz wait to make this change after Alex gives it a pass for the PSM use cases and we are sure about only using the `swapExactIn`.

contracts/src/instances/savings-usds-l2/SavingsUSDSL2Base.sol
```diff
@@ -0,0 +1,164 @@
...
+        ERC20(address(_baseToken)).forceApprove(address(_PSM), _baseAmount + 1);
+
+        // Depositing amounts to swapping USDS for SUSDS in the PSM.
+        uint256 sharesMinted = _PSM.swapExactIn(
+            address(_baseToken),
+            address(_vaultSharesToken),
+            _baseAmount,
+            _convertToShares(_baseAmount),
+            address(this),
+            0
```
jalextowle, line 64: Might be good to reach out to Spark for a referral code. Charles would be a good person to ask about this. It doesn't need to block the PR, but I don't want us to forget about this

jalextowle (approved):
Really nice job on this! Looks good to me once any lingering comments from me or Sean are addressed

PR #1239, opened 2025-02-04
Matching Engine V2 new PR on top of the part 1 PR#1225 with already approved code

### LINKED ISSUE #1222: Hyperdrive Matching Engine

Sean329:
# HyperdriveMatchingEngineV2 PR Description

## Overview
This PR introduces `HyperdriveMatchingEngineV2`, an enhanced matching engine for Hyperdrive that facilitates efficient order matching and position trading. The engine supports direct position minting/burning and position transfers between traders.

### Order Types Support
- **OpenLong**: Open a long position
- **OpenShort**: Open a short position
- **CloseLong**: Close an existing long position
- **CloseShort**: Close an existing short position

### Core Functionalities
1. **Order Matching (`matchOrders`)**
   - Matches complementary orders (e.g., OpenLong + OpenShort)
   - Supports position transfers between traders
   - Handles partial fills
   - Includes surplus distribution mechanism

2. **Direct Order Filling (`fillOrder`)**
   - Allows immediate order execution by takers
   - Supports maker-taker model
   - Maintains order state tracking

3. **EIP-712 Signature Support**
   - Secure order signing and verification
   - Support for both EOA and contract signatures

### Safety Features
- Reentrancy protection
- Order expiration checks
- Counterparty validation
- Maturity time validation
- Fund amount validation
- Token amount buffer for gas optimization

## Technical Improvements
- Direct mint/burn operations instead of flash loans
- Optimized gas usage with TOKEN_AMOUNT_BUFFER
- Enhanced order state tracking
- Comprehensive error handling

## Testing Coverage
The implementation includes extensive test coverage:
- Basic order matching scenarios
- Position transfer cases
- Partial fill scenarios
- Edge cases and failure conditions
- Fuzzing tests for TOKEN_AMOUNT_BUFFER validation

## Interface Changes
The `IHyperdriveMatchingEngineV2` interface provides:
- Clear error definitions
- Comprehensive event logging
- Structured order intent definition
- Order amount tracking

## Security Considerations
- All monetary operations use SafeERC20
- Strict validation of order parameters
- Protected against common attack vectors
- Clear access control mechanisms

## Breaking Changes
None. This is a new contract implementation that can coexist with previous versions.

## Next Steps
- [ ] Complete security audit
- [ ] Add integration tests
- [ ] Update documentation
- [ ] Deploy to testnet

## Related Issues
- Closes #[issue-number]
- References #[related-issue-number]

resolves #1222


contracts/src/matching/HyperdriveMatchingEngineV2.sol
```diff
@@ -72,154 +69,151 @@ contract HyperdriveMatchingEngineV2 is
     /// @notice Matches two orders.
```
jalextowle, line 69: I think we need some more documentation about how the orders are structured in the different matching situations. The idea that `order1` needs to be `OpenLong` if `order2` is `OpenShort` is pretty opinionated. I honestly think it would be easy enough to handle both cases and just record the `longOrder` and `shortOrder` at the start of the if-statement, but if we're going to stick with this kind of implementation, then it needs to be spelled out clearly for integrators.

contracts/src/matching/HyperdriveMatchingEngineV2.sol
```diff
@@ -228,56 +222,607 @@ contract HyperdriveMatchingEngineV2 is
...
 
     /// @notice Allows traders to cancel their orders.
     /// @param _orders Array of orders to cancel.
-    function cancelOrders(OrderIntent[] calldata _orders) external nonReentrant {
+    function cancelOrders(
+        OrderIntent[] calldata _orders
+    ) external nonReentrant {
         bytes32[] memory orderHashes = new bytes32[](_orders.length);
-        
+
```
jalextowle, line 825: Nit: Either remove this or add a comment per the style guide

contracts/src/matching/HyperdriveMatchingEngineV2.sol
```diff
@@ -605,41 +1245,183 @@ contract HyperdriveMatchingEngineV2 is
...
+        //      first, then transfer the needed amount to the close trader.
+        _fundToken.safeTransferFrom(
+            _openOrder.trader,
+            address(this),
+            _fundTokenAmountOpenOrder
+        );
+        _fundToken.safeTransfer(
+            _closeOrder.options.destination,
+            _minFundAmountCloseOrder
+        );
```
jalextowle, line 1335: It would be quite a bit more efficient if we just did the transfer directly. I get what you're trying to do, but the likelihood that there are donation tokens feels low enough to not warrant the additional cost IMO

contracts/src/matching/HyperdriveMatchingEngineV2.sol
```diff
@@ -605,41 +1245,183 @@ contract HyperdriveMatchingEngineV2 is
...
+        IHyperdrive.PoolConfig memory config = _hyperdrive.getPoolConfig();
+
+        // Calculate checkpoint and maturity time.
+        uint256 latestCheckpoint = _latestCheckpoint(config.checkpointDuration);
+        maturityTime = latestCheckpoint + config.positionDuration;
+
+        // Get vault share prices.
+        // @dev TODO: there is another way to get the info without calling
+        //      getPoolInfo()?
+        uint256 vaultSharePrice = _hyperdrive.getPoolInfo().vaultSharePrice;
```
jalextowle, line 1369: Yes, you can call `hyperdrive.convertToBase(ONE)`.

contracts/src/matching/HyperdriveMatchingEngineV2.sol
Sean329: Spelled out documentation.

contracts/src/matching/HyperdriveMatchingEngineV2.sol
Sean329: Agreed

contracts/src/matching/HyperdriveMatchingEngineV2.sol
Sean329: Is ONE always supposed to be 1e18? What if the base is 1e6 like USDC?

contracts/src/matching/HyperdriveMatchingEngineV2.sol
Sean329: Actually, I just recalled why I did it this way --- even if we directly transfer the `_minFundAmountCloseOrder` from the openOrder trader to the closeOrder destination, we still need to transfer any surplus amount `_fundTokenAmountOpenOrder  - _minFundAmountCloseOrder` into `address(this)` because the handling of surplus to the caller or taker happens in the upper level function call. Therefore, we need 2 transfers no matter how.

contracts/src/matching/HyperdriveMatchingEngineV2.sol
Sean329: Implemented

jalextowle (approved):

PR #1241, opened 2025-02-25
Adding more tests to Matching Engine V2

Sean329:
Adding more tests to Matching Engine V2

### DIFF
--- a/contracts/src/matching/HyperdriveMatchingEngineV2.sol
+++ b/contracts/src/matching/HyperdriveMatchingEngineV2.sol
@@ -1459,6 +1459,18 @@ contract HyperdriveMatchingEngineV2 is
         uint256 governanceFee = 2 * flatFee.mulDown(config.fees.governanceLP);
         cost += governanceFee;
 
+        // @dev we can use hyperdrive.convertToBase and hyperdrive.convertToShares
+        //      to simulate the conversions made within the yield source. When we
+        //      are using base, what happens under the hood is that the calculations
+        //      will flow as:
+        //      1. Deposit base into yield source.
+        //      2. Calculate the shares minted with hyperdrive.convertToShares(baseAmount)
+        //      3. Calculate the final base amount with
+        //         finalBaseAmount =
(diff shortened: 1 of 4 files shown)


contracts/src/matching/HyperdriveMatchingEngineV2.sol
```diff
@@ -1453,6 +1458,14 @@ contract HyperdriveMatchingEngineV2 is
         // NOTE: Round the governance fee calculation down to match other flows.
         uint256 governanceFee = 2 * flatFee.mulDown(config.fees.governanceLP);
         cost += governanceFee;
+
+        if (_asBase) {
+            // NOTE: Round up to overestimate the cost.
```
jalextowle, line 1463: It would be good to add some explanation for this comment. You can use the explanation I gave you on Discord.

test/units/matching/HyperdriveMatchingEngineV2Test.t.sol
```diff
@@ -853,5 +1105,13 @@ contract HyperdriveMatchingEngineV2Test is HyperdriveTest {
         // NOTE: Round the governance fee calculation down to match other flows.
         uint256 governanceFee = 2 * flatFee.mulDown(config.fees.governanceLP);
         cost += governanceFee;
+
```
jalextowle, line 1108: What is this function used for? Generally, I try to avoid duplicated logic like this in tests, but let me know if I'm missing something.

Sean329: In the fuzzing function `testFuzz_tokenAmountBuffer(uint256 bondAmount)` I just want to make sure the sum of the two funding sources are absolutely sufficient so I made it `uint256 fundAmount2 = cost + 10 - fundAmount1` which needs this piece of cost estimation logic. But actually I think I can make the fund way more than sufficient like `fundAmount1 = 3 * bondAmount` so I do not need to calculate what the cost is, then this helper function can be avoided. 
I'll update the code.

jalextowle (approved):

